PR #13, opened 2018-06-20
Generate manifest response, mainly copying from Jetpack

### LINKED ISSUE #11: Standardize a WP approach to app manifest
As a developer of a PWA, I should use an app manifest to describe what my application does - knowing that there can only be one. I should be respectful of other app approaches, and I should have some way to play nicely if there are several app manifests.

AC1: Based on discovery of WP approaches (core, plugins, themes) from #1, document how app manifests are used.
AC2: Provide a basic hierarchy (or some core definition) on a perhaps per-theme basis to establish direction.

Scenario:
A user installs _Super Awesome Plugin_ and _Cool PWA theme_. They both use PWA features; whose app manifest rules them both? We _shouldn't_ go overkill (i.e. [theme template overrides](https://developer.wordpress.org/themes/basics/template-hierarchy/), which works well but is quite complex) but we should find a straightforward approach here.

kienstra:
* Uses the methods in [Jetpack_PWA_Manifest](https://github.com/Automattic/jetpack/blob/40cdf9974b73b337ffe8a7693272d68a490e1b51/modules/pwa/class.jetpack-pwa-manifest.php) and [Jetpack_PWA_Helpers](https://github.com/Automattic/jetpack/blob/40cdf9974b73b337ffe8a7693272d68a490e1b51/modules/pwa/class.jetpack-pwa-helpers.php), with slight revisions. Props [ebinnion](https://github.com/ebinnion).
* Outputs the manifest response, mainly based on existing values like `get_bloginfo( 'name' )`.
* There's still a [filter](https://github.com/xwp/pwa-wp/compare/add/11-app-manifest?expand=1#diff-f875394ddc0965b517a89f44c0e877e2R151) to override or add anything.
* There isn't any specific `/wp-admin` UI for entering manifest values.

app-manifest

Closes #11.


php/class-pwawp-app-manifest.php
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * @package PWA
+ */
+
+/**
+ * PWAWP_APP_Manifest class.
+ *
+ * Mainly copied from Jetpack_PWA_Manifest and Jetpack_PWA_Helpers.
+ */
+class PWAWP_APP_Manifest {
```
westonruter: Let's name things as if they are going to be merged into core, so this would be `WP_Web_App_Manifest`

php/class-pwawp-app-manifest.php
```diff
@@ -0,0 +1,171 @@
...
+	 * Gets the theme color for the manifest.
+	 *
+	 * Mainly copied from Jetpack_PWA_Helpers::get_theme_color().
+	 * First looks for the header background color in the AMP for WordPress plugin, if it's active.
+	 * This color displays on loading the app.
+	 *
+	 * @return string $theme_color The theme color for the manifest.json file, as a hex value.
+	 */
+	public function get_theme_color() {
+		if ( class_exists( 'AMP_Customizer_Settings' ) ) {
```
westonruter: I don't think we should be considering the AMP plugin here. Only standard theme features should be used.

php/class-pwawp-app-manifest.php
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * Outputs the <link> and <meta> tags for the app manifest.
+	 *
+	 * Mainly copied from Jetpack_PWA_Manifest::render_manifest_link().
+	 */
+	public function manifest_link_and_meta() {
+		?>
+		<link rel="manifest" href="<?php echo esc_url( add_query_arg( self::MANIFEST_QUERY_ARG, '1', home_url( '/' ) ) ); ?>">
+		<meta name="theme-color" content="<?php echo esc_attr( $this->get_theme_color() ); ?>">
```
westonruter: Is it better to define this in meta vs the manifest?

php/class-pwawp-app-manifest.php
```diff
@@ -0,0 +1,171 @@
...
+			}
+		} elseif ( current_theme_supports( 'custom-background' ) ) {
+			$background_color = get_background_color(); // This returns a hex value without the leading #, or an empty string.
+			if ( $background_color ) {
+				$theme_color = "#$background_color";
+			}
+		}
+
+		if ( ! isset( $theme_color ) ) {
+			$theme_color = self::FALLBACK_THEME_COLOR;
```
westonruter: What about just omitting the theme color if one is not detected from theme mods?

php/class-pwawp-app-manifest.php
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * Sends the json response of the manifest.
+	 *
+	 * Mainly copied from Jetpack_PWA_Helpers::render_manifest_json().
+	 */
+	public function send_manifest_json() {
+
+		// Don't load the manifest in multiple locations.
+		if ( is_front_page() && ! empty( $_GET[ self::MANIFEST_QUERY_ARG ] ) ) { // WPCS: CSRF ok.
```
westonruter: I think we may want to register an actual query var so that we can just do `get_query_var( 'web_app_manifest' )` for example.

westonruter: On second thought, why not just leverage the REST API for the manifest endpoint?

php/class-pwawp-app-manifest.php
```diff
@@ -0,0 +1,171 @@
...
+	public $default_manifest_icon_sizes = array( 192, 512 );
+
+	/**
+	 * Inits the class.
+	 *
+	 * Mainly copied from Jetpack_PWA_Manifest::__construct().
+	 */
+	public function init() {
+		add_action( 'wp_head', array( $this, 'manifest_link_and_meta' ) );
+		add_action( 'amp_post_template_head', array( $this, 'manifest_link_and_meta' ) );
```
westonruter: This AMP action should be omitted. It should be up to the AMP plugin to add the link.

php/class-pwawp-app-manifest.php
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	/**
+	 * Outputs the <link> and <meta> tags for the app manifest.
+	 *
+	 * Mainly copied from Jetpack_PWA_Manifest::render_manifest_link().
+	 */
+	public function manifest_link_and_meta() {
+		?>
+		<link rel="manifest" href="<?php echo esc_url( add_query_arg( self::MANIFEST_QUERY_ARG, '1', home_url( '/' ) ) ); ?>">
```
westonruter: Note below about using REST API for the URL

php/class-pwawp-app-manifest.php
```diff
@@ -0,0 +1,171 @@
...
+	 */
+	public function send_manifest_json() {
+
+		// Don't load the manifest in multiple locations.
+		if ( is_front_page() && ! empty( $_GET[ self::MANIFEST_QUERY_ARG ] ) ) { // WPCS: CSRF ok.
+			$theme_color = $this->get_theme_color();
+			$manifest    = array(
+				'name'             => get_bloginfo( 'name' ),
+				'start_url'        => get_home_url(),
+				'display'          => 'standalone',
```
westonruter: I think the default `display` should be `minimal-ui` or `browser`. See https://core.trac.wordpress.org/ticket/43328#comment:4

gravityrail:
I don't have anything to add to @westonruter's comments, except to say thanks for doing this and I'm happy to see Jetpack's code has been useful as a starting point. 👏

php/class-pwawp-app-manifest.php
kienstra: Good point, [a74140](https://github.com/xwp/pwa-wp/pull/13/commits/a74140d231fc31958c4010429abd2637dfbb08b9#diff-f875394ddc0965b517a89f44c0e877e2L72) removes references to the AMP plugin.

php/class-pwawp-app-manifest.php
kienstra: Thanks, [a7414](https://github.com/xwp/pwa-wp/pull/13/commits/a74140d231fc31958c4010429abd2637dfbb08b9#diff-f875394ddc0965b517a89f44c0e877e2L46) removes the AMP action.

php/class-pwawp-app-manifest.php
kienstra: Sure, [efc785](https://github.com/xwp/pwa-wp/pull/13/commits/efc785fb32e1b2e2a7cedf0458e62d4fa115b954) changes the `display` to `minimal-ui`. Thanks for linking to the Trac ticket, also. It's interesting to see that.

php/class-pwawp-app-manifest.php
kienstra: Good question. I'm not sure, but it looks like it's suggested to have both:
>The audit passes if Lighthouse finds a theme-color meta tag in the page's HTML and a theme_color property in the Web App Manifest. Lighthouse does not test whether the values are valid CSS color values.

([source](https://developers.google.com/web/tools/lighthouse/audits/address-bar#more-info))
 
But this suggests that it ignores the manifest `theme-color` if it's present in the `<meta>` tag:

>If you page already has a theme-color meta tag — for example <meta name="theme-color" content="#2196F3"> — then the page level configuration will be used instead of the value in the manifest.

([source](https://developers.google.com/web/updates/2015/08/using-manifest-to-set-sitewide-theme-color))

php/class-pwawp-app-manifest.php
kienstra: That makes sense, as an arbitrary `FALLBACK_THEME_COLOR` seems strange. But it looks like there's a Lighthouse error when there's no `background_color` or `theme_color` at all:

error-background-color-theme-color

php/class-pwawp-app-manifest.php
kienstra: Sure, [6df56](https://github.com/xwp/pwa-wp/pull/13/commits/6df5656d56a0f2c9e4893b63d326be6f1369ee15) renames the classes.

php/class-pwawp-app-manifest.php
kienstra: Sounds good. I'll probably do that tomorrow (Thursday) if that's alright.

kienstra: [3bc37](https://github.com/xwp/pwa-wp/pull/13/commits/3bc378a31ce34a6bc0621ed9e55c71ae9e4421ce) adds a REST API endpoint for the manifest, and uses it in the `<link rel="manifest">`. 

The [namespace and route](https://github.com/xwp/pwa-wp/pull/13/commits/3bc378a31ce34a6bc0621ed9e55c71ae9e4421ce#diff-2ef052ceff51a4e22091fcdab572fc69R34) still might need to be changed. Right now, the endpoint is `/wp-json/wp/v2/pwa-manifest`.

It looks like almost all of the endpoints that Core registers have the namespace of `wp/v2`:

core-namespaces

php/class-pwawp-app-manifest.php
kienstra: [3bc37](https://github.com/xwp/pwa-wp/pull/13/commits/3bc378a31ce34a6bc0621ed9e55c71ae9e4421ce) adds a REST API endpoint for the manifest, and uses it in the `<link rel="manifest">`.

westonruter:
For the errors in WP<4.9, I suggest eliminating older versions of WP from the build matrix. The plugin should make WP 4.9 as the minimum version supported.

php/class-wp-app-manifest.php
```diff
@@ -0,0 +1,209 @@
...
+		 * Gets the 'short_name' by limiting the blog name to 12 characters.
+		 * And if this cuts off a word, it omits the word entirely by using a positive look-ahead in the regex.
+		 * For example, the first 12 characters of 'My PWA WordPress Site' are 'My PWA WordP'.
+		 * Because this cuts off the last word, this removes 'WordP' entirely: 'My PWA'.
+		 *
+		 * @link https://stackoverflow.com/questions/12646197/cut-the-string-to-be-80-characters-and-must-keep-the-words-without-cutting-th#answer-12646400
+		 */
+		preg_match( '/^.{0,12}(?= |$)/', get_bloginfo( 'name' ), $short_name_matches );
+		if ( $short_name_matches ) {
+			$manifest['short_name'] = $short_name_matches[0];
```
kienstra: The `short_name` might need more thought. This limits it to the maximum of 12 characters that [Chrome recommends](https://developer.chrome.com/apps/manifest/name#short_name). But this could still be awkward.

kienstra:
**Request For Another Review**

Hi @westonruter, 
Thanks for reviewing this earlier. All of your points are applied (pending your review), including the [REST API endpoint](https://github.com/xwp/pwa-wp/pull/13#discussion_r197239393).

The endpoint is `/wp-json/wp/v2/pwa-manifest`

Here's an example response:
```json
{
    "name": "Local PWA Development",
    "start_url": "https://loc.test",
    "display": "minimal-ui",
    "lang": "en_US",
    "dir": "ltr",
    "short_name": "Local PWA",
    "background_color": "#1e73be",
    "theme_color": "#1e73be",
    "description": "Just another WordPress site",
    "icons": [
        {
            "src": "https://loc.test/wp-content/uploads/2018/06/cropped-lorem-2000-1-192x192.jpeg",
            "sizes": "192x192",
            "type": "image/jpeg"
        },
        {
            "src": "https://loc.test/wp-content/uploads/2018/06/cropped-lorem-2000-1.jpeg",
            "sizes": "512x512",
            "type": "image/jpeg"
        }
    ]
}
```

php/class-wp-app-manifest.php
```diff
@@ -0,0 +1,202 @@
...
+		if ( ! isset( $theme_color ) ) {
+			$theme_color = self::FALLBACK_THEME_COLOR;
+		}
+
+		/**
+		 * Enables overriding the PWA theme color.
+		 *
+		 * @param string $theme_color Hex color value.
+		 */
+		return apply_filters( 'pwa_background_color', $theme_color );
```
westonruter: Instead of adding a filter for this, we can just let the color be filtered via the `pwa_manifest_json` filter.

php/class-wp-app-manifest.php
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * @package PWA
+ */
+
+/**
+ * WP_APP_Manifest class.
+ *
+ * Mainly copied from Jetpack_PWA_Manifest and Jetpack_PWA_Helpers.
+ */
+class WP_APP_Manifest {
```
westonruter: This can be named `WP_Web_App_Manifest`

php/class-wp-app-manifest.php
```diff
@@ -0,0 +1,202 @@
...
+	 * @var string
+	 */
+	const REST_NAMESPACE = 'wp/v2';
+
+	/**
+	 * The REST API route for the manifest request.
+	 *
+	 * @var string
+	 */
+	const REST_ROUTE = '/pwa-manifest';
```
westonruter: I think `/webmanifest` is better since `.webmanifest` it is the file extension. Or rather, we can call it `/web-app-manifest` to be more accurate.

php/class-wp-app-manifest.php
```diff
@@ -0,0 +1,202 @@
...
+
+		/**
+		 * Enables overriding the manifest json.
+		 *
+		 * There are more possible values for this, including 'orientation' and 'scope.'
+		 * See the documentation: https://developers.google.com/web/fundamentals/web-app-manifest/
+		 *
+		 * @param array $manifest The manifest to send in the REST API response.
+		 */
+		return apply_filters( 'pwa_manifest_json', $manifest );
```
westonruter: This filter would probably be better named just `web_app_manifest`. It's not JSON at this point, but rather an array.

php/class-wp-app-manifest.php
```diff
@@ -0,0 +1,202 @@
...
+	 * @var string
+	 */
+	const FALLBACK_THEME_COLOR = '#fff';
+
+	/**
+	 * The REST API namespace for the manifest request.
+	 *
+	 * @var string
+	 */
+	const REST_NAMESPACE = 'wp/v2';
```
westonruter: I wonder if a different namespace should be used?

westonruter:
Interestingly in Chrome on Mac there is an “Add to homescreen” link. If I click it then I get an error:

> Site cannot be installed: the manifest does not contain a suitable icon - PNG format of at least 144px is required, the sizes attribute must be set, and the purpose attribute, if set, must include "any".

This is probably something to investigate further.

php/class-wp-app-manifest.php
kienstra: **Namespace**

Hi @westonruter,

Good question. Maybe our endpoint could be:
`web-app/v1/manifest`

Instead of the current:
`wp/v2/web-app-manifest`

It looks like `oembed` also [registers](https://github.com/WordPress/wordpress-develop/blob/a830dbcab3147e67b3076d3d15c89975df675fbb/src/wp-includes/class-wp-oembed-controller.php#L35) a different endpoint from most of the Core ones:

`oembed/1.0/embed`

oembed-routes

westonruter: Or `app/v1/web-manifest` which would better align with the `.webmanifest` file extension.

kienstra:
**Icon Type**

Hi @westonruter,

>Interestingly in Chrome on Mac there is an “Add to homescreen” link. If I click it then I get an error:

Thanks for finding that. I also see the error, as I was using a `.jpeg` as the Site Icon. But there wasn't an error when I changed the Site Icon to a `.png`.

What are the type(s) of the icons in your manifest?

I didn't know this before, but it looks like the [W3C recommends](https://www.w3.org/2005/10/howto-favicon) using `.png`, `.gif`, or `.ico` files for the favicon. As I understand it, that's the primary use case of the [Site Icon](https://make.wordpress.org/core/2015/07/27/site-icon/).

kienstra:
**Props**

We should be sure to give props [ebinion](https://github.com/ebinnion), who wrote the 2 Jetpack classes that this is based on:
* [Jetpack_PWA_Manifest](https://github.com/Automattic/jetpack/blob/40cdf9974b73b337ffe8a7693272d68a490e1b51/modules/pwa/class.jetpack-pwa-manifest.php)
* [Jetpack_PWA_Helpers](https://github.com/Automattic/jetpack/blob/40cdf9974b73b337ffe8a7693272d68a490e1b51/modules/pwa/class.jetpack-pwa-helpers.php)

php/class-wp-app-manifest.php
kienstra: Sure, PR #15 changes the endpoint to `app/v1/web-manifest`.

westonruter:
My icons consist of:

```json
[
        {
            "sizes": "192x192",
            "src": "https://src.wordpress-develop.test/wp-content/uploads/2017/11/cropped-espresso-137-1-192x192.jpg",
            "type": "image/jpeg"
        },
        {
            "sizes": "512x512",
            "src": "https://src.wordpress-develop.test/wp-content/uploads/2017/11/cropped-espresso-137-1.jpg",
            "type": "image/jpeg"
        }
]
```

Naturally this is because I selected a JPEG as the site icon. Maybe we should add a warning to the Site Icon control in the Customizer when the user selects a non-PNG image?

Issue #8, opened 2018-06-12
Work with a (or several) popular SW-enabled PWA plugin or theme, fork it, add our own code


postphotos:
As a developer, I should have a common approach to service workers to develop a collaborative approach for PWA features.

AC1: Identify a list of WordPress plugins (from #2) of a good fit for this project.
AC2: Use our prototype approach (#7) in place of their service worker code. Determine what level of effort it took, perhaps providing a pull request back to the master plugin branch _if_ the plugin is managed in a open repo.

kienstra:
**Possible Plugins**

Hi @miina,
* https://github.com/SuperPWA/Super-Progressive-Web-Apps
This looks good. It's under active development, and the [service worker registration](https://plugins.trac.wordpress.org/browser/super-progressive-web-apps/trunk/public/js/register-sw.js
) looks straight-forward.

* https://github.com/SayHelloGmbH/progressive-wordpress
This also looks good. The last commit in the repo was 5 days ago, and the author [seems to be responsive](https://github.com/SayHelloGmbH/progressive-wordpress/issues?q=is%3Aissue+is%3Aclosed) to issues opened on the repo. It looks like [this method](https://github.com/SayHelloGmbH/progressive-wordpress/blob/a3883ec51644438481d4d4afb2a5135e33304cde/Classes/class-serviceworker.php#L46) outputs the service worker script.

* https://github.com/appticles/wordpress-mobile-pack
Otherwise, this would probably be fine. but it's a little troubling that the [latest commit](https://github.com/appticles/wordpress-mobile-pack/commit/5fcc9f905d738c9636cec9f90a8aa171c4946186) to the repo was in February.

miina:
@kienstra Great, thank you for the reviewing the plugins.

Starting work on the first plugin (https://github.com/SuperPWA/Super-Progressive-Web-Apps).

miina:
Starting work on the second plugin (https://github.com/SayHelloGmbH/progressive-wordpress), work on the first plugin is on hold due to potential shortcomings of the API prototype, see [the detailed comment](https://github.com/xwp/pwa-wp/pull/14#issuecomment-400594124).

Forked repo for the second plugin: https://github.com/xwp/progressive-wordpress

miina:
PR for using the API prototype within [Progressive WordPress plugin](https://github.com/xwp/progressive-wordpress):
https://github.com/xwp/progressive-wordpress/pull/1

miina:
[...]

PR #14, opened 2018-06-21
Prototype for Service Workers API

### LINKED ISSUE #7: Prototype an API to register service workers
As a developer, I should have a discrete way to register service workers to allow my plugin (or theme) to leverage a common approach that allows for collaborative integration features.

AC1: Create documentation based on what we think the API should look like, anticipating its evolution through the project.
AC2: Based on the work we're doing in #1, demo a service worker integration that allow for other tools to also register in our common framework.

miina:
Fixes #7.

~~Currently mainly just a skeleton for registering service workers, however, it is functional.~~
For example you could add `test.js` to your plugin root directory and then call `wp_register_service_worker( 'test', '/wp-content/path-to-your-file/test.js' );` -- it should work.

Updated description:
- Allows registering static SW scripts via `wp_register_service_worker( $handle, $src, $deps = array(), $scopes = array() )` where:
  - Handle is a unique identifier for the script;
  - Source is either a callback method or path within WP_CONTENT directory;
  - Deps -- array of SW script dependencies;
  - Scopes: Array of scopes;

- Concatenates the scripts dynamically based on scopes, the following URL is used for registering:
 `/?wp_service_workers=$scope` and registers service worker per each scope. Default scope is the relative site URL;
- In case of using a source file and not a callback method: The content of SW scripts is received from the absolute path by using `$wp_filesystem->get_contents()` and not via HTTP request meaning that the content has to be static.


pwa-wp.php
```diff
@@ -41,4 +41,40 @@ function pwawp_init() {
...
+	/** WordPress Service Workers Class */
+	require PWAWP_PLUGIN_DIR . '/wp-includes/class.wp-service-workers.php';
+
+	/** WordPress Scripts Functions */
+	require PWAWP_PLUGIN_DIR . '/wp-includes/functions.wp-service-workers.php';
+
+	add_action( 'wp_print_footer_scripts', 'wp_print_service_workers' );
+
+	// Alternative for this could be in wp-includes/functions.php.
+	add_action( 'template_redirect', 'pwawp_maybe_display_sw_script' );
```
westonruter: Following the REST API, this can be done at the `parse_request` action, which is where `rest_api_loaded()` is then called.

pwa-wp.php
```diff
@@ -41,4 +41,40 @@ function pwawp_init() {
...
+
+	add_action( 'init', 'pwawp_add_sw_rewrite_rules' );
+}
+
+/**
+ * Register rewrite rules for Service Workers.
+ */
+function pwawp_add_sw_rewrite_rules() {
+	add_rewrite_tag( '%wp_service_worker%', '(0|1)' );
+	add_rewrite_tag( '%scope%', '([^&]+)' );
```
westonruter: What do you think about combining these two tags? Since the `wp_service_worker` would always get the value of `1`, could we not let the `scope` be the value of that query var?

pwa-wp.php
```diff
@@ -41,4 +41,40 @@ function pwawp_init() {
...
+	add_rewrite_rule( '^wp-service-worker.js?', 'index.php?wp_service_worker=1', 'top' );
+}
+
+/**
+ * If it's a service worker script page, display that.
+ */
+function pwawp_maybe_display_sw_script() {
+
+	if (
+		true === filter_var( get_query_var( 'wp_service_worker' ), FILTER_VALIDATE_BOOLEAN ) &&
```
westonruter: The `filter_var()` functions aren't used in core. So this can just check if `get_query_var( 'wp_service_worker' )` is not empty.

pwa-wp.php
```diff
@@ -41,4 +41,40 @@ function pwawp_init() {
...
+		strlen( get_query_var( 'scope' ) )
```
westonruter: See note above about re-using `wp_service_worker` query var for containing the `scope`.

wp-includes/functions.wp-service-workers.php
```diff
@@ -0,0 +1,78 @@
...
+ */
+function wp_print_service_workers() {
+
+	foreach ( wp_service_workers()->scopes as $scope => $value ) {
+		?>
+	<script>
+		if ( navigator.serviceWorker ) {
+			navigator.serviceWorker.register(
+				<?php echo wp_json_encode( wp_get_service_worker_url( $scope ) ); ?>,
+				{ scope: <?php echo wp_json_encode( $scope ); ?> }
```
westonruter: This would be a bit cleaner as:

```php
echo wp_json_encode( compact( 'scope' ) );
```

Eventually there may be more options we want to merge the `$scope` into.

wp-includes/functions.wp-service-workers.php
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return '?' . wp_service_workers()->query_var . '=1&scope=' . $scope;
+}
+
+/**
+ * Print service workers' scripts.
+ */
+function wp_print_service_workers() {
+
+	foreach ( wp_service_workers()->scopes as $scope => $value ) {
```
westonruter: The `$value` isn't used here. So it could just be `array_keys( wp_service_workers()->scopes )`

wp-includes/functions.wp-service-workers.php
```diff
@@ -0,0 +1,78 @@
...
+ *
+ * @param string $scope Scope, for example 'wp-admin'.
+ * @return string Service Worker URL.
+ */
+function wp_get_service_worker_url( $scope ) {
+
+	if ( get_option( 'permalink_structure' ) ) {
+		return site_url( '/', 'relative' ) . 'wp-service-worker.js?scope=' . $scope;
+	}
+	return '?' . wp_service_workers()->query_var . '=1&scope=' . $scope;
```
westonruter: Should `site_url()` not also be used as the basis of this URL?

wp-includes/functions.wp-service-workers.php
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Get service worker URL by scope.
+ *
+ * @param string $scope Scope, for example 'wp-admin'.
+ * @return string Service Worker URL.
+ */
+function wp_get_service_worker_url( $scope ) {
+
+	if ( get_option( 'permalink_structure' ) ) {
+		return site_url( '/', 'relative' ) . 'wp-service-worker.js?scope=' . $scope;
```
westonruter: Could be better as:

```php
return add_query_arg( compact( 'scope' ), site_url( '/wp-service-worker.js', 'relative' ) );
```

wp-includes/functions.wp-service-workers.php
```diff
@@ -0,0 +1,78 @@
...
+	return $registered;
+}
+
+/**
+ * Get service worker URL by scope.
+ *
+ * @param string $scope Scope, for example 'wp-admin'.
+ * @return string Service Worker URL.
+ */
+function wp_get_service_worker_url( $scope ) {
```
westonruter: The `$scope` could have a default value of `'/'`.

wp-includes/functions.wp-service-workers.php
```diff
@@ -0,0 +1,78 @@
...
+ * @param string $path   Relative path of the service worker.
+ * @param array  $deps   Optional. An array of registered script handles this depends on. Default empty array.
+ * @param string $scope  Optional Scope of the service worker.
+ * @return bool Whether the script has been registered. True on success, false on failure.
+ */
+function wp_register_service_worker( $handle, $path, $deps = array(), $scope = null ) {
+	$wp_service_workers = wp_service_workers();
+
+	// If the path is not relative. @todo should we try formatting it instead?
+	if ( ! preg_match( '/^\//', $path ) ) {
```
westonruter: The logic here can be mirrored from [`AMP_Style_Sanitizer::get_validated_url_file_path()`](https://github.com/Automattic/amp-wp/blob/70711f0cd647f233b02ffa1bbbd58496eeeea510/includes/sanitizers/class-amp-style-sanitizer.php#L441-L511), which is itself inspired by `\WP_Scripts::in_default_dir()` in core.

wp-includes/functions.wp-service-workers.php
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/**
+ * Print service workers' scripts.
+ */
+function wp_print_service_workers() {
+
+	foreach ( wp_service_workers()->scopes as $scope => $value ) {
+		?>
+	<script>
```
westonruter: This `<script>` could be moved outside the `foreach` loop so it isn't printed redundantly.

wp-includes/class.wp-service-workers.php
```diff
@@ -0,0 +1,153 @@
...
+	 *
+	 * @param string $handle Handle.
+	 * @param bool   $group Group.
+	 * @return void
+	 */
+	public function do_item( $handle, $group = false ) {
+		global $wp_filesystem;
+
+		$obj           = $this->registered[ $handle ];
+		$this->output .= $wp_filesystem->get_contents( site_url() . $obj->src ) . '
```
westonruter: Better to use `"\n"` for a newline.

westonruter: Will not using `site_url()` here result in an HTTP request?

wp-includes/class.wp-service-workers.php
```diff
@@ -0,0 +1,153 @@
...
+	 *
+	 * @param string $scope Scope of the Service Worker.
+	 */
+	public function do_service_worker( $scope ) {
+
+		header( 'Content-Type: text/javascript; charset=utf-8' );
+		header( 'Cache-Control: no-cache' );
+
+		if ( ! isset( $this->scopes[ $scope ] ) ) {
+			echo '';
```
westonruter: How about sending a 404 here via `status_header( 404 )`?

wp-includes/class.wp-service-workers.php
```diff
@@ -0,0 +1,153 @@
...
+
+	/**
+	 * Get service worker logic for scope.
+	 *
+	 * @param string $scope Scope of the Service Worker.
+	 */
+	public function do_service_worker( $scope ) {
+
+		header( 'Content-Type: text/javascript; charset=utf-8' );
+		header( 'Cache-Control: no-cache' );
```
westonruter: This could use `nocache_headers()` instead.

wp-includes/class.wp-service-workers.php
```diff
@@ -0,0 +1,153 @@
...
+
+		$this->output = '';
+		$this->do_items( $scope_items );
+
+		$file_hash = md5( $this->output );
+		header( "Etag: $file_hash" );
+
+		$etag_header = isset( $_SERVER['HTTP_IF_NONE_MATCH'] ) ? trim( $_SERVER['HTTP_IF_NONE_MATCH'] ) : false;
+		if ( $file_hash === $etag_header ) {
+			header( 'HTTP/1.1 304 Not Modified' );
```
westonruter: This can use `status_header( 304 )`

wp-includes/class.wp-service-workers.php
```diff
@@ -0,0 +1,153 @@
...
+	 * @param array  $deps   Optional. An array of registered item handles this item depends on. Default empty array.
+	 * @param bool   $ver    Always false for service worker.
+	 * @param mixed  $scope  Optional. Scope of the service worker. Default relative path.
+	 * @return bool Whether the item has been registered. True on success, false on failure.
+	 */
+	public function add( $handle, $path, $deps = array(), $ver = false, $scope = null ) {
+
+		// Set default scope if missing.
+		if ( ! $scope ) {
+			$scope = site_url( '/', 'relative' );
```
westonruter: On my system, `site_url( '/', 'relative' )` returns `https://src.wordpress-develop.test/`. Is that expected?

westonruter: Something else to take care of here is to make sure this works properly with subdirectory installs of WP. Maybe that's what the `relative` part is supposed to take care of, but I would expected that if I have a subdirectory install at `/blog/` then this is what _must_ be supplied as the base scope. We would never want a subdirectory install supplying a scope of `/`, as that would mean one subdirectory install could install a service worker for another subdirectory install, or even the network site.

wp-includes/class.wp-service-workers.php
```diff
@@ -0,0 +1,153 @@
...
+	 * Registers service worker if no item of that name already exists.
+	 *
+	 * @param string $handle Name of the item. Should be unique.
+	 * @param string $path   Path of the item relative to the WordPress root directory.
+	 * @param array  $deps   Optional. An array of registered item handles this item depends on. Default empty array.
+	 * @param bool   $ver    Always false for service worker.
+	 * @param mixed  $scope  Optional. Scope of the service worker. Default relative path.
+	 * @return bool Whether the item has been registered. True on success, false on failure.
+	 */
+	public function add( $handle, $path, $deps = array(), $ver = false, $scope = null ) {
```
westonruter: Can we just remove the `$ver` entirely?

pwa-wp.php
```diff
@@ -41,4 +41,40 @@ function pwawp_init() {
...
+function pwawp_add_sw_rewrite_rules() {
+	add_rewrite_tag( '%wp_service_worker%', '(0|1)' );
+	add_rewrite_tag( '%scope%', '([^&]+)' );
+	add_rewrite_rule( '^wp-service-worker.js?', 'index.php?wp_service_worker=1', 'top' );
+}
+
+/**
+ * If it's a service worker script page, display that.
+ */
+function pwawp_maybe_display_sw_script() {
```
westonruter: This function can go into `wp-includes/service-workers.php` functions file. It could be named `service_worker_loaded()` for parity with `rest_api_loaded()`, and again it can run at `parse_request` action.

pwa-wp.php
```diff
@@ -41,4 +41,40 @@ function pwawp_init() {
...
+
+	/** WordPress Scripts Functions */
+	require PWAWP_PLUGIN_DIR . '/wp-includes/functions.wp-service-workers.php';
+
+	add_action( 'wp_print_footer_scripts', 'wp_print_service_workers' );
+
+	// Alternative for this could be in wp-includes/functions.php.
+	add_action( 'template_redirect', 'pwawp_maybe_display_sw_script' );
+
+	add_action( 'init', 'pwawp_add_sw_rewrite_rules' );
```
westonruter: Let's create a new file called `wp-includes/default-filters.php` where we put all of the filter and action hooks in the same way they are done in core. This will make it easier for a merge.

pwa-wp.php
```diff
@@ -41,4 +41,40 @@ function pwawp_init() {
...
+}
+
+/**
+ * Register rewrite rules for Service Workers.
+ */
+function pwawp_add_sw_rewrite_rules() {
```
westonruter: Per below, let's put this in `wp-includes/service-workers.php`.

wp-includes/class.wp-service-workers.php
```diff
@@ -0,0 +1,153 @@
...
+	 * @var string
+	 */
+	public $query_var = 'wp_service_worker';
+
+	/**
+	 * Array of scopes.
+	 *
+	 * @var array
+	 */
+	public $scopes = array();
```
westonruter: Is this `$scopes` needed? There is already `scope` data attached to each registered service worker, so then when getting the scripts to print when rendering the service worker it can just iterate over all the scripts and gather up the matching scope. If there are none matching, then return nothing with a 404.

By eliminating the var, then we don't have to keep track of the scopes in two different places, such as when a service worker script is unregistered.

wp-includes/class.wp-service-workers.php
westonruter: See also note above about whether we need this additional `scopes` member var.

wp-includes/class.wp-service-workers.php
```diff
@@ -0,0 +1,153 @@
...
+
+		$etag_header = isset( $_SERVER['HTTP_IF_NONE_MATCH'] ) ? trim( $_SERVER['HTTP_IF_NONE_MATCH'] ) : false;
+		if ( $file_hash === $etag_header ) {
+			header( 'HTTP/1.1 304 Not Modified' );
+			exit;
+		}
+
+		// @codingStandardsIgnoreLine
+		echo $this->output;
+		exit;
```
westonruter: I think the `exit` can be removed from here (and above) in favor of doing it in what is currently named `pwawp_maybe_display_sw_script()` (which I proposed could be named `service_worker_loaded()`).  This would be similar to how `rest_api_loaded()` works and it would allow for this method to be unit-tested.

wp-includes/class.wp-service-workers.php
```diff
@@ -0,0 +1,153 @@
...
+		}
+		return true;
+	}
+
+	/**
+	 * Get service worker logic for scope.
+	 *
+	 * @param string $scope Scope of the Service Worker.
+	 */
+	public function do_service_worker( $scope ) {
```
westonruter: This could be better named `serve_request` for parity with `\WP_REST_Server::serve_request()`. See `rest_api_loaded()`.

wp-includes/class.wp-service-workers.php
miina: True. I was originally thinking of using only `$scopes` instead of using handles, however, this way we wouldn't be able to leverage the existing `do_items()` method which takes care of the `$deps` as well.

pwa-wp.php
```diff
@@ -41,4 +41,40 @@ function pwawp_init() {
 
 	$wp_web_app_manifest = new WP_Web_App_Manifest();
 	$wp_web_app_manifest->init();
+
+	// These could be in ABSPATH . WPINC . '/script-loader.php' file.
+	/** WordPress Service Workers Class */
+	require PWAWP_PLUGIN_DIR . '/wp-includes/class.wp-service-workers.php';
+
+	/** WordPress Scripts Functions */
+	require PWAWP_PLUGIN_DIR . '/wp-includes/functions.wp-service-workers.php';
```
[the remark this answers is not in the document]
miina: Changing the filename. However, I was following the naming conventions of `WP_Scripts` instead since the logic is similar. Isn't it be confusing to reuse the methods and logic of `WP_Scrips` and at the same time follow the naming conventions and hierarchy of REST API? Or are we generally trying to follow the logic of the REST API and can expect more similarities to the REST API? Or perhaps the `WP_Scripts` and `WP_Styles` naming conventions are to be changed at some point, too ( e.g. `class-*` instead of `class.*`)? Thoughts?

pwa-wp.php
miina: To confirm if I understand correctly:
Do you mean that we could ignore `wp_service_worker` as a separate query var since it's always `1` and just use `scope`, for example by using the following rewrite rule instead and just adding a rewrite tag for `scope`?
`add_rewrite_rule( '^wp-service-worker.js\\?scope=([^&]+)?', 'index.php?scope=$matches[1]', 'top' );`

Added the `wp_service_worker` tag separately since in case the pretty permalinks aren't enabled then we'd need to check the `wp_service_worker=1` separately, otherwise the logic will always run if `$_GET['scope']` is set, even if the `wp_service_worker` is not.

Thoughts?

(Likely that you meant something else by the comment.)

miina: Perhaps you meant the opposite -- that we could just use `wp_service_worker`, for example:
`add_rewrite_tag( '%wp_service_worker%', '([^&]+)' );` where scope is the value of `wp_service_worker` and use 
`add_rewrite_rule( '^wp-service-worker.js\\?scope=([^&]+)?', 'index.php?wp_service_worker=$matches[1]', 'top' );`?

In this case as well in case of non-pretty permalinks this wouldn't work with the current registering URL structure, unless we would use the return URL as `?wp_service_worker=$scope` for the return URL in case of non-pretty permalinks. Thoughts?

wp-includes/class.wp-service-workers.php
miina: We could if we would create a function with different name, otherwise it wouldn't be compatible with the parent method (`WP_Dependencies::add`). Do you think it would be better to create a separate method? Thoughts?

wp-includes/class.wp-service-workers.php
miina: Hmm, yes, that's what `site_url( '/', 'relative' )` was supposed to take care of, it's not excepted that it's returning the full URL. In `wp-includes/link-template.php` this line should get the relative path:
`$url = ltrim( preg_replace( '#^\w+://[^/]*#', '', $url ) );`.

For example locally running `https://src.wordpress-develop.test/blog` through the method returns `/blog` and `https://src.wordpress-develop.test/` returns `/`.

pwa-wp.php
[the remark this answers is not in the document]
miina: Guessing that following the REST API naming conventions is relevant since with Service Workers (as with REST API) we're sending a HTTP response not just managing the scripts as in case of `WP_Script`.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,216 @@
...
+
+		$file_path = null;
+		if ( 0 === strpos( $url, $content_url ) ) {
+			$file_path = WP_CONTENT_DIR . substr( $url, strlen( $content_url ) - 1 );
+		} elseif ( 0 === strpos( $url, $includes_url ) ) {
+			$file_path = ABSPATH . WPINC . substr( $url, strlen( $includes_url ) - 1 );
+		} elseif ( 0 === strpos( $url, $admin_url ) ) {
+			$file_path = ABSPATH . 'wp-admin' . substr( $url, strlen( $admin_url ) - 1 );
+		} else {
+			$file_path = ABSPATH . substr( $url, strlen( $this->remove_url_scheme( $base_url ) ) );
```
miina: Also added the `else` here for now.

Wondering if we should restrict the SW file path to validating against `WP_CONTENT_DIR` only? Then the potential default core WP SW scripts could be registered via `WP_Service_Workers::register_default_service_workers` which would skip the check for `get_validated_file_path`. Not sure if plugins/themes should be able to register scripts that are outside of `WP_CONTENT_DIR`. Thoughts?

wp-includes/service-workers.php
```diff
@@ -0,0 +1,121 @@
...
+ * @return string Service Worker URL.
+ */
+function wp_get_service_worker_url( $scope ) {
+
+	if ( ! $scope ) {
+		$scope = site_url( '/', 'relative' );
+	}
+
+	if ( get_option( 'permalink_structure' ) ) {
+		return add_query_arg( compact( 'scope' ), site_url( '/wp-service-worker.js', 'relative' ) );
```
miina: Do we actually need a separate case and rewrite for pretty permalinks at all? This URL wouldn't actually be directly used anyway in browser. Thoughts?

wp-includes/service-workers.php
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return add_query_arg( array(
+		'scope'             => $scope,
+		'wp_service_worker' => 1,
+	), site_url( '/', 'relative' ) );
```
miina: Was testing and debugging it more locally with VVV and it seems to always return the correct relative path in case of `https://src.wordpress-develop.test`. This is how I see the added SW:
screen shot 2018-06-26 at 3 38 56 pm
with
screen shot 2018-06-26 at 3 39 27 pm
@westonruter Are you still seeing the full URL instead?

pwa-wp.php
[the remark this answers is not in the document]
westonruter: The REST API is newer to core, whereas the Dependencies system is older. So generally the newer stuff has better naming conventions than the older stuff. That's why I looked to the REST API for precedence.

pwa-wp.php
westonruter: Yes, I meant the latter, just using the one `wp_service_worker` query var. If pretty permalinks are disabled, then the URL would be `/?wp_service_worker=$scope`.

wp-includes/class.wp-service-workers.php
westonruter: I think I only tested using WP-CLI. If I get `site_url( '/', 'relative' )` in WordPress normally it is successfully returning just `/` 👍

wp-includes/class.wp-service-workers.php
westonruter: Oh yeah. PHP will issue a notice about the function signatures not matching, correct? Alas the problem with positional parameters.

We could introduce a `register` method instead which would be the only method used instead of `add` or `enqueue`. I don't think it makes sense to ever add a service worker script and then not have it be enqueued ever, right? Setting the `scope` is really the way that you are enqueueing it.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,216 @@
...
+	 * Registers service worker if no item of that name already exists.
+	 *
+	 * @param string $handle Name of the item. Should be unique.
+	 * @param string $path   Path of the item relative to the WordPress root directory.
+	 * @param array  $deps   Optional. An array of registered item handles this item depends on. Default empty array.
+	 * @param bool   $ver    Always false for service worker.
+	 * @param mixed  $scope  Optional. Scope of the service worker. Default relative path.
+	 * @return bool Whether the item has been registered. True on success, false on failure.
+	 */
+	public function add( $handle, $path, $deps = array(), $ver = false, $scope = null ) {
```
westonruter: Something that just came to mind: what if a script needs to be “enqueued” in multiple scopes?

wp-includes/class-wp-service-workers.php
westonruter: Since plugins and themes would only enqueue their scripts that are inside `WP_CONTENT_DIR` I think it is safe to assume that the same would be true for service worker scripts.

wp-includes/service-workers.php
westonruter: You're right. If we can avoid the rewrite rules then we can save headaches. I think it is fine to just use the non-pretty URLs.

wp-includes/service-workers.php
westonruter: Per above, it's because I was using WP-CLI and not WP normally 😊

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,216 @@
...
+
+		// Set default scope if missing.
+		if ( ! $scope ) {
+			$scope = site_url( '/', 'relative' );
+		}
+
+		if ( false === parent::add( $handle, $path, $deps, false, $scope ) ) {
```
westonruter: Instead of passing `$scope` as the args, I think it should pass `compact( 'scope' )`. This will allow other args to be added in the future in addition to the scope.

Also, maybe a script should be able to added to multiple scopes as I mentioned above, so maybe it should be `scopes` instead.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,216 @@
...
+	/**
+	 * Get all scopes.
+	 *
+	 * @return array Array of scopes.
+	 */
+	public function get_scopes() {
+
+		$scopes = array();
+		foreach ( $this->registered as $handle => $item ) {
+			$scopes[] = $item->args;
```
westonruter: Per above, I think we should consider other args being supplied and for multiple scopes to be registered for one script. So this could instead be:

```php
$scopes = array_merge( $scopes, $item->args['scopes'] );
```

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,216 @@
...
+		if ( ! in_array( $scope, $this->get_scopes(), true ) ) {
+			status_header( 404 );
+			return;
+		}
+
+		$scope_items = array();
+
+		// Get handles from the relevant scope only.
+		foreach ( $this->registered as $handle => $item ) {
+			if ( $scope === $item->args ) {
```
westonruter: Per above, I think `args` should be preserved as an array to allow additional args in the future. Also, maybe a script should be in multiple scopes. So in that case, this could be:

```php
if ( in_array( $item->args['scopes'], $scope, true ) ) {
```

wp-includes/class-wp-service-workers.php
miina: Just pushed the changes for registering scopes as an array.

pwa-wp.php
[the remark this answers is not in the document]
miina: Good to know, thanks for clarifying.

miina:
@westonruter Started looking into modifying [the forked repo](https://github.com/xwp/Super-Progressive-Web-Apps) of Super Progressive Web Apps and ran into some issues right away with the current API implementation:
- The service worker logic includes dynamic data coming from the user-configured settings, this data is added using PHP, there is no static SW logic, see [here](https://github.com/xwp/Super-Progressive-Web-Apps/blob/master/public/sw.php#L95-L112) for example.
- There is no actually existing SW JS file to register, the logic is dynamically created at `[SITE_URL]/superpwa-sw.js`.

I recall seeing other plugins using similar logic for registering SW in case of dynamic data.

Currently the content of the registered SW static files is taken by the file path. One option could be that we'd still use the URL instead of the path [here](https://github.com/xwp/pwa-wp/blob/7c28ba465cbcc3eab0f82c0c378b93376922abb5/wp-includes/class-wp-service-workers.php#L150) and would be able to get the dynamic content via HTTP request. Alternatively we could add a method that allows registering the SW logic via PHP method and adds the output of the method to the final output.

Thoughts?

westonruter:
@miina Thanks for raising this question about dynamically-generated JS. I suppose we need something like `wp_add_inline_script()` for service workers, except there needs to be a way to add the script without it being attached to any other script. 

It seems we need to be able to register a service worker that either points to a file _or_  the JS contents are provided for you instead. I believe this could be modeled in `WP_Scripts` by defining a dependency with a `src` of `false` and then adding inline script data via `wp_add_inline_script()`. But there would need to be a better API than this for registering the two kinds of service workers sources (JS file path and JS strings).

I think we should definitely avoid making HTTP requests to fetch the scripts for the service worker. 

Maybe there should be a `wp_register_service_worker_file()` which you provide a file path and a `wp_register_service_worker_script()` which you provide a JS script? I don't like that either. What about switching to use an args array instead?

```php
wp_register_service_worker( $handle, array(
    'scopes' => array( /* ... */ ),
    'deps' => array( /* ... */ ),

    'path' => $path_to_file,
    // or
    'script' => '// JS goes here',
) )
```

westonruter:
Alternatively, I see that the Mozilla Service Worker Manager plugin uses:

```php
$swmgr->sw()->add_content( $callback_function );
```

Where the `$callback_function` then `echo`es the JS, and it is up to the callback to decide whether to get the file from elsewhere on disk or to supply it inline. I like this. I think it would be better to `return` rather than `echo`, however. In that way, it could be like a `render` callback in the REST API. 

So then the function signature could be:

```php
wp_register_service_worker( $handle, $callback, $deps, $scopes );
```

miina:
An example of implementation for allowing both static scripts + callback: https://github.com/xwp/pwa-wp/pull/14/commits/645dc19d5c44f2e7f97545eccedd8ec57eb35043

westonruter:
@miina that's a good idea: `$source` could either be a `callable` or a `string`.

wp-includes/service-workers.php
```diff
@@ -0,0 +1,112 @@
...
+	return $wp_service_workers;
+}
+
+/**
+ * Register a new service worker.
+ *
+ * @since ?
+ *
+ * @param string $handle   Name of the service worker. Should be unique.
+ * @param string $callback Callback method or relative path of the service worker.
```
westonruter: The type here can be `string|callable` with the param then being `$src`, where if it `is_callable()` then the return value is then used as the _source_ when printing.

wp-includes/service-workers.php
```diff
@@ -0,0 +1,112 @@
...
+ * @param array  $deps     Optional. An array of registered script handles this depends on. Default empty array.
+ * @param array  $scopes   Optional Scopes of the service worker.
+ * @return bool Whether the script has been registered. True on success, false on failure.
+ */
+function wp_register_service_worker( $handle, $callback, $deps = array(), $scopes = array() ) {
+	$service_workers = wp_service_workers();
+
+	// If the path is not correct.
+	if ( ! is_callable( $callback ) && is_wp_error( $service_workers->get_validated_file_path( $callback ) ) ) {
+		_doing_it_wrong(
```
westonruter: This condition would probably be better moved to `WP_Service_Workers::register()`, no?

wp-includes/service-workers.php
```diff
@@ -0,0 +1,112 @@
...
+ * @param string $callback Callback method or relative path of the service worker.
+ * @param array  $deps     Optional. An array of registered script handles this depends on. Default empty array.
+ * @param array  $scopes   Optional Scopes of the service worker.
+ * @return bool Whether the script has been registered. True on success, false on failure.
+ */
+function wp_register_service_worker( $handle, $callback, $deps = array(), $scopes = array() ) {
+	$service_workers = wp_service_workers();
+
+	// If the path is not correct.
+	if ( ! is_callable( $callback ) && is_wp_error( $service_workers->get_validated_file_path( $callback ) ) ) {
```
westonruter: What if someone passes a non-string value as `$callback`, like an `array( $class_instance, 'not_existing_method' )`? I think that an `is_string()` check should be used before passing the arg to `get_validated_file_path`.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,212 @@
...
+		do_action_ref_array( 'wp_default_service_workers', array( &$this ) );
+	}
+
+	/**
+	 * Register service worker.
+	 *
+	 * Registers service worker if no item of that name already exists.
+	 *
+	 * @param string $handle   Name of the item. Should be unique.
+	 * @param string $callback Path of the item relative to the WordPress root directory.
```
westonruter: As noted below, this should be:

```
@param string|callable $src URL to the source in the WordPress install, or a callback that returns the JS to include in the service worker.
```

And so it's not actually _relative_. It can be an absolute URL.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,212 @@
...
+	 */
+	public function do_item( $handle, $group = false ) {
+		global $wp_filesystem;
+
+		$obj = $this->registered[ $handle ];
+
+		if ( false === $obj->src ) {
+			$this->output .= call_user_func( $obj->args['callback'] ) . "\n";
+		} else {
+			$this->output .= $wp_filesystem->get_contents( $this->get_validated_file_path( $obj->src ) ) . "\n";
```
westonruter: It would be nice for debugging if this was preceded by something like: `$this->output .= sprintf( "\n/* Source: %s */", $obj->src );`

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,212 @@
...
+	 * @param string $handle Handle.
+	 * @param bool   $group Group.
+	 * @return void
+	 */
+	public function do_item( $handle, $group = false ) {
+		global $wp_filesystem;
+
+		$obj = $this->registered[ $handle ];
+
+		if ( false === $obj->src ) {
```
westonruter: Maybe better to check `is_callable( $obj->args['callback'] )` here instead.

wp-includes/class-wp-service-workers.php
westonruter: Instead of doing the `_doing_it_wrong()` check when registering, why not just do it here when printing? Then there would only be one `get_validated_file_path` call per registered file. The error could be printed into the script in comments as well, maybe in a `console.warn()` even.

westonruter:
This section of the Mozilla plugin's readme seems important to also account for: https://github.com/mozilla/wp-sw-manager#accessing-service-worker-registration

There needs to be a way to capture the `Promise` returned by `navigator.serviceWorker.register()` for the various scopes so that other scripts on the page can reference it. The Mozilla plugin achieves this by enqueueing an [external script](https://github.com/mozilla/wp-sw-manager/blob/master/lib/js/sw-registrar.js) and so that other scripts can then mark it as a dependency. Having an external script is not ideal. Maybe there could be a “dummy” service worker registration script with a handle like `service-worker-registration` but then `src` of `false`. Then the current logic in `wp_print_service_workers()` could be moved to a `wp_add_inline_script( 'service-worker-registration', $contents_of_wp_print_service_workers, 'before' );` Then any other script could add a script that depends on the `service-worker-registration` handle and it will be assured to get printed afterward.

wp-includes/service-workers.php
miina: Added the check for `is_string()` to `get_validated_file_path`, this way the `WP_Error` message will also reflect the reason for failing.

miina:
Started looking into using the API in case of [WordPress Mobile Pack](https://github.com/appticles/wordpress-mobile-pack) plugin: Looks like the plugin is using a [3rd party library](https://github.com/xwp/wordpress-mobile-pack/blob/master/frontend/sections/notification-banner/lib/noty.js) for the notifications logic. Registering a Service Worker seems to happen within Noty object and it's registered only in case of some fulfilled conditions:
```
{
    key: 'requestSubscription',
    value: function requestSubscription() {
    ...
      if (result === 'granted') {
          _this2.fire('onPermissionGranted');

          if ('serviceWorker' in navigator) {
            navigator.serviceWorker.register(_this2.workerPath).then(function () {
              navigator.serviceWorker.ready.then(function (serviceWorkerRegistration) {
                ...
              });
            });
          }
      }
}
```
Any thoughts on what could be the best approach for these cases for the Service Workers API Prototype -- when the included SW logic comes from a 3rd party library and is not so easily separable?

miina:
>This section of the Mozilla plugin's readme seems important to also account for: https://github.com/mozilla/wp-sw-manager#accessing-service-worker-registration

Definitely! Will start looking into this now.

westonruter:
> Also realized that when concatenating different scripts together then the declared constants might get duplicated, for example offlinePage. Do you think wrapping the pieces in self-invoking functions as in 09506d1 could be a suitable approach?

@miina Yes, that's probably a good idea.

> Any thoughts on what could be the best approach for these cases for the Service Workers API Prototype -- when the included SW logic comes from a 3rd party library and is not so easily separable?

That's an interesting one. I am half thinking that this conditional installation of the service worker shouldn't be supported for the core API. It makes sense in their case that the service worker would only get installed if the permissions are set, but in core there would be other scripts registered and so the service worker would need to get installed regardless. In that case the JS on the page should send a message to the service worker once `onPermissionGranted` happens.

In this way it would be similar to how WordPress handles enqueueing scripts: there is no facility for conditionally enqueueing scripts in response to some client-side interaction. If a script gets enqueued then it gets printed, and it is up to the script to look at its environment to check when/if it should do something. I think the same applies for service workers.

miina:
> Maybe there could be a “dummy” service worker registration script with a handle like service-worker-registration but then src of false. Then the current logic in wp_print_service_workers() could be moved to a wp_add_inline_script( 'service-worker-registration', $contents_of_wp_print_service_workers, 'before' );

This would work well, however, looks like `wp_add_inline_script` is always called with `$echo = false`:
https://github.com/xwp/wordpress-develop/blob/master/src/wp-includes/class.wp-scripts.php#L263-L264

And it's eventually printed only if the `$src` is set, too, otherwise just returns before printing:
https://github.com/xwp/wordpress-develop/blob/master/src/wp-includes/class.wp-scripts.php#L315

Wondering if a filter could be added, for example something like:
```
if ( ! $obj->src ) {
    $echo = apply_filters( "$handle_echo_inline_script", false );
}
```

Alternatively we could use `wp_scripts()->print_inline_script( 'service-worker-registration', 'before', true );` directly but this wouldn't assure the service worker getting registered before the dependent script loaded. In theory we could use `wp_print_footer_scripts` with priority `9` for printing the script and always enqueueing the dependent script within footer, for example:
```
foreach ( array( 'wp_print_footer_scripts', 'admin_print_footer_scripts', 'customize_controls_print_footer_scripts' ) as $filter ) {
	add_filter( $filter, 'wp_print_service_workers', 9 );
}

function wp_print_service_workers() {
	wp_scripts()->print_inline_script( 'service-worker-registration', 'before', true );
}

// In plugin's `wp_enqueue_script` action.
wp_enqueue_script( 'test-script', plugin_dir_url( PWAWP_PLUGIN_FILE ) . '/test.js', array( 'service-worker-registration' ), false, true );
```

But this doesn't seem like a good option -- the user of the API would always have to use `true` for the footer, the dependencies wouldn't be set via the default logic of `WP_Scripts`.

Thoughts?

westonruter:
> And it's eventually printed only if the $src is set, too, otherwise just returns before printing

I was afraid of this. I had a hunch this was what might happen, but I was hoping it wasn't. I have rarely added a script with a `false` src, as I assume is true for many devs, so I'm not surprised inline scripts don't really work with it.

I wonder if before printing scripts we could check if any of them have a dependency on the service worker handle, and if so, add the service worker registration as the inline script right before the first dependency that requires it? Otherwise, the service worker registration would just be output at `wp_print_footer_scripts`? That doesn't seem the cleanest.

Maybe we're just making it more difficult than it needs to be. If we force the service workers to always be registered before other scripts are printed, then that would have the same effect as having an explicit dependency linking the two, no? We may want to register the service workers in the `wp_head` in that case before _any_ scripts get printed.

miina:
> We may want to register the service workers in the wp_head in that case before any scripts get printed.

Why not, we could for now just register the service workers in the `wp_head` and that would have the same effect.
However, this will also ensure the registered SW will be the lowest priority compared to any other potential service workers registered later. Perhaps it's good to have consistent behavior, hah.

miina:
Made the changes to add the SW registering to `wp_head` instead and saving the promises within 4cc95ca.

wp-includes/service-workers.php
```diff
@@ -0,0 +1,104 @@
...
+	if ( empty( $scopes ) ) {
+		return;
+	}
+	?>
+	<script>
+		<?php
+		foreach ( $scopes as $scope ) {
+			?>
+			if ( navigator.serviceWorker ) {
+				if ( ! window.$serviceWorkersRegister ) {
```
westonruter: This variable should perhaps get namespaces under the `wp` object. For example, `wp.serviceWorkerRegistrations`.

wp-includes/service-workers.php
```diff
@@ -0,0 +1,104 @@
...
+ * Register rewrite tag for Service Workers.
+ */
+function wp_add_sw_rewrite_tags() {
+	add_rewrite_tag( '%wp_service_worker%', '([^&]+)' );
+}
+
+/**
+ * If it's a service worker script page, display that.
+ */
+function service_worker_loaded() {
```
westonruter: This function should get prefixed by `wp_`

wp-includes/service-workers.php
```diff
@@ -0,0 +1,104 @@
...
+		?>
+	</script>
+	<?php
+}
+
+/**
+ * Register rewrite tag for Service Workers.
+ */
+function wp_add_sw_rewrite_tags() {
+	add_rewrite_tag( '%wp_service_worker%', '([^&]+)' );
```
westonruter: We don't need the rewrite tag anymore, do we?

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,224 @@
...
+					sprintf( esc_html__( 'Service worker src is incorrect: %s', 'pwa' ), esc_html( $obj->src ) ),
+					'0.1'
+				);
+
+				/* translators: %s is file URL */
+				$this->output .= "console.warn( '" . sprintf( esc_html__( 'Service worker src is incorrect: %s', 'pwa' ), esc_html( $obj->src ) ) . "' );\n";
+			} else {
+				/* translators: %s is file URL */
+				$this->output .= sprintf( esc_html( "\n/* Source: %s */\n" ), esc_url( $obj->src ) );
+				$this->output .= "( function() {\n";
```
westonruter: On second thought, this IIFE may cause confusion since it is a different behavior than WordPress normally handles scripts. If a service worker script doesn't want to conflict with others, than it shod wrap itself in an IIFE, no?

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,224 @@
...
+	 */
+	public function __construct() {
+		parent::__construct();
+		global $wp_filesystem;
+
+		if ( ! class_exists( 'WP_Filesystem' ) ) {
+			require_once ABSPATH . '/wp-admin/includes/file.php';
+		}
+
+		if ( null === $wp_filesystem ) {
```
westonruter: Would this be better in `do_items`?

wp-includes/service-workers.php
miina: Sounds good, however, just in case reminding that the idea for the name of the function came from `rest_api_loaded()`.

westonruter: True. I think it should have been prefixed in core as well.

wp-includes/service-workers.php
miina: Without this checking `if ( ! empty( $GLOBALS['wp']->query_vars['wp_service_worker'] ) ) {` probably wouldn't work. Thoughts?

wp-includes/service-workers.php
miina: Would this also mean that we should wrap registering the service worker with ` window.addEventListener('load', function() {` for `wp` to be available? Actually according to [this document](https://developers.google.com/web/fundamentals/primers/service-workers/registration) this is the best practice anyway.

wp-includes/class-wp-service-workers.php
miina: Makes sense!

wp-includes/class-wp-service-workers.php
miina: `do_items` is in the parent class only but will move it to `do_item`.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,224 @@
...
+	 * Registers service worker if no item of that name already exists.
+	 *
+	 * @param string          $handle Name of the item. Should be unique.
+	 * @param string|callable $src    URL to the source in the WordPress install, or a callback that returns the JS to include in the service worker.
+	 * @param array           $deps   Optional. An array of registered item handles this item depends on. Default empty array.
+	 * @param array           $scopes Optional. Scopes of the service worker. Default relative path.
+	 * @return bool Whether the item has been registered. True on success, false on failure.
+	 */
+	public function register( $handle, $src, $deps = array(), $scopes = array() ) {
+
```
amedina: Should we check here for scope collisions? That is, what happens if two entities call register on the same scope?

wp-includes/service-workers.php
westonruter: I was thinking you'd just define the object of it isn't available yet. I believe that is what all scripts that interact with the `wp` global do. In other words, `if ( ! window.wp ) { window.wp = {}; }`

wp-includes/service-workers.php
westonruter: I think then you just need to register the query var alone, via the `query_vars` filter. Just amend the query var to the array that is passed.

wp-includes/class-wp-service-workers.php
westonruter: It is by design that two scripts can have the same scope. In fact this is the reason for this core API in the first place, to allow plugins and themes to each register their own scripts to run in the same scope as others'.

wp-includes/service-workers.php
westonruter: Hummm, if it is a best practice to wait until `load` then how would other scripts get access to the registration promises?

miina: Wouldn't waiting until `load` within other scripts as well work?

amedina:
Great progress! We should discuss some of the fundamental aspects of the implementation in the context of the Service Worker API Spec : https://w3c.github.io/ServiceWorker/

wp-includes/service-workers.php
westonruter: If this script runs first then I suppose so, yes. It would calling the `load` event handlers in the order they were registered, which is the spec'ed behavior I believe.

The drawback here I see being that it would require every script that wants to access the `ServiceWorkerRegistration` object would have to add a `load` event handler. Maybe that is not a bad idea. I suppose we could require that now. If we decide otherwise, then an alternative could be to create our own `Promise` which then, after `load` is happens and we call `navigator.serviceWorker.register()` and it resolves, we can then resolve the original promise we created during page load with the registration object that get get later. But that is perhaps overkill here.

westonruter: 👉  Also, I wonder if we actually need `wp.serviceWorkerRegistrations` or if it is redundant. I see actually that there is [`navigator.serviceWorker.getRegistration()`](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/getRegistration) which seems like a built-in way to get the same thing?

miina: Made the other changes and added waiting until `load` for now -- considering that we might change it if it becomes necessary in the future.

wp-includes/service-workers.php
```diff
@@ -0,0 +1,111 @@
...
+	if ( empty( $scopes ) ) {
+		return;
+	}
+	?>
+	<script>
+		<?php
+		foreach ( $scopes as $scope ) {
+			?>
+			if ( navigator.serviceWorker ) {
+				window.addEventListener('load', function() {
```
westonruter: I think the `if` statement and `addEventListener` call should be placed outside the `foreach` loop, as otherwise the code inside will get called multiple times unnecessarily. So like:

```php
if ( navigator.serviceWorker ) {
    window.addEventListener('load', function() {
        <?php foreach ( $scopes as $scope ) : ?>
...
```

wp-includes/service-workers.php
```diff
@@ -0,0 +1,111 @@
...
+	<?php
+}
+
+/**
+ * Register query var.
+ *
+ * @param array $query_vars Query vars.
+ * @return array Query vars.
+ */
+function wp_add_sw_query_vars( $query_vars ) {
```
westonruter: Probably better named as `wp_add_service_worker_query_var()`. I don't think we need to abbreviate, since the other instances of service worker aren't abbreviated.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,221 @@
...
+	public $output = '';
+
+	/**
+	 * WP_Service_Workers constructor.
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		if ( ! class_exists( 'WP_Filesystem' ) ) {
+			require_once ABSPATH . '/wp-admin/includes/file.php';
```
westonruter: This can be moved into `do_item` as well, and then the `__construct` method can be eliminated.

wp-includes/service-workers.php
miina: >Also, I wonder if we actually need wp.serviceWorkerRegistrations or if it is redundant. I see actually that there is navigator.serviceWorker.getRegistration() which seems like a built-in way to get the same thing?

Good point, testing.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,221 @@
...
+		$file_hash = md5( $this->output );
+		header( "Etag: $file_hash" );
+
+		$etag_header = isset( $_SERVER['HTTP_IF_NONE_MATCH'] ) ? trim( $_SERVER['HTTP_IF_NONE_MATCH'] ) : false;
+		if ( $file_hash === $etag_header ) {
+			status_header( 304 );
+			return;
+		}
+
+		// @codingStandardsIgnoreLine
```
westonruter: Instead of this, let's do:

```php
echo $this->output; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,221 @@
...
+					sprintf( esc_html__( 'Service worker src is incorrect: %s', 'pwa' ), esc_html( $obj->src ) ),
+					'0.1'
+				);
+
+				/* translators: %s is file URL */
+				$this->output .= "console.warn( '" . sprintf( esc_html__( 'Service worker src is incorrect: %s', 'pwa' ), esc_html( $obj->src ) ) . "' );\n";
+			} else {
+				/* translators: %s is file URL */
+				$this->output .= sprintf( esc_html( "\n/* Source: %s */\n" ), esc_url( $obj->src ) );
+				$this->output .= $wp_filesystem->get_contents( $this->get_validated_file_path( $obj->src ) ) . "\n";
```
westonruter: I wonder actually if we need to use `$wp_filesystem->get_contents()` and shouldn't just use `file_get_contents()`? Core actually just uses `file_get_contents()` when it needs to get a file from the filesystem, for example:

https://github.com/WordPress/wordpress-develop/blob/97cacadc50716695d8c8d3afe129d901f4230266/src/wp-includes/class-wp-theme.php#L1133-L1138

I don't see any core uses of calls to `$wp_filesystem->get_contents()` at all.

miina: Saw that as well, wasn't sure if it's because the `@file_get_contents` was in use already before the `WP_Filesystem` or if it's just OK for core.

If that's OK within core then we can just remove using the `WP_Filesystem` and add ignore for the coding standards.

westonruter: Yes, let's do that. The only reason for the coding standards warning is the usage of this function to make HTTP requests. Since we are validating the path is local to the filesystem, the coding standards warning does not apply.

wp-includes/service-workers.php
miina: Looks like you're right -- the promise can be accessed via `navigator.serviceWorker.getRegistration()` instead.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,206 @@
...
+				);
+
+				/* translators: %s is file URL */
+				$this->output .= "console.warn( '" . sprintf( esc_html__( 'Service worker src is incorrect: %s', 'pwa' ), esc_html( $obj->src ) ) . "' );\n";
+			} else {
+				/* translators: %s is file URL */
+				$this->output .= sprintf( esc_html( "\n/* Source: %s */\n" ), esc_url( $obj->src ) );
+
+				// @codingStandardsIgnoreLine
+				$this->output .= @file_get_contents( $this->get_validated_file_path( $obj->src ) ) . "\n";
```
miina: This causes 4 different warnings, added `// @codingStandardsIgnoreLine`.

nico-martin:
Hi there, I'm the author of the plugin "Progressive WordPress".

I just went through all your comments and your code and I'm really amazed about the progress and the decissions you made. Here are some thoughts:

As proposed I think it would be awesome to either pass a callback or a file to `wp_register_service_worker`. This way its a lot more flexible.

Yes, there could be a conflict between different constants with the same name when mergin the scripts toghether. But I don't think thats a big problem. As always, we as plugin or theme devs need to prefix such things.

westonruter:
@nico-martin Thanks! What do you think would make sense for a next step to validate the implementation here? What comes to mind is that you could create a branch of your plugin to try plugging in the core API proposed by this plugin. If all of the use cases of your plugin are accounted for by this PR, then I'd say it's good to merge and we can move on to other plugins to verify it accounts for their use cases as well.

miina:
>What comes to mind is that you could create a branch of your plugin to try plugging in the core API proposed by this plugin.

@nico-martin @westonruter Just for information that here is a PR that starts the implementation by registering the service worker of the plugin Progressive Wordpress almost as it is (with some small modifications such as wrapping the SW logic into IIFE): https://github.com/xwp/progressive-wordpress/pull/1.

nico-martin:
@miina yes, I've seen your implementations and I will try them.
@westonruter yes, I'll be happy to run some test

nico-martin:
Additionally to my comment on #8 I did run some tests with the API: https://github.com/SayHelloGmbH/progressive-wordpress/blob/pwawp/Classes/class-serviceworker.php#L39
It looks like everything works as expected.

wp-includes/class-wp-service-workers.php
```diff
@@ -0,0 +1,205 @@
...
+	 *
+	 * @param string $scope Scope of the Service Worker.
+	 */
+	public function serve_request( $scope ) {
+
+		header( 'Content-Type: text/javascript; charset=utf-8' );
+		nocache_headers();
+
+		if ( ! in_array( $scope, $this->get_scopes(), true ) ) {
+			status_header( 404 );
```
westonruter: See #26. This should perhaps not return a `404` but rather return a `200` and a service worker script that simply unregisters itself.

westonruter:
Demonstration of the service worker API from this PR being used in the AMP plugin to cache the CDN assets: https://github.com/Automattic/amp-wp/pull/1261

wp-includes/class-wp-service-workers.php
westonruter: (My suggestion won't work.)

westonruter:
In a subsequent PR let's re-consider the scope, and whether we should allow arbitrary scopes or restrict to just `all`, `front`, and `admin`.

PR #16, opened 2018-06-25
Check for HTTPS support, and store the result in an option

### LINKED ISSUE #6: Add reliable method for detecting HTTPS support
The `is_ssl()` function is woefully inadequate to determine whether WordPress is being served via HTTPS. It is not reliable because HTTPS termination may be happening at a proxy and not at the server running WordPress. So it is not reliable to look at the server environment to determine if HTTPS is being served or is able to be served.

I believe the best way to detect HTTPS support is to actually perform a loopback request onto the home URL of the site with HTTPS and check that it responds properly with `sslverify` enabled. For added resilience, the request should include a token which the response then hashes with a secret which can then be compared to confirm that in fact it is the same WordPress install itself that is responding, and not some other WordPress site that is configured on HTTPS.

Once the HTTPS check is performed, the result should be stored in an option, and the option can then be checked instead of `is_ssl()`.

This would allow for a site to make use of HTTPS exclusively without manually adding a `FORCE_SSL_ADMIN` constant.

kienstra:
**Request For Review**

Hi @westonruter,
Could you please review this PR, based on [your ideas](https://github.com/xwp/pwa-wp/issues/6#issue-326244070) in #6?

* Schedules a cron event to check for support.
* Makes an HTTP loopback request, and verifies that the response is HTTPS.
* Then, it stores this in an option.
* Instead of `is_ssl()`, you call `get_option( 'is_https_supported' )`

Closes #6.

### DIFF
--- a/php/class-wp-https-detection.php
+++ b/php/class-wp-https-detection.php
@@ -68,12 +68,10 @@ class WP_HTTPS_Detection {
 	 * Returns true if the URL has the scheme of HTTPS, and it has the correct token.
 	 * The token is only to verify that the request is for the same site.
 	 * Also, in wp_remote_get(), 'sslverify' is true by default.
-	 * This method is private, as it's not part of the public API to get HTTPS support.
-	 * Instead, please use get_option( WP_HTTPS_Detection::HTTPS_SUPPORT_OPTION_NAME ).
 	 *
 	 * @return boolean Whether HTTPS is supported.
 	 */
-	private function is_https_supported() {
+	public function is_https_supported() {
 		$request = wp_remote_get( add_query_arg(
 			self::REQUEST_TOKEN_QUERY_ARG,
 			$this->get_token(),
--- a/php/class-wp-https-detection.php
+++ b/php/class-wp-https-detection.php
@@ -58,9 +58,13 @@ class WP_HTTPS_Detection {
 
 	/**
 	 * Makes a request to find whether HTTPS is supported, and stores the result in an option.
+	 * If the request is a WP_Error, don't update the option.
 	 */
 	public function update_option_https_support() {
-		update_option( self::HTTPS_SUPPORT_OPTION_NAME, $this->is_https_supported() );
+		$https_support = $this->check_https_support();
+		if ( is_bool( $https_support ) ) {
+			update_option( self::HTTPS_SUPPORT_OPTION_NAME, $https_support );
+		}
 	}
 
 	/**
@@ -70,43 +74,26 @@ class WP_HTTPS_Detection {
 	 * The token is only to verify that the request is for the same site.
 	 * Also, in wp_remote_get(), 'sslverify' is true by default.
 	 *
-	 * @return boolean Whether HTTPS is supported.
+	 * @return boolean|WP_Error Whether HTTPS is supported, or a WP_Error if the loopback request resulted in an error.
 	 */
-	public function is_https_supported() {
-		$request = wp_remote_request(
+	public function check_https_support() {
+		$response = wp_remote_request(
 			add_query_arg(
 				self::REQUEST_TOKEN_QUERY_ARG,
 				$this->get_token(),
-				home_url( '/', 'http' )
+				home_url( '/', 'https' )
 			),
 			array(
-				'timeout' => 10,
 				'headers' => array(
 					'Cache-Control' => 'no-cache',
 				),
 			)
 		);
 
-		if ( is_wp_error( $request ) || ! method_exists( $request['http_response'], 'get_response_object' ) ) {
-			return false;
+		if ( is_wp_error( $response ) ) {
+			return $response;
 		}
-
-		$response   = $request['http_response']->get_response_object();
-		$parsed_url = wp_parse_url( $response->url );
-		if ( ! $response->success || ! isset( $parsed_url['query'] ) ) {
-			return false;
-		}
-
-		$query_args = wp_parse_args( $parsed_url['query'] );
-		$is_https   = (
-			isset( $parsed_url['scheme'], $query_args[ self::REQUEST_TOKEN_QUERY_ARG ] )
-			&&
-			'https' === $parsed_url['scheme']
-			&&
-			$this->get_token() === $query_args[ self::REQUEST_TOKEN_QUERY_ARG ]
-		);
-
-		return $is_https;
+		return 200 === wp_remote_retrieve_response_code( $response );
 	}
 
 	/**
--- a/php/class-wp-https-detection.php
+++ b/php/class-wp-https-detection.php
@@ -46,7 +46,6 @@ class WP_HTTPS_Detection {
 		add_action( self::CRON_HOOK, array( $this, 'update_option_https_support' ) );
 		add_filter( 'cron_request', array( $this, 'ensure_http_if_sslverify' ), PHP_INT_MAX );
 		add_action( 'parse_query', array( $this, 'verify_https_check' ) );
-		add_action( 'init', array( $this, 'check_https_support' ) );
 	}
 
 	/**
--- a/php/class-wp-https-detection.php
+++ b/php/class-wp-https-detection.php
@@ -118,12 +118,8 @@ class WP_HTTPS_Detection {
 	 * @return array $request The filtered cron request arguments.
 	 */
 	public function ensure_http_if_sslverify( $request ) {
-		if ( 0 === strpos( $request['url'], 'https' ) ) {
-			if ( isset( $request['args'] ) ) {
-				$request['args']['sslverify'] = false;
-			} else {
-				$request['args'] = array( 'sslverify' => false );
-			}
+		if ( 0 === strpos( $request['url'], 'https://' ) ) {
+			$request['args']['sslverify'] = false;
 		}
 		return $request;
 	}
--- a/php/class-wp-https-detection.php
+++ b/php/class-wp-https-detection.php
@@ -18,25 +18,25 @@ class WP_HTTPS_Detection {
 	const CRON_HOOK = 'check_https_support';
 
 	/**
-	 * Option name for whether HTTPS is supported.
+	 * The option name for whether HTTPS is supported.
 	 *
 	 * @var string
 	 */
 	const HTTPS_SUPPORT_OPTION_NAME = 'is_https_supported';
 
 	/**
-	 * Secret for the HTTPS detection request.
+	 * The query var key for HTTPS detection, used to verify that the request is from the same origin.
 	 *
 	 * @var string
 	 */
-	const REQUEST_SECRET = 'https_detection_secret';
+	const REQUEST_QUERY_VAR = 'https_detection_token';
 
 	/**
-	 * Query arg key for the HTTPS detection token.
+	 * The number passed to the query var.
 	 *
-	 * @var string
+	 * @var int
 	 */
-	const REQUEST_TOKEN_QUERY_ARG = 'https_detection_token';
+	public $query_var_number;
 
 	/**
 	 * Inits the class.
@@ -45,6 +45,7 @@ class WP_HTTPS_Detection {
 		add_action( 'wp', array( $this, 'schedule_cron' ) );
 		add_action( self::CRON_HOOK, array( $this, 'update_option_https_support' ) );
 		add_filter( 'cron_request', array( $this, 'ensure_http_if_sslverify' ), PHP_INT_MAX );
+		add_action( 'parse_query', array( $this, 'verify_https_check' ) );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+	 * Also, in wp_remote_get(), 'sslverify' is true by default.
+	 *
+	 * @return boolean Whether HTTPS is supported.
+	 */
+	public function is_https_supported() {
+		$request = wp_remote_request(
+			add_query_arg(
+				self::REQUEST_TOKEN_QUERY_ARG,
+				$this->get_token(),
+				home_url( '/', 'http' )
```
kienstra: In my VVV, this `wp_remote_request()` results in:
>cURL error 28: Connection timed out after X milliseconds

But for testing purposes, changing the `home_url()` to an external URL seems to work. For example, http://wptavern.com

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+	/**
+	 * Gets the token, based on the secret.
+	 *
+	 * This token is not intended for security, as much as ensuring that the request is for this site.
+	 * By calling wp_hash(), it uses this site's salts, which should be different from other sites.
+	 *
+	 * @return string $token The token for the request.
+	 */
+	public function get_token() {
+		return wp_hash( self::REQUEST_SECRET );
```
kienstra: Maybe this isn't what you had in mind in this [comment](https://github.com/xwp/pwa-wp/issues/6#issue-326244070):

>For added resilience, the request should include a token which the response then hashes with a secret which can then be compared to confirm that in fact it is the same WordPress install itself that is responding, and not some other WordPress site that is configured on HTTPS.

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+
+		$response   = $request['http_response']->get_response_object();
+		$parsed_url = wp_parse_url( $response->url );
+		if ( ! $response->success || ! isset( $parsed_url['query'] ) ) {
+			return false;
+		}
+
+		$query_args = wp_parse_args( $parsed_url['query'] );
+		$is_https   = (
+			isset( $parsed_url['scheme'], $query_args[ self::REQUEST_TOKEN_QUERY_ARG ] )
```
westonruter: Instead of looking at the URL's query vars, I think we should be looking at the body of the response. That is, look at `wp_remote_retrieve_body( $response )` for the expected token. A query param in the requested URL can be the signal for whether the response should include the ✅ indicator that we then look for here.

php/class-wp-https-detection.php
westonruter: You could use a nonce instead like the AMP plugin is doing: https://github.com/Automattic/amp-wp/blob/a0eaa5328cdf60b22ed2d1f3e74753c6c04686e8/includes/validation/class-amp-validation-manager.php#L1410-L1435

The main thing is to just have a way to confirm that the response is actually being handled by the server we are intending. It could be a static string we look for like `WP_HTTPS_OK` but there could be a possibility that a site mentions the string and there is a false positive.

westonruter: So to summarize, the logic in `is_https_supported` could 

1. Generate a random number: `$number = wp_rand()`.
2. Make a loopback request to an HTTPS URL for the site with this `$number` added as a query var.
3. Have handler that runs at `parse_query` action that looks for this query var and then does `die( wp_hash( $query_var, 'nonce' ) )`
4. The method then takes the response body and does `wp_remote_retrieve_body( $response ) === wp_hash( $query_var, 'nonce' )`
5. If the values match, then we can confirm that the loopback request was answered by the same server. Otherwise, it did not.

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+	 *
+	 * @param array $request The cron request arguments.
+	 * @return array $request The filtered cron request arguments.
+	 */
+	public function ensure_http_if_sslverify( $request ) {
+		if ( 0 === strpos( $request['url'], 'https' ) ) {
+			if ( isset( $request['args'] ) ) {
+				$request['args']['sslverify'] = false;
+			} else {
+				$request['args'] = array( 'sslverify' => false );
```
westonruter: PHP allows for lazy-construction of multi-dimensional arrays, so this `if`/`else` statement can be replaced with just:

```php
$request['args']['sslverify'] = false;
```

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+	 * If the 'cron_request' arguments include an HTTPS URL, this ensures sslverify is false.
+	 *
+	 * Prevents an issue if HTTPS breaks,
+	 * where there would be a failed attempt to verify HTTPS.
+	 *
+	 * @param array $request The cron request arguments.
+	 * @return array $request The filtered cron request arguments.
+	 */
+	public function ensure_http_if_sslverify( $request ) {
+		if ( 0 === strpos( $request['url'], 'https' ) ) {
```
westonruter: I suggest replacing `https` with `https://` for good measure.

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+	/**
+	 * Makes a request to the home URL to determine whether HTTPS is supported.
+	 *
+	 * Returns true if the URL has the scheme of HTTPS, and it has the correct token.
+	 * The token is only to verify that the request is for the same site.
+	 * Also, in wp_remote_get(), 'sslverify' is true by default.
+	 *
+	 * @return boolean Whether HTTPS is supported.
+	 */
+	public function is_https_supported() {
```
westonruter: I think this would be marginally better-named `check_https_support()`. A `is_https_supported()` function I think would be more on the lines of doing `return (bool) get_option( self::HTTPS_SUPPORT_OPTION_NAME )`, in other words a direct replacement of `is_ssl()`.

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,140 @@
...
+	 * This uses wp_die() instead of die(),
+	 * because it allows filtering the handler: 'wp_die_handler'.
+	 * This prevents stopping the unit tests.
+	 *
+	 * @param WP_Query $wp_query The query object.
+	 */
+	public function verify_https_check( $wp_query ) {
+		$query_var = $wp_query->get( self::REQUEST_QUERY_VAR );
+		if ( ! empty( $query_var ) ) {
+			wp_die( wp_hash( $query_var, 'nonce' ) ); // WPCS: XSS ok.
```
kienstra: If it's alright, this uses `wp_die()` instead of `die()`. That allows filtering the handler: `'wp_die_handler'`, which prevents stopping the unit tests.

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,140 @@
...
+	 * @var string
+	 */
+	const REQUEST_QUERY_VAR = 'https_detection_token';
+
+	/**
+	 * The number passed to the query var.
+	 *
+	 * @var int
+	 */
+	public $query_var_number;
```
kienstra: This is a property, instead of a simple variable, so that the unit test can access it.

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,140 @@
...
+					'Cache-Control' => 'no-cache',
+				),
+			)
+		);
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		if ( false === strpos( wp_remote_retrieve_body( $response ), wp_hash( $this->query_var_number, 'nonce' ) ) ) {
```
kienstra: This uses `strpos()` instead of a simple comparison because `wp_die()` wraps the `<body>` contents in some markup.

php/class-wp-https-detection.php
kienstra: Thanks for the detailed solution. [check_https_support()](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR80) implements it mainly as you described, though the [parse_query handler](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR137) uses `wp_die()` instead of `die()`, if that's alright.

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+			array(
+				'timeout' => 10,
+				'headers' => array(
+					'Cache-Control' => 'no-cache',
+				),
+			)
+		);
+
+		if ( is_wp_error( $request ) || ! method_exists( $request['http_response'], 'get_response_object' ) ) {
+			return false;
```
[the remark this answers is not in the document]
kienstra: Thanks, it [now returns](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR100) a `WP_Error`, and only updates the option if `is_bool()`.

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+	 */
+	public function is_https_supported() {
+		$request = wp_remote_request(
+			add_query_arg(
+				self::REQUEST_TOKEN_QUERY_ARG,
+				$this->get_token(),
+				home_url( '/', 'http' )
+			),
+			array(
+				'timeout' => 10,
```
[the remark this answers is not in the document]
kienstra: Sure, the timeout argument is [now removed](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR91).

php/class-wp-https-detection.php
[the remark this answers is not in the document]
kienstra: Sure, this [now passes](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR860) 'https' to `home_url()`.

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+	 * Makes a request to the home URL to determine whether HTTPS is supported.
+	 *
+	 * Returns true if the URL has the scheme of HTTPS, and it has the correct token.
+	 * The token is only to verify that the request is for the same site.
+	 * Also, in wp_remote_get(), 'sslverify' is true by default.
+	 *
+	 * @return boolean Whether HTTPS is supported.
+	 */
+	public function is_https_supported() {
+		$request = wp_remote_request(
```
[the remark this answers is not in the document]
kienstra: Thanks, that's [fixed here](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR82).

php/class-wp-https-detection.php
```diff
@@ -0,0 +1,144 @@
...
+				),
+			)
+		);
+
+		if ( is_wp_error( $request ) || ! method_exists( $request['http_response'], 'get_response_object' ) ) {
+			return false;
+		}
+
+		$response   = $request['http_response']->get_response_object();
+		$parsed_url = wp_parse_url( $response->url );
```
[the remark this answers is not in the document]
kienstra: Thanks, here's the [newly-refactored](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR106) method.

php/class-wp-https-detection.php
kienstra: Thanks, good suggestion. It's [applied here](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR120).

php/class-wp-https-detection.php
kienstra: Sure, [it's now](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR119) `https://`.

php/class-wp-https-detection.php
kienstra: Thanks, [it's renamed](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR80) to `check_https_support()`.

kienstra: Also, here's the new method [is_https_supported()](https://github.com/xwp/pwa-wp/pull/16/files#diff-1d91a59d72444acecd2a72e8baa63c2cR56).

php/class-wp-https-detection.php
westonruter: Right before calling `wp_die()` you could do this:

```php
add_filter( 'wp_die_handler', function( $message ) {
    die( $message );
}, 1 );
```

This will then get overridden during unit tests to be `\WP_UnitTestCase::wp_die_handler()` via `\WP_UnitTestCase::get_wp_die_handler()`

php/class-wp-https-detection.php
westonruter: See https://github.com/xwp/pwa-wp/pull/16/files#r198711771

Issue #5, opened 2018-05-24
Propose a Workbox solution for WordPress


postphotos:
**As a contributor to this project, after I've understood and and documented how the Workbox APIs work, I should propose an architectural plan using Workbox that will allow plugins, themes and site owners alike to share PWA component features in a single service worker implementation without conflict.**

Related #1. #2.

miina:
Thinking that maybe with implementing Workbox registering an actual JS file with SW logic wouldn't be required for plugins using the API, and we could just add PHP functions for Workbox functionalities and then concatenate the relevant Workbox code together into dynamically created SW logic per scope, as it is now in the API Prototype within #14.

For example there’s a network first Workbox caching strategy which can be used for specific routes (`workbox.strategies.networkFirst()`). Instead of creating and registering a separate SW script file including the logic maybe with Workbox just a method something like `wp_add_sw_network_first_strategy( $route, $args )` could be used and then the feature plugin would add the relevant code to the SW logic (based on the current API prototype generated at `"/?wp_service_workers=$scope"`)

This would also make it possible to manage potential conflicts.

Thoughts?
cc @westonruter @kienstra

kienstra:
**Workbox Strategy**

Hi @miina,
Good idea to have a PHP function to generate code for the Workbox API. I think that'd be useful for some cases, like the `wp_add_sw_network_first_strategy( $route, $args )` method you mentioned.

Is there a plan for a `/wp-admin` UI to show possible conflicts? Like if one plugin [uses workbox.strategies.cacheFirst](https://developers.google.com/web/ilt/pwa/lab-workbox#5_add_routes_to_the_service_worker) on a route, while another uses `workbox.strategies.StaleWhileRevalidate`?

Otherwise, I'm not sure how useful it would be to create a full API for the Workbox API. In the example conflict above, it seems that one just overrides the other. There's no console error. 

Your service worker API in #14 already outputs service worker code that plugins or themes register via `wp_register_service_worker()`. 

If we're not going to have a `/wp-admin` UI to show conflicts, it seems like plugins could just include any Workbox calls in their service worker JS file that your service worker API above registers.

Having a full API to the Workbox API seems like a lot. Developers would have to learn a new API, and we'd have to maintain documentation.

Of course, I could have overlooked some benefits of having this API.

miina:
@kienstra Thanks for looking into this.

True that there it wouldn't make sense to create a full API for the full Workbox API. 

[...]

PR #27, opened 2018-07-10
Replace arbitrary service worker script scopes with just three: all, front, and admin

### LINKED ISSUE #26: Unregister unused ServiceWorker scopes
By design, once you register a ServiceWorker it stays registered untill you unregister it.
Now since we have the ServiceWorker API we update an active ServiceWorker based on the Scope on each pageload.
Code that is not longer in use will be removed once the ServiceWorker updates (so on every pageload). Thats good.

But what if a plugin uses a special scope? Once the plugin is disabled, the registration/update method won't be included anymore and the ServiceWorker will still be active in browsers that visited the page when the plugin was active.

Should we add an unregister method for all ServiceWorker/Scopes that have ever been registered?

westonruter:
In #14 the ability was added for a service worker script to be registered with arbitrary scopes. This ability caused a problem (#26) whereby if a plugin registered a service worker script to an arbitrary scope and then that plugin was deactivated, that the service worker would be left installed indefinitely.

In reality, I believe only two distinct service workers would be relevant in a WordPress context: the frontend (`home_url('/')`) and the wp-admin (`admin_url('/')`). These are really the two separate environments in WordPress. On the frontend, all requests get routed by `WP_Rewrite` and loaded via `/index.php` and it makes sense that all service worker scripts used on the frontend should similarly get the corresponding `/` scope. A frontend service worker would likely be used to cache _frontend_ assets (such as for an app shell), which should be very different from how caching is handled in the admin. In the admin it would make sense for there to be scripts registered which are specific to the apps running in that context (e.g. Gutenberg) which would be needless to be included in the service worker on the frontend. Also, it is important to note that the frontend of a site and the WP admin can be served from different domains (e.g. `example.com` and `example.wordpress.com/wp-admin`) and so again, there is a natural differentiation between them. 

Because of this potential for differing front/admin domains, I think it is important for consistency to always have a service worker registered at `/` and another at `/wp-admin/`. Otherwise, it could be that a plugin registers a script to run at the `/` scope and would then expect it to run at `/wp-admin/` but when the the admin is accessed on a different, this wouldn't happen as expected: the service worker installed on the home origin would not be the same service worker installed when accessing the admin. A separate admin service worker would have to get installed anyway.

So I think we should always register a service worker at the `home_url( '/' )` scope for the frontend, and another service worker scope at `site_url( '/wp-admin/' )`. When the home and site URLs are the same, then the admin service worker could be installed when accessing the frontend, but we can limit this installation to only happen when the user is logged in; this installation can also happen when accessing the login screen so the admin service worker installs while logging in.

Finally, it should also be possible register a service worker script to run in both the front and admin contexts. For example in https://github.com/Automattic/amp-wp/pull/1261 I added a service worker to the AMP plugin to cache AMP CDN assets (e.g. so that they can be loaded while developing offline). Such a service worker logic is relevant to the frontend and wp-admin, and such a script a script can be registered with an `all` scope, instead of a `front` or `admin` scope.

# Usage

```php
// Register script to only run on the frontend, with dependency on some other app-shell SW script.
wp_register_service_worker(
    'foo', // Handle.
    plugin_dir_url( __FILE__ ) . 'foo.js', // Source.
    array( 'app-shell' ), // Dependency.
    WP_Service_Workers::SCOPE_FRONT // Scope.
);

// Register script (here via render callback instead of URL) to only run only in the admin.
wp_register_service_worker( 
    'bar', 
    function() { return 'console.info("Hello admin!")'; }, 
    array(), // No deps.
    WP_Service_Workers::SCOPE_ADMIN
);

// Register script with to run in both the frontend and wp-admin.
wp_register_service_worker( 'baz', plugin_dir_url( __FILE__ ) . 'baz.js', array(), WP_Service_Workers::SCOPE_ALL );

// The default values for $deps and $scope are array() and WP_Service_Workers::SCOPE_ALL  respectively, so this is same as previous.
wp_register_service_worker( 'baz', plugin_dir_url( __FILE__ ) . 'baz.js' );
```

# Changes

* Replace arbitrary scopes with just three: all, admin, and front.
* Install service workers on login screen and other non-front/admin screens
* Serve scripts registered to 'all' scope in service worker requests for both 'front' and 'admin'.
* When requesting service worker other than front/admin respond with 400 error.
* Remove no-cache header which conflicted with logic to handle not-modified responses.
* Improve comments in service worker before each script is printed, indicating handle and error.
* Let remove_url_scheme and get_validated_file_path methods be protected.
* Limit installation of service workers for front and admin based on whether they can be installed (on basis of current origin and whether user would access admin).

Closes #26. See #7.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 build/
 composer.lock
+/vendor
--- a/tests/test-class-wp-service-workers.php
+++ b/tests/test-class-wp-service-workers.php
@@ -23,37 +23,199 @@ class Test_WP_Service_Workers extends WP_UnitTestCase {
 	 * @inheritdoc
 	 */
 	public function setUp() {
+		global $wp_actions, $wp_default_service_workers;
 		parent::setUp();
-		$this->instance = new WP_Service_Workers();
+		unset( $wp_default_service_workers );
+		unset( $wp_actions['wp_default_service_workers'] );
+		$this->instance = wp_service_workers();
+	}
+
+	/**
+	 * Tear down.
+	 */
+	public function tearDown() {
+		parent::tearDown();
+		unset( $GLOBALS['wp_service_workers'] );
 	}
 
 	/**
 	 * Test class constructor.
 	 *
 	 * @covers WP_Service_Workers::__construct()
+	 * @covers WP_Service_Workers::init()
+	 * @covers wp_service_workers()
 	 */
 	public function test_construct() {
-		$service_workers = new WP_Service_Workers();
-		$this->assertEquals( 'WP_Service_Workers', get_class( $service_workers ) );
+		global $wp_service_workers;
+		$this->assertEquals( 1, did_action( 'wp_default_service_workers' ) );
+		$this->assertSame( $wp_service_workers, $this->instance );
+		$this->assertInstanceOf( 'WP_Service_Workers', $this->instance );
+		$this->instance->init();
+
+		unset( $wp_service_workers );
+		wp_service_workers();
+		$this->assertEquals( 2, did_action( 'wp_default_service_workers' ) );
+		wp_service_workers();
+		$this->assertEquals( 2, did_action( 'wp_default_service_workers' ) );
+	}
+
+	/**
+	 * Get scripts.
+	 */
+	public function get_scripts() {
+		return array(
+			'default' => array(
+				'foo-all',
+				'/test-sw.js',
+				array( 'bar' ),
+				null,
+			),
+			'all'     => array(
+				'foo',
+				'/test-sw.js',
+				array(),
+				'all',
+			),
+			'front'   => array(
+				'foo',
+				'/test-sw.js',
+				array(),
+				'front',
+			),
+			'admin'   => array(
+				'foo',
+				'/test-sw.js',
+				array(),
+				'admin',
+			),
+		);
 	}
 
 	/**
 	 * Test adding new service worker.
 	 *
+	 * @param string $handle Handle.
+	 * @param string $src    Source.
+	 * @param array  $deps   Dependencies.
+	 * @param string $scope  Scope.
+	 * @dataProvider get_scripts
 	 * @covers WP_Service_Workers::add()
 	 */
-	public function test_register() {
-		$this->instance->register( 'foo', '/test-sw.js', array( 'bar' ) );
+	public function test_register( $handle, $src, $deps, $scope = null ) {
+		if ( $scope ) {
+			$this->instance->register( $handle, $src, $deps, $scope );
+		} else {
+			$this->instance->register( $handle, $src, $deps );
+		}
+		if ( ! $scope ) {
+			$scope = 'all';
+		}
+		$this->assertEquals( $scope, $this->instance->registered[ $handle ]->args['scope'] );
+		$this->assertArrayHasKey( $handle, $this->instance->registered );
+		$registered_sw = $this->instance->registered[ $handle ];
+		$this->assertEquals( $src, $registered_sw->src );
+		$this->assertEquals( $deps, $registered_sw->deps );
+	}
 
-		$default_scope = site_url( '/', 'relative' );
+	/**
+	 * Test using invalid scope.
+	 *
+	 * @expectedIncorrectUsage WP_Service_Workers::register
+	 */
+	public function test_register_invalid_scope() {
+		$this->instance->register( 'foo', '/test-sw.js', array( 'bar' ), 'bad' );
+		$this->assertEquals( 'all', $this->instance->registered['foo']->args['scope'] );
+	}
 
-		$this->assertTrue( in_array( $default_scope, $this->instance->get_scopes(), true ) );
-		$this->assertTrue( isset( $this->instance->registered['foo'] ) );
+	/**
+	 * Test serve_request.
+	 *
+	 * @covers WP_Service_Workers::serve_request()
+	 * @covers WP_Service_Workers::do_items()
+	 */
+	public function test_serve_request() {
+		wp_service_workers()->register( 'bar', array( $this, 'return_bar_sw' ), array( 'foo' ), 'front' );
+		wp_service_workers()->register( 'baz', array( $this, 'return_baz_sw' ), array( 'foo' ), 'admin' );
+		wp_service_workers()->register( 'foo', array( $this, 'return_foo_sw' ), array(), 'all' );
 
-		$registered_sw = $this->instance->registered['foo'];
+		ob_start();
+		wp_service_workers()->serve_request( 'bad' );
+		$this->assertContains( 'invalid_scope_requested', ob_get_clean() );
 
-		$this->assertEquals( '/test-sw.js', $registered_sw->src );
-		$this->assertTrue( in_array( $default_scope, $registered_sw->args['scopes'], true ) );
-		$this->assertEquals( array( 'bar' ), $registered_sw->deps );
+		ob_start();
+		wp_service_workers()->serve_request( 'front' );
+		$output = ob_get_clean();
+		$this->assertContains( $this->return_foo_sw(), $output );
+		$this->assertContains( $this->return_bar_sw(), $output );
+		$this->assertTrue(
+			strpos( $output, $this->return_foo_sw() ) < strpos( $output, $this->return_bar_sw() )
+		);
+
+		ob_start();
+		wp_service_workers()->serve_request( 'admin' );
+		$output = ob_get_clean();
+
+		$this->assertContains( $this->return_foo_sw(), $output );
+		$this->assertContains( $this->return_baz_sw(), $output );
+		$this->assertTrue(
+			strpos( $output, $this->return_foo_sw() ) < strpos( $output, $this->return_baz_sw() )
+		);
+	}
+
+	/**
+	 * Test serve_request for bad src callback.
+	 *
+	 * @expectedIncorrectUsage WP_Service_Workers::register
+	 * @covers WP_Service_Workers::serve_request()
+	 * @covers WP_Service_Workers::do_items()
+	 */
+	public function test_serve_request_bad_src_callback() {
+		wp_service_workers()->register( 'bar', array( 'Does_Not_Exist', 'return_bar_sw' ) );
+		ob_start();
+		wp_service_workers()->serve_request( 'admin' );
+		$output = ob_get_clean();
+		$this->assertContains( 'Service worker src is invalid', $output );
+	}
+
+	/**
+	 * Test serve_request bad src URL.
+	 *
+	 * @expectedIncorrectUsage WP_Service_Workers::register
+	 * @covers WP_Service_Workers::serve_request()
+	 * @covers WP_Service_Workers::do_items()
+	 */
+	public function test_serve_request_bad_src_url() {
+		wp_service_workers()->register( 'bar', '/food.png' );
+		ob_start();
+		wp_service_workers()->serve_request( 'front' );
+		$output = ob_get_clean();
+		$this->assertContains( 'Service worker src is invalid', $output );
+	}
+
+	/**
+	 * Get some JS code.
+	 *
+	 * @return string JS example code.
+	 */
+	public function return_foo_sw() {
+		return 'console.info("Hello Foo World");';
+	}
+
+	/**
+	 * Get some JS code.
+	 *
+	 * @return string JS example code.
+	 */
+	public function return_bar_sw() {
+		return 'console.info("Hello Bar World");';
+	}
+
+	/**
+	 * Get some JS code.
+	 *
+	 * @return string JS example code.
+	 */
+	public function return_baz_sw() {
+		return 'console.info("Hello Baz World");';
 	}
 }
--- a/tests/test-class-wp-service-workers.php
+++ b/tests/test-class-wp-service-workers.php
@@ -147,6 +147,7 @@ class Test_WP_Service_Workers extends WP_UnitTestCase {
 		$output = ob_get_clean();
 		$this->assertContains( $this->return_foo_sw(), $output );
 		$this->assertContains( $this->return_bar_sw(), $output );
+		$this->assertNotContains( $this->return_baz_sw(), $output );
 		$this->assertTrue(
 			strpos( $output, $this->return_foo_sw() ) < strpos( $output, $this->return_bar_sw() )
 		);
@@ -156,6 +157,7 @@ class Test_WP_Service_Workers extends WP_UnitTestCase {
 		$output = ob_get_clean();
 
 		$this->assertContains( $this->return_foo_sw(), $output );
+		$this->assertNotContains( $this->return_bar_sw(), $output );
 		$this->assertContains( $this->return_baz_sw(), $output );
 		$this->assertTrue(
 			strpos( $output, $this->return_foo_sw() ) < strpos( $output, $this->return_baz_sw() )
--- a/tests/test-class-wp-service-workers.php
+++ b/tests/test-class-wp-service-workers.php
@@ -74,19 +74,19 @@ class Test_WP_Service_Workers extends WP_UnitTestCase {
 				'foo',
 				'/test-sw.js',
 				array(),
-				'all',
+				WP_Service_Workers::SCOPE_ALL,
 			),
 			'front'   => array(
 				'foo',
 				'/test-sw.js',
 				array(),
-				'front',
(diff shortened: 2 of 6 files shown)


westonruter:
I terms of implementation, maybe instead of using a custom `scope` script arg we should re-use the dependency `group` instead, where `0` is all, `1` is frontend, and `2` is wp-admin. It wouldn't make any difference in practice. It's just we're not using the `$group` argument in`WP_Service_Workers::do_items()` and instead are manually collecting the handles in `$scope_items`.

JJJ:
Login/Register/Activate pages are unique in that they are neither front nor admin.

Could we foresee API specific service workers? If so, is REST unique from XML-RPC? Should WPGraphQL do the same front/admin specific scope instead of providing their own?

RSS/feeds? (I know support is waning.)

Any concerns about admin-ajax, trackbacks, cron, comments-post, or other core files that are anomalies in how they are loaded and what they load with them?

gravityrail:
> RSS/feeds? (I know support is waning.)

Feeds don't support JavaScript :)

westonruter:
@JJJ: 
> Login/Register/Activate pages are unique in that they are neither front nor admin.

The idea here is that if you are accessing the login screen or the register/activate screen that it is likely you will soon be accessing the admin. This being the case, the `admin` service worker should get installed while you're on login/register/activate so that it's all primed by the time you do access the admin. 

> Could we foresee API specific service workers? If so, is REST unique from XML-RPC? Should WPGraphQL do the same front/admin specific scope instead of providing their own? ¶ Any concerns about admin-ajax, trackbacks, cron, comments-post, or other core files that are anomalies in how they are loaded and what they load with them?

I'm not sure. I'm thinking specifically in the context of the service worker that is running when you're at `/` vs `/wp-admin/` and the `fetch` events it might handle. The REST API runs under `home_url( '/wp-json/' )` so requests to it would run through the `front` scope as proposed above. This makes sense to me for example in the context of an app shell PWA that needs to have control over the URL space over the frontend of the site and across the the REST API endpoints in order to effectively serve a site offline. RSS feeds would be included here, but I doubt that modern JS applications would be be fetching posts via RSS instead of via the REST API. The `fetch` handlers will need to whitelist which routes they handle, so they should be ignoring `xmp-rpc.php` and other such special endpoints. 

In #5 there is an investigation into using Workbox to provide an abstraction layer to the service worker API which we can then in turn provide an PHP API abstraction in WordPress. We'd probably want to fail with as `_doing_it_wrong()` for any `fetch` handlers that want to proxy the network requests such special WordPress endpoints.

nico-martin:
Interresting feed so far. One thing first. I don't think we need to find solutions for all usecases right now. We only need to think about the scopes of serviceworkers.
On a regular installation we would have the frontend scope `https://site.com/` and the admin scope `https://site.com/wp-admin/`. So the frontend scope will cover the admin scope as well.
But to split it makes sense because sometimes we have different urls and not all visitors need all the admin SW stuff (but sll admins do need the frontend SW stuff).

@JJJ We don't need API specific ServiceWorkes. As @gravityrail suggested you can't register a SW on an API request. But I think you are not talking about API specific ServiceWorkes but more about how we handle API requests inside the frontend ServiceWorker, right?

I'd say we shouldn't touch them. APIs should always provide the latest informations. Let's say I'm writing an SPA toghether with the WP Rest API. I'd rather have a 404 response than an offline/cached version of the Endpoint. If I have the 404 I can decide what my SPA should do. Otherwise I have no way to check wether the request was successfull and my data is up to date or I recieved outdated data.

Maybe there are some rare usecases where an offline-API could make sence. But then it would make more sense to create a plugin that adds those events to the frontend-SW.

nico-martin:
I'm a bit concerned about the `all` scope.
Let's say you have some SW-JS that handles the fetch events for all JS files.
```JS
workbox.routing.registerRoute(
    new RegExp('.*\.js'),
    workbox.strategies.networkFirst()
);
```
Lets say we put that inside the `all` scope and we are currently inside the wp-admin area. That would mean we have two SW registered in our browser and both have the same event for handling JS files. Doesn't make sense, right?

Speaking of that. What happens if we have an `all` scope SW-JS and a `admin` scope SW-JS?
```JS
// all scope
workbox.routing.registerRoute(
    new RegExp('.*\.js'),
    workbox.strategies.cacheFirst()
);
```
```JS
// admin scope
workbox.routing.registerRoute(
    new RegExp('.*\.js'),
    workbox.strategies.networkFirst()
);
```

In that case w e would have 3 JS events in two ServiceWorkers. We have the cacheFirst strategy from the FrontentSW, we have the cacheFirst strategy from the AdminSW **and** we have the networkFirst strategy from the AdminSW. Thats all a bit confusing..

felixarntz:
I am in favor of using the group approach with integers, however I'd strongly advise for using constants there, either globally defined (such as `SW_GROUP_ALL`, `SW_GROUP_FRONTEND`, and `SW_GROUP_ADMIN`) or, better even, in a "enum-like" class (such as `Service_Worker_Group::ALL`, `Service_Worker_Group::FRONTEND`, and `Service_Worker_Group::ADMIN`). Having constants makes it obvious which values are available, allows for autocompletion support and prevents typos from causing unexpected behavior because it would trigger a notice instead.
Also, in terms of common practices, it would make sense to use values that work correctly with bit operators, for example, the value for `Service_Worker_Group::ALL` should be the value of `Service_Worker_Group::FRONTEND | Service_Worker_Group::ADMIN`. 0 should only be used for something like `Service_Worker_Group::NONE` (which I don't think is needed, so we should skip that here).

Regarding the scope in which those apply, I wonder whether we could be inspired by how this is handled for cookies. Either way, I think having constants available for this as well would make a lot of sense, however in this case not for consuming developers to use (like plugin developers), but more so for site owners, where they can define those relative URLs in `wp-config.php`, and if not, they are defined later with defaults, just like currently `COOKIE_PATH` etc. That may be useful especially for multisite where service workers should possibly be able to go beyond the scope of a single site.

nico-martin:
Ok. Interresting thread here:
https://stackoverflow.com/a/45285119/2949413
> The behavior is well-defined: when a client page makes a request, it will trigger the fetch handlers of the controlling service worker one by one, in the order in which they were registered, until the first call is made to event.respondWith(). One one fetch event handler calls respondWith(), no other fetch event handlers will be triggered, and it's the sole responsibility of that handler to (eventually) return a Response to the client page.

I assume that applies also to multiple fetch handlers from multiple ServiceWorkers.
But does anybody know whats the order of already registered ServiceWorkers? If someone visits the frontend, the frontendSW will be registered. Then he visits the admin area, the adminSW will be registered. How can we make sure the events from the adminSW are executed **before** the events from the frontendSW?

westonruter:
> I am in favor of using the group approach with integers, however I'd strongly advise for using constants there, either globally defined (such as `SW_GROUP_ALL`, `SW_GROUP_FRONTEND`, and `SW_GROUP_ADMIN`) or, better even, in a "enum-like" class (such as `Service_Worker_Group::ALL`, `Service_Worker_Group::FRONTEND`, and `Service_Worker_Group::ADMIN`). Having constants makes it obvious which values are available, allows for autocompletion support and prevents typos from causing unexpected behavior because it would trigger a notice instead.

Yes, good idea to use constants here.

> Also, in terms of common practices, it would make sense to use values that work correctly with bit operators, for example, the value for `Service_Worker_Group::ALL` should be the value of `Service_Worker_Group::FRONTEND | Service_Worker_Group::ADMIN`. 0 should only be used for something like `Service_Worker_Group::NONE` (which I don't think is needed, so we should skip that here).

I thought about this. And yes, we could do this, but it just seems a little bit overkill. I think `FRONT` could be the value `1` (`b1`) and `ADMIN` could have value of `2` (`b10`) and `ALL` could be `3` (`b11`). You're right that a `NONE`group wouldn't make sense.

> Regarding the scope in which those apply, I wonder whether we could be inspired by how this is handled for cookies. Either way, I think having constants available for this as well would make a lot of sense, however in this case not for consuming developers to use (like plugin developers), but more so for site owners, where they can define those relative URLs in `wp-config.php`, and if not, they are defined later with defaults, just like currently `COOKIE_PATH` etc. That may be useful especially for multisite where service workers should possibly be able to go beyond the scope of a single site.

Multisite is currently handled by making sure that the scopes' paths relative to the location of the subdirectory install. So it does not use the scope of `/` for `FRONT` but rather `home_url( '/', 'relative' )`. Likewise, for `ADMIN` it uses the scope of `admin_url( '/' )` to ensure it runs under `/wp-admin/` (and also may be on a different domain). So is there a need for having additional configuration via constants?

Now, in regards to having a separate service worker for login, signup, activate pages (as …

In the case of the login screen I think the `FRONT` service worker could be running but it just probably not have anything to do, since the page wouldn't be making requests that the service worker should be intercepting. I didn't think login pages warrant having a separate scope. What would the login screen need a dedicated service worker for? It's not a page with assets that would likely need to be proxied by a service worker. I can't imagine it anyway.

JJJ:
I can imagine there being a need for multiple front scopes. It’s not uncommon for plugins to define some “root” page/slug for themselves to naturally reduce their scope already. bbPress does this with forums. BuddyPress does this with each of its components. EDD has specific pages for products/cart/success/fail.

BuddyPress and bbPress also rely on the vanilla login forms, but both force a redirect back to the front for Subscriber role users who shouldn’t have access to wp-admin. That makes me question the assumption that those pages are admin-scope.

westonruter:
@JJJ If a plugin were to add a service worker script scoped to their own path on the site, then I'd worry that it would cause unexpected results. In particular, let's say a theme wants to add a service worker script to handle caching of stylesheets and this service worker is registered at the `/` scope:

```js
self.addEventListener( 'fetch', ( event ) => {
    const url = new URL( event.request.url );
    if ( /\/wp-content\/themes\/example\/.*\.css$/.test( url.pathname ) ) {
        // Handle response.
    }
} );
```

_But_ then you have another plugin (e.g. EDD) that adds offline support for the shopping cart page (for example), and it only is registered to the `/store/` scope:

```
self.addEventListener( 'fetch', ( event ) => {
    const url = new URL( event.request.url );
    if ( /\/store\/cart\/$/.test( url.pathname ) ) {
        // Handle response.
    }
} );
```

The result is that the service worker registered with the `/store/` scope would take control over the page and exclude the service worker registered at the `/` scope. Only one service worker can handle `fetch` events on a given page, and the service worker that does runs is the one that has the longest-matching scope for the current `window.location`. So in this case the theme's logic (above) for handling the caching for stylesheets would **not run** because the store-specific service worker is more specific. 

See more at https://stackoverflow.com/a/41707357/93579:

> It's possible to register multiple service workers with different scopes for a given origin. (Aside: If you try to register multiple service workers with the *same* scope for a given origin, even if the service workers [have different URLs](https://developers.google.com/web/fundamentals/instant-and-offline/service-worker/lifecycle#avoid_changing_the_url_of_your_service_worker_script), the subsequent registrations will be treated like service worker [updates](https://developers.google.com/web/fundamentals/instant-and-offline/service-worker/lifecycle#updates), not as separate registrations.)
>
>If multiple service workers are registered, then in order to determine which one controls a give client page, the scopes are checked. The service worker [whose registered scope is **most specific**](https://w3c.github.io/ServiceWorker/#scope-match-algorithm), i.e. is the longest matching scope, is the one which will control a given client page.

All of this to say, however, a theme wanting to add `fetch` handler at the root can work well with a plugin wanting to add a `fetch` handler for an e-commerce plugin wanting to handle requests under `/store/`. The `fetch` handlers just need to make sure they are explicit about the URLs they are handling. If this is done, then they can all be concatenated into a single `front` service worker which is registered at the `/` scope: 

```js
// From theme:
self.addEventListener( 'fetch', ( event ) => {
    const url = new URL( event.request.url );
    if ( /\/wp-content\/themes\/example\/.*\.css$/.test( url.pathname ) ) {
        // Handle response.
    }
} );

// From store plugin:
self.addEventListener( 'fetch', ( event ) => {
    const url = new URL( event.request.url );
    if ( /\/store\/cart\/$/.test( url.pathname ) ) {
        // Handle response.
    }
} );
```

(The `event.request.url` is not the current `window.location` but rather the URL for the given HTTP request being made, even to URLs that are outside the current scope, including to entirely different domains. The SW `scope` controls which service worker gets installed for a given HTML document you access.)

Note that the fist `fetch` handler that handles an event will cause all subsequent handlers to be skipped. So it's important that the URL pattern matches be as specific as possible. In #5 there's an exploration into using [Workbox](https://developers.google.com/web/tools/workbox/) to facilitate registering the routes and their caching strategies in a way that we can detect `fetch` collisions which would otherwise be silently skipped.

I think this explains why having a single `front` scope is advantageous and that allowing for arbitrary scopes on the frontend would just increase the likelihood for unexpected behavior.

JJJ:
@westonruter [that explanation](https://github.com/xwp/pwa-wp/pull/27#issuecomment-403861700) helps immensely. Thanks!

nico-martin:
Sounds good to me. Especially the service worker hierarchy. I didn't know that, but it makes sense.

westonruter:
> I am in favor of using the group approach with integers, however I'd strongly advise for using constants there, either globally defined (such as `SW_GROUP_ALL`, `SW_GROUP_FRONTEND`, and `SW_GROUP_ADMIN`) or, better even, in a "enum-like" class (such as `Service_Worker_Group::ALL`, `Service_Worker_Group::FRONTEND`, and `Service_Worker_Group::ADMIN`). Having constants makes it obvious which values are available, allows for autocompletion support and prevents typos from causing unexpected behavior because it would trigger a notice instead.
Also, in terms of common practices, it would make sense to use values that work correctly with bit operators, for example, the value for `Service_Worker_Group::ALL` should be the value of `Service_Worker_Group::FRONTEND | Service_Worker_Group::ADMIN`. 0 should only be used for something like `Service_Worker_Group::NONE` (which I don't think is needed, so we should skip that here).

@felixarntz What do think of https://github.com/xwp/pwa-wp/pull/27/commits/1cb73acd1df3ef2f114d996798128ac5a4ead751? Switches to:

* `\WP_Service_Workers::SCOPE_FRONT` (b01) instead of `'front'`
* `\WP_Service_Workers::SCOPE_ADMIN` (b10) instead of `'admin'`
* `\WP_Service_Workers::SCOPE_ALL` (b11) instead of `'all'`

postphotos:
This is fantastic. Thanks, @westonruter for your efforts here! 🎉

felixarntz:
@westonruter 
> Multisite is currently handled by making sure that the scopes' paths relative to the location of the subdirectory install. So it does not use the scope of `/` for `FRONT` but rather `home_url( '/', 'relative' )`. Likewise, for `ADMIN` it uses the scope of `admin_url( '/' )` to ensure it runs under `/wp-admin/` (and also may be on a different domain). So is there a need for having additional configuration via constants?

I think using constants would make sense, for example for a case where a subdirectory configuration would like the service workers to not act in the relative home URL scope (`/my-subsite/`), but rather in the network URL scope (`/`). On a subdirectory network, that would allow service workers to work on the full network level. I see that it might be a bit far fetched at this point, but I don't see any harm in introducing constants for this. Multisite (and multinetwork especially) is used in tons of ways that we should open up flexibility for here IMO.

> What do think of 1cb73ac?

I like those, looks good to me!

> I thought about this. And yes, we could do this, but it just seems a little bit overkill. I think `FRONT` could be the value `1` (`b1`) and `ADMIN` could have value of `2` (`b10`) and `ALL` could be `3` (`b11`). You're right that a `NONE` group wouldn't make sense.

That's fair. However, if the decision is made to work with integers here, I definitely suggest using the numbers like that so that they are logical in terms of bit operators.

westonruter:
> I like those, looks good to me! I just saw you went with the integer approach for scopes as well right? I wonder whether we can get rid of the `front` and `admin` strings that are still in `wp_get_service_worker_url()` and `wp_print_service_workers()` somehow.

@felixarntz Good catch! Updated in 36704a7.

aristath:
A frontent/wp-admin/all separation sounds like a plan to me 👍

justlevine:
Just my (wholly unsolicited) 2 cents regarding Login/Register/Activate: 
A decent amount of sites use these for the frontend as well (e-commerce, forums, social, etc), and the majority of logged-in users aren't even allowed to access the backend.

Assumingly, most in this use case who use these (instead of a plugin-provided solution) are capable enough to use a filter/action to say with group of service workers load, but don't take it for granted that sites are always going to want/need wp-admin service workers loading up.

westonruter:
Yeah, in that case the admin service worker could get installed for users but it would never be used. If a filter is used to turn that off, however, then admin users wouldn't get the benefit of the service worker installing during login. Maybe that's something they'd just have to live with.

Issue #25, opened 2018-07-07
Implement precaching


westonruter:
As a site admin, I should have a way to precache important assets to give my users a faster perceived experience of using my WordPress site. 

~AC1: Propose a user interface to select which assets should be pre-cached.~
AC2: As site admins might need to declare precached asset support in different ways, also propose a theme function and/or wp-cli script to select which assets should be precached.

_Note: additional ACs likely to be added._

---

As noted by @jeffposnick in https://github.com/xwp/pwa-wp/issues/5#issuecomment-403082892:

> Something I've not seen mentioned yet has to do with precaching—here's [an explainer](https://developers.google.com/web/tools/workbox/modules/workbox-precaching) about how Workbox handles precaching. Precaching gives you the advantage of ensuring that, whenever your service worker is installed, a core set of URLs will be cached and kept up to date, even if they correspond to assets that haven't been used yet. (In other words, you get to cache important things in advance.)
>
> This differs from using routing + caching strategies to implement runtime caching, in which assets and HTML is only cached after a user visits a page that uses them for the first time.
>
> Precaching offers a number of advantages, both in terms of "priming" the cache, as well as efficient updates (a network request is only required when something in the precache manifest actually changes). The downside of precaching is that it requires some integration with a build process, as it Workbox needs to create hash "fingerprints" of local assets on the filesystem in order to create its precache manifest. That might not be viable in a generic Wordpress solution, but I wanted to throw it out there if there are hooks into a local build process that could be taken advantage of.

Precaching is indeed something we haven't talked about yet, but I think that WordPress has some good tooling we can use here. For example, external scripts and stylesheets get registered with dependency system where each URL get a stable handle that does not change. Each dependency gets a version as well which is used to cache-bust the URL, but in the case of precaching we could strip the version (`ver`) parameter when generating the response since the service worker could handle the cache busting itself.

westonruter:
To elaborate, the precaching should be done for all registered scripts and styles, not just the ones that are enqueued. We'd also need to include the offline page (#23) among the precached URLs. 

[...]

Issue #9, opened 2018-06-12
Research a method for caching and using data offline


postphotos:
As a user, I should have a way to experience a website in poor internet conditions. There should be a way for me to easily save the content that the site has already loaded in cache and not lose it when I click on the next link. In this offline state, I should have meaningful content available to me.

AC1: Continue to extract the learnings from #1 and document approaches in this ticket.
AC2: Provide a prototype or a design approach in this ticket outlining how this should work.

kienstra:
Hi @postphotos 
Thanks for the great ACs. Offline commenting might be good to implement with this.

westonruter:
Offlin commenting I think is out of scope here. That really gets into the realm of background sync.

kienstra:
**Question About offline.php**

Hi @postphotos,
Does this issue mainly involve something like an `offline.php` page, for when there's no connection? Otherwise, this might be covered by #14 or #5.

westonruter:
@kienstra I think this is separate from the `offline.php`: it's building on top of having an offline template. The bare minimum for offline support is not covered in this issue, and that is to have a page that is served when the site is offline. Eventually this offline page could surface links to content that the user has saved offline, but that would require the capability to save offline pages first. So the first step is to create a new issue that is specifically about serving the offline page. And as @mehigh suggested, we should not limit it to an `offline.php` but consider giving the user the ability to designate which page is served as the offline page. This would allow them to select any page in their site along with any page template they desire.

kienstra:
**Request For New Issue**

Hi @postphotos,
When you have a chance, could you please open a new issue to capture @westonruter's [comment](https://github.com/xwp/pwa-wp/issues/9#issuecomment-401875699) above?

>...a new issue that is specifically about serving the offline page. And as @mehigh suggested, we should not limit it to an offline.php but consider giving the user the ability to designate which page is served as the offline page. This would allow them to select any page in their site along with any page template they desire.

mehigh:
This provides the most basic implementation which allows non-tech people to enable a PWA basic-checklist feature. At the most basic level selecting a page to use for offline gets you going. 
Plugins can take it a step further and define a custom template for the offline page with more bells and whistles, etc.
I feel core needs to provide a starting point, even a basic one is a huge step forward.

nico-martin:
[...]

PR #28, opened 2018-07-11
Add UI to select an offline page

### LINKED ISSUE #23: Add ability to select a default offline page to be served when the user or site is offline
##*User story: As a WordPress administrator, I want to be able to select a page that displays when visitors are offline.*

- [x] AC1 - As @mehigh suggested, create a UI to select the page that is displayed when offline. 
- ~[ ] AC2 - This page should be served when offline.~ This AC has moved to ticket #45.

This could be similar to the Settings > Privacy Settings page:

privacy-settings-page

The offline page UI could look something like:
progressive-web-apps-offline-page

This is based on @westonruter's suggestion in [#9 (comment)](https://github.com/xwp/pwa-wp/issues/9#issuecomment-401875699).

kienstra:
**WIP Pull Request**

* Mainly taken from the privacy policy page UI in `wp-admin/privacy.php`.
* @todos include adding the custom post status and adding the href value

Closes #23 

offline page ui

There's no "Use This Page" element as shown in [Issue 23 (comment)](https://github.com/xwp/pwa-wp/issues/23#issuecomment-404005382).

That's an `<input type="submit">`, which submits the `<form>`. The entire "Reading Settings" page is a form, which should probably only submit with the "Save Changes" element at the bottom.

If you try to save a post that doesn't exist (error mainly taken from the Privacy Page UI):

error-page-does-not-exist

If the post is in the trash (also mainly taken from `wp-admin/privacy.php`:

page-in-trash


wp-includes/class-wp-offline-page.php
```diff
@@ -0,0 +1,157 @@
...
+				'sanitize_callback' => array( $this, 'sanitize_callback' ),
+			)
+		);
+	}
+
+	/**
+	 * Sanitize callback for the setting.
+	 * Mainly taken from wp-admin/privacy.php.
+	 *
+	 * @todo: Ensure this is stored with a custom post status.
```
kienstra: For background, please see @westonruter's [https://github.com/xwp/pwa-wp/issues/9#issuecomment-403262531](https://github.com/xwp/pwa-wp/issues/9#issuecomment-403262531).

wp-includes/class-wp-offline-page.php
```diff
@@ -0,0 +1,157 @@
...
+		}
+	}
+
+	/**
+	 * Adds a settings field to the 'Reading Settings' page.
+	 */
+	public function settings_field() {
+		add_settings_field(
+			self::SETTING_ID,
+			__( 'Progressive Web App Offline Page', 'pwa' ),
```
westonruter: Instead of “Progressive Web App Offline Page” I think this should be something like:

>  Page displays when offline

This would align more with the existing:

> Your homepage displays

Then there can be a `p.description` element which explains that this is is for PWA, and how provide more details about how it works, maybe mentioning it is “like a 404 page but it shows when someone's internet is down (e.g. during flight or going through tunnel), as opposed to when content is not found.”

kienstra:
So far, this PR doesn't actually serve the offline page. It only saves the post ID in [WP_Offline_Page::OPTION_NAME](https://github.com/xwp/pwa-wp/pull/28/files#diff-d195b472d1de4998580649b1b6212753R26).

westonruter:
That's fine. I think it's good to work on the UX for selecting the offline page separately from the logic from actually serving the offline page. The serving of the offline page should probably incorporate Workbox (#5) and also it will probably involve precaching (#25) since we'll need to make sure the offline page and its required assets get added to the cache.

wp-includes/class-wp-offline-page.php
hellofromtonya: @westonruter Is this what you are thinking:

pwa-wp-pr28

kienstra:
**Offline Page**

Hi @hellofromtonya,
It's nice how you excluded the 'offline page' from the homepage `<select>` elements:

offline-page-select

Also, the "create a new one" link works well:

page-created-offline-link

hellofromtonya:
At this point, we have the following functionality in place with much more to do.

### Create New Offline Page

The `create new page` link is now working.  It follows the privacy policy's button strategy while using the hyperlink inside of a form button:
- inserts a new page with the title of "Offline Page" and "draft" status.
- assigns that page as the `page_for_offline`
- redirects you to that new page

TODOs
[ ] Create the shortcode (for the classic editor)
[ ] Create the block (for Gutenberg) 
[ ] pre-populate the content with either.

### Tagging the Offline Page in the List Table

screen shot 2018-07-13 at 3 28 21 pm copy

### Exclude Static Pages from Selection

As @kienstra showed above, the new changes exclude static pages from the page pick dropdowns.

- The offline page cannot be selected as the homepage, posts page, or privacy page.  To ensure, it is removed from those dropdowns.
- Similarly the homepage, posts page, and privacy page cannot be selected as the offline page.  All of those are excluded from the offline page dropdown.

### Error Messaging

Error handling occurs on:
- on saving settings
- loading the Reading page

Currently errors include:

1. when the selected offline page is in the trash:

screen shot 2018-07-13 at 3 48 43 pm

2. when no offline page is selected

screen shot 2018-07-13 at 3 52 20 pm

3. when the selected offline page has been deleted

screen shot 2018-07-13 at 3 53 08 pm

4. when an error occurs creating a new offline page

wp-includes/class-wp-offline-page.php
```diff
@@ -0,0 +1,116 @@
...
+	 * Gets the offline page's ID.
+	 *
+	 * @param bool $hard Optional. When true or no ID, requests ID from `get_option`; else, uses the ID property.
+	 *                   Default is `false`.
+	 *
+	 * @return int
+	 */
+	public function get_offline_page_id( $hard = false ) {
+		if ( $hard || $this->offline_page_id < 1 ) {
+			$this->offline_page_id = (int) get_option( self::OPTION_NAME, 0 );
```
westonruter: There's actually no need to store the `offline_page_id` in a variable. Since the option is autoloaded, it will aways be loaded from the cache and not from the DB.

wp-includes/class-wp-offline-page.php
```diff
@@ -0,0 +1,116 @@
...
+	/**
+	 * Gets the static pages by storing each page's ID into a property.
+	 *
+	 * @param bool $hard Optional. When true, rebuilds the static pages cache by getting the option; else, it returns
+	 *                   the property.  Default is `false`.
+	 *
+	 * @return array
+	 */
+	public function get_static_pages( $hard = false ) {
+		if ( ! $hard ) {
```
westonruter: Since these options are autoloaded, they will aways be loaded from the cache and not from the DB. So there is no need to add another cache.

wp-includes/class-wp-offline-page.php
```diff
@@ -0,0 +1,116 @@
...
+
+	/**
+	 * Array of the static pages.
+	 *
+	 * @var array
+	 */
+	protected $static_pages = array(
+		'page_on_front'              => 0,
+		'page_for_posts'             => 0,
+		'wp_page_for_privacy_policy' => 0,
```
westonruter: Shouldn't this be the same as `self::OPTION_NAME`?

tests/test-class-wp-offline-page-ui.php
```diff
@@ -0,0 +1,326 @@
...
+
+		$this->assertContains(
+			'<a href="' . admin_url( 'options-reading.php?action=create-offline-page' ) . '">create a new one</a>',
+			$output
+		);
+
+		// Check that it excludes the configured static pages.
+		update_option( 'page_on_front', (int) $page_ids[0] );
+		update_option( 'page_for_posts', (int) $page_ids[1] );
+		update_option( 'wp_page_for_privacy_policy', (int) $page_ids[2] );
```
westonruter: This string should instead be `\WP_Offline_Page::OPTION_NAME`.

wp-includes/class-wp-offline-page-ui.php
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	/**
+	 * Gets the configured static pages IDs.
+	 *
+	 * @param bool $join Optional. When true, a comma-delimited list is returned; else, an array is returned.
+	 *
+	 * @return array|string
+	 */
+	protected function get_static_pages_ids( $join = false ) {
```
westonruter: See note below about how `$join` isn't necessary since we can just `implode( ',', ... )` when calling.

Also, I don't think we need `get_static_pages_ids` either. We can just do:

```php
array_filter( $this->manager->get_static_pages() )
```

wp-includes/class-wp-offline-page-ui.php
```diff
@@ -0,0 +1,327 @@
...
+	protected function render_page_dropdown() {
+		wp_dropdown_pages(
+			array(
+				'name'              => esc_html( WP_Offline_Page::OPTION_NAME ),
+				/* Translators: %1$s: A long dash */
+				'show_option_none'  => sprintf( esc_html__( '%1$s Select %1$s', 'pwa' ), '&mdash;' ),
+				'option_none_value' => '0',
+				'selected'          => intval( $this->manager->get_offline_page_id() ),
+				'post_status'       => array( 'draft', 'publish' ),
+				'exclude'           => esc_html( $this->get_static_pages_ids( true ) ),
```
westonruter: This needs to not exclude the offline page, or else it won't be selectable.

wp-includes/class-wp-offline-page.php
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * WP_Offline_Page class.
+ *
+ * @package PWA
+ */
+
+/**
+ * This class manages the Offline Page.
+ */
+class WP_Offline_Page {
```
westonruter: This class is pretty small. Would it make more sense to have `WP_Offline_Page_UI` and `WP_Offline_Page_Excluder` be just integrated into this one class? It could be called `WP_Offline_Manager` perhaps.

wp-includes/class-wp-offline-page-ui.php
```diff
@@ -0,0 +1,327 @@
...
+	 * Creates a new offline page.
+	 *
+	 * @return int|WP_Error The page ID or WP_Error on failure.
+	 */
+	public function create_new_page() {
+
+		$page_id = wp_insert_post(
+			wp_slash( array(
+				'post_title'   => __( 'Offline', 'pwa' ),
+				'post_status'  => 'draft',
```
westonruter: As discussed, let's have this create a `publish`'ed page. If we allow selecting a draft page then it wouldn't be accessible to the user.

wp-includes/class-wp-offline-page-excluder.php
```diff
@@ -0,0 +1,114 @@
...
+	 * @param WP_Query $query The WP_Query instance.
+	 */
+	public function exclude_from_query( WP_Query $query ) {
+		if ( $this->is_offline_page_query( $query ) ) {
+			$query->is_404      = true;
+			$query->is_page     = false;
+			$query->is_singular = false;
+			$query->set( 'page_id', 0 );
+		} elseif ( $this->is_okay_to_exclude( $query ) ) {
+			$query->set( 'post__not_in', array( $this->manager->get_offline_page_id() ) );
```
westonruter: The `post__not_in` query var is advised against for performance reasons: https://vip.wordpress.com/documentation/performance-improvements-by-removing-usage-of-post__not_in/

Maybe that doesn't apply here.

wp-includes/class-wp-offline-page-excluder.php
```diff
@@ -0,0 +1,114 @@
...
+	 * Exclude the offline page from the query when doing a search on the frontend or on the backend Menus page.
+	 *
+	 * @param WP_Query $query The WP_Query instance.
+	 */
+	public function exclude_from_query( WP_Query $query ) {
+		if ( $this->is_offline_page_query( $query ) ) {
+			$query->is_404      = true;
+			$query->is_page     = false;
+			$query->is_singular = false;
+			$query->set( 'page_id', 0 );
```
westonruter: Are these being done in anticipation of there being an `is_offline`? 

I don't think that we should necessarily make these changes here. The offline page is a page, so `is_page()` should return `true`. But what we could do here is `$query->is_offline = $this->is_offline_page_query( $query )` to dynamically define a new member variable that can be looked at similarly to the others. We can then introduce a new global function:

```php
function is_offline() {
	global $wp_query;
	if ( ! isset( $wp_query ) ) {
		_doing_it_wrong( __FUNCTION__, __( 'Conditional query tags do not work before the query is run. Before then, they always return false.' ), '3.1.0' );
		return false;
	}
	return ! empty( $wp_query->is_offline );
}
```

And this `is_offline()` function can be used in the template to conditionally show content for offline responses.

wp-includes/class-wp-offline-page-excluder.php
```diff
@@ -0,0 +1,114 @@
...
+	protected function is_offline_page_query( WP_Query $query ) {
+		if ( $query->is_admin ) {
+			return false;
+		}
+
+		if ( ! $query->is_page || ! $query->is_singular ) {
+			return false;
+		}
+
+		return ( $this->manager->get_offline_page_id() === $query->get( 'get_id' ) );
```
westonruter: Is `get_id` right here?

I think this method should largely just do:

```php
return $query->is_singular() && $this->manager->get_offline_page_id() === (int) $query->get_queried_object_id();
```

wp-includes/class-wp-offline-page-ui.php
```diff
@@ -0,0 +1,327 @@
...
+		unset( $non_offline_static_pages[ WP_Offline_Page::OPTION_NAME ] );
+
+		wp_dropdown_pages(
+			array(
+				'name'              => esc_html( WP_Offline_Page::OPTION_NAME ),
+				/* Translators: %1$s: A long dash */
+				'show_option_none'  => sprintf( esc_html__( '%1$s Select %1$s', 'pwa' ), '&mdash;' ),
+				'option_none_value' => '0',
+				'selected'          => intval( $this->manager->get_offline_page_id() ),
+				'post_status'       => array( 'draft', 'publish' ),
```
westonruter: Per above, let's only include `publish` here.

wp-includes/class-wp-offline-page-ui.php
```diff
@@ -0,0 +1,327 @@
...
+	/**
+	 * Renders the page dropdown.
+	 */
+	protected function render_page_dropdown() {
+		$non_offline_static_pages = $this->manager->get_static_pages();
+		unset( $non_offline_static_pages[ WP_Offline_Page::OPTION_NAME ] );
+
+		wp_dropdown_pages(
+			array(
+				'name'              => esc_html( WP_Offline_Page::OPTION_NAME ),
```
westonruter: Should actually be `esc_attr()`.

wp-includes/class-wp-offline-page-ui.php
```diff
@@ -0,0 +1,327 @@
...
+	/**
+	 * Checks if there are any pages to display in the page dropdown.
+	 *
+	 * @return bool Whether there are pages to show in the dropdown.
+	 */
+	protected function has_pages() {
+		$query = new WP_Query( array(
+			'post_type'              => 'page',
+			'posts_per_page'         => 1,
+			'post_status'            => array( 'publish', 'draft' ),
```
westonruter: Per above, let's only include `publish` here.

wp-includes/class-wp-offline-page-excluder.php
westonruter: 👉 This is not accounting for whether there is any existing `post__not_in` query var set. So at the very least this should merge with any existing value.

hellofromtonya: @westonruter I consider that in my research.  Let's talk through it to see if we agree that this use case does not apply _or_ if there is a better way to comply but also be performant.

## The Concerns with `post__not_in` var

The concerns about the usage of `post__not_in` var include:

>the query cache hitrate will often be a lot lower

True.  But in this use case, we do not want the offline page included the caching as it should not be included or rendered on the frontend.

>make the query slower _if the exclusion list is large_.

Notice that _slower_ occurs _when the list is large_.  In this use case, our list consists of one (1) item, i.e. one page ID.

## The VIP Proposed Solution - Iterate.

>In almost all cases you can gain great speed improvements by requesting more posts and skipping the posts in PHP.

The proposed alternative solution then is to include the offline page in the query and then remove it or skip over it when iterating through the collection of posts.  

In our use case, this solution would require us to provide the means to exclude the offline page from the collection of posts _after_ we get it back from the database.  Let's think about what this means.

1. The offline page would be cached.  
That's not desirable in our use case.

2. We'd need to remove it from the array of posts.
We could hook into one of the posts filters, such as `posts_results`.  However, iterating over the collection to remove one item is not performant either.

## My Thoughts

I think our use case is the perfect use case for `post__not_in`.  Why?

1. PWA is proposed to be merged into core.
2. We need to exclude the offline page ID from all frontend queries and all searches.
3. It's more performant to exclude in the SQL query than to iterate over the collection of posts for removal.
4. We do not want the offline page to ever be in the cache.

IMO I think using `post__not_in` is the best strategy overall for this use case.

@westonruter What do you think?  Are there other ways we could explore for excluding it from the query that do not involve using `post__not_in`?

wp-includes/class-wp-offline-page.php
hellofromtonya: @westonruter This one is for the privacy page and not the Offline Page.  I was using it for the following purposes:

1. Target each of their page select dropdowns to exclude the Offline Page from it.
2. Exclude each of them from the offline page's select dropdown.

tests/test-class-wp-offline-page-ui.php
hellofromtonya: Privacy policy page is a special page too.

wp-includes/class-wp-offline-page-excluder.php
westonruter: Your reasoning resonates with me! ✅

wp-includes/class-wp-offline-page-ui.php
```diff
@@ -0,0 +1,327 @@
...
+
+	/**
+	 * Check if the offline page does not exist.
+	 *
+	 * @param WP_Post|int $offline_page The offline page to check.
+	 *
+	 * @return bool
+	 */
+	protected function does_not_exist( $offline_page ) {
+		if ( is_int( $offline_page ) ) {
```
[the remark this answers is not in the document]
hellofromtonya: @westonruter good catch 👍

wp-includes/class-wp-offline-page.php
hellofromtonya: The reason I like doing that is performance as `get_option()` has quite a bit of functionality built into it including filtering.  It's tiny nanoseconds, I know.

I typically think about if the value will change.  If no, then I like to cache it on the object to reuse instead of recalling `get_option()` over and over again.

westonruter: That's true, but then it makes it more complicated when the value changes as then you have to invalidate multiple caches. Another plugin could update the option and then we'd not know about it unless we always the id _hardly_, which invalidates the benefit of having the additional cache in the first place. In core the preference is probably going to be relying on the internal object cache as opposed to having another caching layer.

tests/test-class-wp-offline-page-ui.php
westonruter: @hellofromtonya Oh, my mistake! I mistakenly read `wp_page_for_privacy_policy` as `wp_page_for_offline_page` or something. I actually hadn't seen the `wp_page_for_privacy_policy` option before, so I mistakenly glossed over it.

wp-includes/class-wp-offline-page.php
westonruter: Your right. I totally missed this.

wp-includes/class-wp-offline-page.php
westonruter: That sounds fine with me. We can always revisit later if it makes sense to merge into a single class for the core merge. But for now we can continue as it is.

wp-includes/class-wp-offline-page.php
```diff
@@ -64,53 +46,29 @@ public function __construct() {
...
-		foreach ( $this->static_pages as $option_name => $default ) {
-			$this->static_pages[ $option_name ] = get_option( $option_name, $default );
-		}
-
-		return $this->static_pages;
+	public function get_static_pages() {
+		return array(
+			'page_on_front'   => (int) get_option( 'page_on_front', 0 ),
+			'page_for_posts'  => (int) get_option( 'page_for_posts', 0 ),
+			self::OPTION_NAME => $this->get_offline_page_id(),
```
westonruter: The `wp_page_for_privacy_policy` needs to be restored here.

tests/test-class-wp-offline-page-ui.php
```diff
@@ -176,8 +176,7 @@ public function test_render_settings() {
 		// Check that it excludes the configured static pages.
 		update_option( 'page_on_front', (int) $page_ids[0] );
 		update_option( 'page_for_posts', (int) $page_ids[1] );
-		update_option( 'wp_page_for_privacy_policy', (int) $page_ids[2] );
-		$this->manager->get_static_pages( true );
+		update_option( WP_Offline_Page::OPTION_NAME, (int) $page_ids[2] );
```
westonruter: I mistakenly equated `wp_page_for_privacy_policy` with `WP_Offline_Page::OPTION_NAME` here.

wp-includes/class-wp-offline-page-excluder.php
hellofromtonya: >Are these being done in anticipation of there being an is_offline?

@westonruter No, it's meant to prevent the offline page from being displayed on the frontend. 
 It's forcing a 404.  It was added to provide the functionality [discussed here in Issue #23](https://github.com/xwp/pwa-wp/issues/23#issuecomment-404395715).

>I don't think that we should necessarily make these changes here.

Okay.  Let's think about this.  Maybe we don't care if the default offline page is requestable or viewable on the frontend.

- We're already excluding it from the query via `post__not_in`.
- We exclude it from the menus.

What's the harm of viewing it on the frontend?  If no or little, then there's no need to set a 404 here.

>And this is_offline() function can be used in the template to conditionally show content for offline responses.

Agreed.  I was thinking about adding that functionality in a separate PR.  We'll want pages/content be marked as `offline` and then set a var to note it.

westonruter: I think there is no need to prevent the offline page from being accessible if someone knows the URL. If the user can access it directly by the URL, that is fine. In fact, that should be actually desirable because it would allow for someone to easily preview their offline page and make changes. The main thing is just that we don't expose the offline page in menus, page lists, searches, etc. In case someone links to the offline page then we could add `wp_no_robots()` to run in `wp_head` for that specific page.

wp-includes/class-wp-offline-page-excluder.php
hellofromtonya: @westonruter It is checking `is_singular` above in the guard clause:

```php
		if ( ! $query->is_page || ! $query->is_singular ) {
			return false;
		}
```

I'll switch that to `$query->is_singular()` instead of `$query->is_singular`.

westonruter: Note that `is_singular` implies `is_page`, so only one of the conditions is needed.

But I was more drawing attention to `$query->get( 'get_id' )`. I haven't seen `get_id` before. I don't believe that is right, and `get_queried_object_id()` is what you intend?

wp-includes/class-wp-offline-page-excluder.php
hellofromtonya: Bam, done in [commit ab4cbc6](https://github.com/xwp/pwa-wp/pull/28/commits/ab4cbc66a3d61e809da99451d2e1a63f17cfa66a).  The _default_ offline page will now render on the frontend.  That makes it easier.  

> In case someone links to the offline page then we could add wp_no_robots() to run in wp_head for that specific page.

Included in this [commit 9669026](https://github.com/xwp/pwa-wp/pull/28/commits/96690264eed185fd748ce763d3049aaa2e74c655).

wp-includes/class-wp-offline-page-excluder.php
```diff
@@ -0,0 +1,125 @@
...
+	protected function is_offline_page_query( WP_Query $query ) {
+		if ( $query->is_admin ) {
+			return false;
+		}
+
+		if ( ! $query->is_singular() ) {
+			return false;
+		}
+
+		return $this->manager->get_offline_page_id() === (int) $query->get_queried_object_id();
```
westonruter: Minor thing but we could combine these:

```php
return (
    ! $query->is_admin
    &&
    $query->is_singular()
    &&
    $this->manager->get_offline_page_id() === (int) $query->get_queried_object_id()
);
```

wp-includes/class-wp-offline-page-excluder.php
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	/**
+	 * Exclude the offline page from the query when doing a search on the frontend or on the backend Menus page.
+	 *
+	 * @param WP_Query $query The WP_Query instance.
+	 */
+	public function exclude_from_query( WP_Query $query ) {
+		if ( $this->is_offline_page_query( $query ) ) {
+			add_action( 'wp_head', 'wp_no_robots' );
```
westonruter: I don't think this is quite right because it could result in `wp_no_robots` being added unintentionally. Consider, for example, a request to a normal post on a site which at `init` or some other early hook does (for some reason):

```php
$offline_query = new WP_Query( array(
    'page_id' => get_option( WP_Offline_Page::OPTION_NAME )
) );
```

This will result in the page getting `wp_no_robots` added since the `parse_query` action runs for every instance of `WP_Query` whether it is the main query or not. 

It would be better to separate the concerns here to add a new method like:

```php
function show_no_robots_on_offline_page() {
    global $wp_query;
    if ( $this->is_offline_page_query( $wp_query ) ) {
        wp_no_robots();
    }
}
```

And then add this method to the `wp_head` action unconditionally in `WP_Offline_Page_Excluder::init()`.

wp-includes/class-wp-offline-page-excluder.php
hellofromtonya: We could.  I tend to break them out separately.  But we can definitely combine.

wp-includes/class-wp-offline-page-excluder.php
hellofromtonya: It is giving a side effect. I was trying to avoid hooking in again and rerunning the check.  But I do like the separation.  Agree on your points.

hellofromtonya: @westonruter Changes added in [commit af9aedd](https://github.com/xwp/pwa-wp/pull/28/commits/af9aedd1c57cff158d5a4b000ef781ce5b054661).

wp-includes/class-wp-offline-page-excluder.php
hellofromtonya: @westonruter Changes added in [commit af9aedd](https://github.com/xwp/pwa-wp/pull/28/commits/af9aedd1c57cff158d5a4b000ef781ce5b054661).

Issue #17, opened 2018-06-26
Create a UI to force HTTPS if it's supported


kienstra:
**Please see the "Revised ACs" [comment](https://github.com/xwp/pwa-wp/issues/17#issuecomment-400434908) for the ACs, instead of this comment.**

Hi @westonruter and @postphotos,
What do you think about these ACs? Feel free to edit this.

**User Story: As a WordPress administrator, I want a UI to ensure that my site uses HTTPS if it's supported.**

- [x] AC1. Use the HTTPS detection logic from Issue #6
- [x] AC2. On the General Settings page, add a `<tr>` for HTTPS support (use JavaScript to move it after the home/site URLs)
- [x] AC3. If HTTPS is supported, but the WordPress Address (URL) and Site Address (URL) don't use HTTPS,  display this notice:
<img width="960" alt="possible-notice" src="https://user-images.githubusercontent.com/4063887/41925762-239456ce-7933-11e8-8148-d92331469d04.png">

- [x] AC4. On clicking "Click here," the WordPress Address (URL) and the Site Address (URL) change to HTTPS, and there's a redirect from HTTP to HTTPS
- [x] AC5. If HTTPS is not supported, display:
<img width="836" alt="not-supported" src="https://user-images.githubusercontent.com/4063887/41926060-db1e7392-7933-11e8-87a5-b28a4f86e351.png">

- [ ] AC6. Clicking "Recheck" triggers the loopback request from #6. Maybe there should be a notice at the top, like "HTTPS is still not supported" or "HTTP is supported."

- [x] AC7. If HTTPS is supported, and the WordPress Address (URL) and Site Address (URL) use HTTPS,  display:

<img width="1370" alt="https-supported" src="https://user-images.githubusercontent.com/4063887/41926277-766ceb58-7934-11e8-9153-b58dc36cc954.png">

- [ ] AC8. If HTTPS is enabled, show the user a toggle for HSTS (see issue #18).

- [x] AC9. Propose a UI and match the designs provided to meet WP core standards and look-and-feel, all to load on the general settings page.

westonruter:
Instead of “click here” I think we should be using links with descriptive text like “re-check” or “set HTTPS”. Also, buttons may be more appropriate than links in some cases.

> AC6

Instead of “not supported” and “supported” I think it should be “Support detected” and “Support not detected”.

There should probably be some link off to somewhere to get more information about HTTPS if support is not detected.

> AC4

[...]

PR #30, opened 2018-07-12
Release 0.1.0

westonruter:
* Update readme with overview of the plugin's status
* Add changelog
* Add build tools
* Add contributing docs (adapted from AMP plugin)
* Add [plugin header and plugin icon](https://developer.wordpress.org/plugins/wordpress-org/plugin-assets/).


westonruter:
@jwold would you help with the [plugin header and plugin icon](https://developer.wordpress.org/plugins/wordpress-org/plugin-assets/)?

bin/verify-version-consistency.php
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env php
+<?php
```
kienstra: This is really useful. It might be nice to add something like this in `wp-dev-lib`.

readme.md
```diff
@@ -1,30 +1,154 @@
...
+<blockquote cite="https://developers.google.com/web/progressive-web-apps/">
+Progressive Web Apps are user experiences that have the reach of the web, and are:
 
-* [Service Workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) (see [#36995](https://core.trac.wordpress.org/ticket/36995))
-* [Web App Manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest) (see [#43328](https://core.trac.wordpress.org/ticket/43328))
-* HTTPS (see [#28521](https://core.trac.wordpress.org/ticket/28521))
+<ul>
+<li><a href="https://developers.google.com/web/progressive-web-apps/#reliable">Reliable</a> - Load instantly and never show the downasaur, even in uncertain network conditions.</li>
+<li><a href="https://developers.google.com/web/progressive-web-apps/#fast">Fast</a> - Respond quickly to user interactions with silky smooth animations and no janky scrolling.</li>
+<li><a href="https://developers.google.com/web/progressive-web-apps/#engaging">Engaging</a> - Feel like a natural app on the device, with an immersive user experience.</li>
```
kienstra: These 3 points are a good way to summarize this.

contributing.md
```diff
@@ -0,0 +1,71 @@
...
+
+This will create an `pwa.zip` in the plugin directory which you can install. The contents of this ZIP are also located in the `build` directory which you can `rsync` somewhere as well.
+
+To create a build of the plugin as it will be deployed to WordPress.org, run:
+
+```bash
+npm run build-release
+```
+
+Note that this will currently take much longer than a regular build because it generates the files required for translation. You also must have WP-CLI installed with the [`i18n-command` package](https://github.com/wp-cli/i18n-command).
```
kadamwhite: Let’s also link out to WP-CLI; I’d suggest changing the grammar here to “To run this command you must also have installed both [WP-CLI](#) and the [`i18n-command` package](#)`

contributing.md
```diff
@@ -0,0 +1,71 @@
...
+$ phpunit --coverage-html /tmp/report
+```
+
+When you push a commit to your PR, Travis CI will run the PHPUnit tests and sniffs against the WordPress Coding Standards.
+
+## Creating a Release
+
+Contributors who want to make a new release, follow these steps:
+
+1. Do `npm run build-release` and install the `pwa.zip` onto a normal WordPress install running a stable release build; do smoke test to ensure it works.
```
kadamwhite: We’re at a juncture where a “normal” wp build may or may not have SSL. Should we call that out?

contributing.md
```diff
@@ -0,0 +1,71 @@
...
+2. Bump plugin versions in `package.json` (×1), `package-lock.json` (×1, just do `npm install` first), `composer.json` (×1), and in `pwa.php` (×2: the metadata block in the header and also the `PWA_VERSION` constant).
+3. Add changelog entry to readme.
+4. Draft blog post about the new release.
```
kadamwhite: Where should this be drafted?

westonruter: Good point. For the AMP plugin we've been blogging at make.xwp.co, but for this feature plugin it would make sense to blog at Make/Core.

contributing.md
westonruter: The plugin should be soon detecting if a WordPress install is being served as HTTPS and warn the user if it isn't (more or less #17). So in that case testing on a WordPress install which does not have SSL would actually be beneficial to test.

contributing.md
westonruter: Good changes. But I copied this bit from the AMP plugin and it isn't relevant used in the PWA plugin (not yet anyway). So I'll just delete the paragraph for now, but I'll apply your patch to AMP.

westonruter:
@kadamwhite how do the changes in ba51eb1 look to clarify the development environment question?

kadamwhite:
@westonruter apologies for the belated response but those changes look superb, thank you!

PR #32, opened 2018-07-13
Spelling correction

josevarghese:
I think the word is installed, not unstalled


westonruter:
Changes amended into https://github.com/xwp/pwa-wp/pull/31

PR #36, opened 2018-07-25
HTTPS Settings UI 

### LINKED ISSUE #17: Create a UI to force HTTPS if it's supported

kienstra:
This is the work for the HTTPS UI.

* Creates the class `WP_HTTPS_UI` and a test class for it.
* Begins to output the UI, including radio buttons.

Closes #17


wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+	/**
+	 * Sanitizes the raw setting value.
+	 *
+	 * @param string $raw_setting The setting before sanitizing it.
+	 * @return string|null The setting, or null if it's not valid.
+	 */
+	public function sanitize_callback( $raw_setting ) {
+		if ( self::OPTION_SELECTED_VALUE === $raw_setting || '0' === $raw_setting ) {
+			return $raw_setting;
+		}
```
westonruter: I think this should return `get_option( self::UPGRADE_HTTPS_OPTION )` if the sanitization fails. But then again, why not just use `wp_validate_boolean()` as the `sanitize_callback`?

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+
+	/**
+	 * Renders the HTTPS settings in /wp-admin on the Reading Settings page.
+	 */
+	public function render_settings() {
+		$upgrade_https_value           = get_option( self::UPGRADE_HTTPS_OPTION );
+		$upgrade_insecure_content      = get_option( self::UPGRADE_INSECURE_CONTENT_OPTION );
+		$https_more_details            = sprintf(
+			'<a href="%s">%s</a>',
+			'https://make.wordpress.org/support/user-manual/web-publishing/https-for-wordpress/',
```
westonruter: This could be translatable to allow locale-specific docs.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+			esc_html__( 'More details', 'pwa' )
+		);
+		$insecure_content_more_details = sprintf(
+			'<a href="%s">%s</a>',
+			'https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content/How_to_fix_website_with_mixed_content',
```
westonruter: This should be translatable to allow locale-specific docs.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+		/**
+		 * Todo: change this description based on what insecure content is found.
+		 *
+		 * Have a separate description for when there's only passive insecure content.
+		 * And one for when there's active and passive insecure content.
+		 *
+		 * @see https://github.com/xwp/pwa-wp/issues/17#issuecomment-406188455
+		 */
+		/* Translators: %s: a link for more details */
+		$insecure_content_description = esc_html__( 'Your home page doesn’t contain insecure URLs. However, there may be URLs on other pages that could be blocked. %s', 'pwa' );
```
westonruter: Using a UTF-8 `’` in the string can have unintended effects on non-UTF-8 installs. It should be an HTML entity instead. For instance: https://github.com/WordPress/wordpress-develop/blob/927cd42632207fb78cf463a3a3f6df4bcd58e7b9/src/wp-includes/class-wp-customize-nav-menus.php#L1028

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+
+	/**
+	 * Whether the options indicate that the site is currently using HTTPS.
+	 *
+	 * @return boolean
+	 */
+	public function is_currently_https() {
+		$urls = array( get_option( 'home' ), get_option( 'siteurl' ) );
+		foreach ( $urls as $url ) {
+			if ( 0 !== strpos( $url, self::HTTPS_PROTOCOL ) ) {
```
westonruter: Note that `strpos()` is going to search the entire string needlessly. If wanting to make this more performant (almost trivially) you could do `self::HTTPS_PROTOCOL !== substr( $url, 0, strlen( self::HTTPS_PROTOCOL ) )`. But that's hard to read. Maybe better to to use `wp_parse_url()` instead:

```php
'https' !== wp_parse_url( $url, PHP_URL_SCHEME )
```

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+		endif;
+	}
+
+	/**
+	 * Whether the options indicate that the site is currently using HTTPS.
+	 *
+	 * @return boolean
+	 */
+	public function is_currently_https() {
+		$urls = array( get_option( 'home' ), get_option( 'siteurl' ) );
```
westonruter: Instead of using the underlying `get_option()` calls it would be better to use `home_url()` and `site_url()` since additional filters apply to their return values which need to be accounted for. That is, unless the goal is to get the underlying value stored in the DB.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+		else :
+			/* Translators: %s: HTTPS more details link */
+			echo wp_kses_post( sprintf( __( 'Your site doesn\'t look to support HTTPS. %s', 'pwa' ), $https_more_details ) );
+		endif;
+	}
+
+	/**
+	 * Whether the options indicate that the site is currently using HTTPS.
+	 *
+	 * @return boolean
```
westonruter: Per https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/ the `@return` should:

> contain all possible return types, and a description for each. Use a period at the end.

Also `bool` should be used instead of `boolean`. So this should be:

```
@return bool Whether currently HTTPS.
```

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+			<p>
+				<label><input name="<?php echo esc_attr( self::UPGRADE_INSECURE_CONTENT_OPTION ); ?>" type="radio" <?php checked( $upgrade_insecure_content, self::OPTION_SELECTED_VALUE ); ?> value="<?php echo esc_attr( self::OPTION_SELECTED_VALUE ); ?>"><?php esc_html_e( 'Yes', 'pwa' ); ?></label>
+				<label><input name="<?php echo esc_attr( self::UPGRADE_INSECURE_CONTENT_OPTION ); ?>" type="radio" <?php checked( self::OPTION_SELECTED_VALUE !== $upgrade_insecure_content ); ?> value="0" ><?php esc_html_e( 'No', 'pwa' ); ?></label>
+			</p>
+			<p class="description">
+				<?php echo wp_kses_post( sprintf( $insecure_content_description, $insecure_content_more_details ) ); ?>
+			<p>
+			<?php
+		else :
+			/* Translators: %s: HTTPS more details link */
```
westonruter: Minor thing, but the convention in core is for “Translators” to be lower-cased: https://codex.wordpress.org/I18n_for_WordPress_Developers

Maybe this is important for GlotPress to parse them out.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+		$insecure_content_description = esc_html__( 'Your home page doesn’t contain insecure URLs. However, there may be URLs on other pages that could be blocked. %s', 'pwa' );
+
+		/*
+		 * Todo: change !== to === as this is only like this for development.
+		 * The WP_HTTPS_Detection doesn't work with my local SSL certificate.
+		 * Also, change $this->is_currently_https() to ! $this->is_currently_https().
+		 * This is also for development only.
+		 * This main if block should only run if the site can use HTTPS, but the 'home' and 'siteurl' options aren't HTTPS.
+		 */
+		if ( $this->is_currently_https() && true !== get_option( WP_HTTPS_Detection::HTTPS_SUPPORT_OPTION_NAME ) ) :
```
westonruter: I believe that `get_option()` always returns a string when it is a scalar value, since `option_value` is a text field. So this would be better to just check if `get_option( WP_HTTPS_Detection::HTTPS_SUPPORT_OPTION_NAME )` is truthy I believe.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+			self::OPTION_GROUP
+		);
+	}
+
+	/**
+	 * Renders the HTTPS settings in /wp-admin on the Reading Settings page.
+	 */
+	public function render_settings() {
+		$upgrade_https_value           = get_option( self::UPGRADE_HTTPS_OPTION );
+		$upgrade_insecure_content      = get_option( self::UPGRADE_INSECURE_CONTENT_OPTION );
```
westonruter: Probably good to cast the previous two values as `bool`.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	/**
+	 * Converts a URL from HTTP to HTTPS
+	 *
+	 * @param string $url The URL to convert.
+	 * @return string $url The converted URL.
+	 */
+	public function convert_to_https( $url ) {
+		return str_replace( 'http://', self::HTTPS_PROTOCOL, $url );
```
westonruter: This is not as robust as it could be. What if I have a URL like: `https://example.com/redirect/url/http://example.org/`. This will not then work as expected. So it would be better to make sure only the beginning is converted, for example:

```php
return preg_replace( '#^http(?=://)#', 'https', $url );
```

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,220 @@
...
+	 */
+	public function filter_site_url_and_home() {
+		if ( self::OPTION_SELECTED_VALUE === get_option( self::UPGRADE_HTTPS_OPTION ) ) {
+			add_filter( 'option_home', array( $this, 'convert_to_https' ), 11 );
+			add_filter( 'option_siteurl', array( $this, 'convert_to_https' ), 11 );
+		}
+	}
+
+	/**
+	 * Converts a URL from HTTP to HTTPS
```
westonruter: Should end in a period per docs coding standards.

kienstra: Thanks, [fbd25](https://github.com/xwp/pwa-wp/pull/36/commits/fbd25818243cbd6f763b7b7406c318c9ba64ba20) adds the period.

wp-includes/class-wp-https-ui.php
kienstra: Thanks, [fbd25](https://github.com/xwp/pwa-wp/pull/36/commits/fbd25818243cbd6f763b7b7406c318c9ba64ba20) changes `Translators` to `translators`.

wp-includes/class-wp-https-ui.php
kienstra: Sure, this could use `wp_validate_boolean` as the `sanitize_callback`. It's probably a good solution, though it will return `true` for most strings, not just `'1'`.

The expected values here are `'1'` [if Yes is selected](https://github.com/xwp/pwa-wp/pull/36/files#diff-cd06676b91b7dea56c702cb7bc9e7ec9R166), and `'0'` [if No is selected](https://github.com/xwp/pwa-wp/pull/36/files#diff-cd06676b91b7dea56c702cb7bc9e7ec9R167). 

That validation callback would return `true` for almost any string other than `'0'` or `'false'`, and `false` for the string `'0'`.

wp-includes/class-wp-https-ui.php
kienstra: Thanks, [fbd258](https://github.com/xwp/pwa-wp/pull/36/commits/fbd25818243cbd6f763b7b7406c318c9ba64ba20) changes the `@return` tag.

wp-includes/class-wp-https-ui.php
kienstra: Thanks, [fbd258](https://github.com/xwp/pwa-wp/pull/36/commits/fbd25818243cbd6f763b7b7406c318c9ba64ba20) uses the character `&#8217;`.

wp-includes/class-wp-https-ui.php
westonruter: Yeah, that should be fine.

kienstra: Thanks, [https://github.com/xwp/pwa-wp/pull/36/commits/c2cbdeea2902ffcbaa94bdc21a857dacb16f43bb](https://github.com/xwp/pwa-wp/pull/36/commits/c2cbdeea2902ffcbaa94bdc21a857dacb16f43bb) passes `wp_validate_boolean` as the `sanitize_callback`.

wp-includes/class-wp-https-ui.php
kienstra: Thanks, [https://github.com/xwp/pwa-wp/pull/36/commits/1326dc1d04ca71c9559fc65e176f057939622222](https://github.com/xwp/pwa-wp/pull/36/commits/1326dc1d04ca71c9559fc65e176f057939622222) adds translation for this URL.

wp-includes/class-wp-https-ui.php
kienstra: Thanks, [https://github.com/xwp/pwa-wp/pull/36/commits/1326dc1d04ca71c9559fc65e176f057939622222](https://github.com/xwp/pwa-wp/pull/36/commits/1326dc1d04ca71c9559fc65e176f057939622222) adds translation for this URL.

wp-includes/class-wp-https-ui.php
kienstra: Sure, [https://github.com/xwp/pwa-wp/pull/36/commits/c2cbdeea2902ffcbaa94bdc21a857dacb16f43bb](https://github.com/xwp/pwa-wp/pull/36/commits/c2cbdeea2902ffcbaa94bdc21a857dacb16f43bb) casts them with `(bool)`.

wp-includes/class-wp-https-ui.php
kienstra: Thanks, [https://github.com/xwp/pwa-wp/pull/36/commits/cef3e592a5c8e837c77f2dd38984507c7dc4b953](https://github.com/xwp/pwa-wp/pull/36/commits/cef3e592a5c8e837c77f2dd38984507c7dc4b953) evaluates whether it's 'truthy.'

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,233 @@
...
+					/* translators: %s: a link for more details */
+					esc_html__( 'HTTPS is essential to securing your WordPress site, we strongly suggest upgrading to HTTPS on your site. %s', 'pwa' ),
+					$https_more_details
+				) );
+				?>
+			</p>
+			<p style="margin-top: 20px;"><strong><?php esc_html_e( 'HTTPS Upgrade', 'pwa' ); ?></strong></p>
+			<p>
+				<label><input name="<?php echo esc_attr( self::UPGRADE_HTTPS_OPTION ); ?>" type="radio" <?php checked( $upgrade_https_value ); ?> value="<?php echo esc_attr( self::OPTION_SELECTED_VALUE ); ?>"><?php esc_html_e( 'Yes', 'pwa' ); ?></label>
+				<label><input name="<?php echo esc_attr( self::UPGRADE_HTTPS_OPTION ); ?>" type="radio" <?php checked( ! $upgrade_https_value ); ?> value="0" ><?php esc_html_e( 'No', 'pwa' ); ?></label>
```
westonruter: I think a checkbox should be used instead of radio buttons. I checked the various admin screens and I don't see there being any examples of Yes/No radio buttons being used. Radio buttons seem to only be used in core when there are more than two options. (The exception here is the Yes/No toggles in the Publish metabox in the classic editor, but even those radio buttons are going away in favor of a toggle in Gutenberg.)

With this done, then to detect for the value going into the `self::UPGRADE_HTTPS_OPTION` option, it should be `isset( $_POST[self::UPGRADE_HTTPS_OPTION] )` as opposed to looking at the value.

wp-includes/class-wp-https-ui.php
kienstra: Thanks, [https://github.com/xwp/pwa-wp/pull/36/commits/330ae829561bcbf99f1741f0e7437c5dcb3b3b4d](https://github.com/xwp/pwa-wp/pull/36/commits/330ae829561bcbf99f1741f0e7437c5dcb3b3b4d) uses your `preg_replace()` snippet above, and adds a test assertion.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,233 @@
...
+ * WP_HTTPS_UI class.
+ */
+class WP_HTTPS_UI {
+
+	/**
+	 * The option group for the offline page option.
+	 *
+	 * @var string
+	 */
+	const OPTION_GROUP = 'reading';
```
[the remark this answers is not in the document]
kienstra: Sounds good. [https://github.com/xwp/pwa-wp/pull/36/commits/f40eb5856cb134e343f670203bc1c4a87efd64a2](https://github.com/xwp/pwa-wp/pull/36/commits/f40eb5856cb134e343f670203bc1c4a87efd64a2) moves the UI to the 'General Settings' page.

wp-includes/class-wp-https-ui.php
kienstra: Thanks, [https://github.com/xwp/pwa-wp/pull/36/commits/f40eb5856cb134e343f670203bc1c4a87efd64a2](https://github.com/xwp/pwa-wp/pull/36/commits/f40eb5856cb134e343f670203bc1c4a87efd64a2) uses your snippet above.

wp-includes/class-wp-https-ui.php
kienstra: Yeah, checkboxes seem more natural for Yes/No options. There was some discussion in [https://github.com/xwp/pwa-wp/issues/17#issuecomment-404330851](https://github.com/xwp/pwa-wp/issues/17#issuecomment-404330851) about the radio buttons showing that this is a reversible decision.

wp-includes/class-wp-https-ui.php
kienstra: Sure, [https://github.com/xwp/pwa-wp/pull/36/commits/2b9eceb5fb1778521cf6e9d16265cebff36af0a5](https://github.com/xwp/pwa-wp/pull/36/commits/2b9eceb5fb1778521cf6e9d16265cebff36af0a5) uses `home_url()` and `site_url()` instead of the `get_option()` calls.

wp-includes/class-wp-https-ui.php
westonruter: Why would a checkbox imply that it is not a reversible decision?

kienstra: I think it's still clear with a checkbox that it's a reversible decision.

westonruter: Let's go with checkboxes then for consistency with the rest of the admin settings screens. /cc @jwold

kienstra: Sounds good, I'll make a commit for that.

kienstra: [https://github.com/xwp/pwa-wp/pull/36/commits/c8da8580e50d33205032fd6fc4ddd04b9615c3a7](https://github.com/xwp/pwa-wp/pull/36/commits/c8da8580e50d33205032fd6fc4ddd04b9615c3a7) changes the radio buttons to checkboxes:

checkboxes

It also [changes the validation](https://github.com/xwp/pwa-wp/commit/c8da8580e50d33205032fd6fc4ddd04b9615c3a7#diff-cd06676b91b7dea56c702cb7bc9e7ec9R103) callback to return `true` if `$_POST` value `isset()`, like:

```php
if ( isset( $_POST[ self::UPGRADE_HTTPS_OPTION ] ) ) { // WPCS: CSRF OK.
	return true;
}
```

If that's not what you had in mind in [https://github.com/xwp/pwa-wp/pull/36#discussion_r205622089](https://github.com/xwp/pwa-wp/pull/36#discussion_r205622089), I'll change this.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,251 @@
...
+
+	/**
+	 * Sanitization callback for the upgrade HTTPS option.
+	 *
+	 * @param string $raw_value The value to sanitize.
+	 * @return bool Whether the option is true or false.
+	 */
+	public function upgrade_https_sanitize_callback( $raw_value ) {
+		unset( $raw_value );
+		if ( isset( $_POST[ self::UPGRADE_HTTPS_OPTION ] ) ) { // WPCS: CSRF OK.
```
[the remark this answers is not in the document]
kienstra: Sure, [https://github.com/xwp/pwa-wp/pull/36/commits/af8f4c7cf658fbc88cac40d472e22b0c9786026c](https://github.com/xwp/pwa-wp/pull/36/commits/af8f4c7cf658fbc88cac40d472e22b0c9786026c) applies your suggestion of simply returning the `isset()` value.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,251 @@
...
+
+	/**
+	 * Sanitization callback for the upgrade insecure content option.
+	 *
+	 * @param string $raw_value The value to sanitize.
+	 * @return bool Whether the option is true or false.
+	 */
+	public function upgrade_insecure_content_sanitize_callback( $raw_value ) {
+		unset( $raw_value );
+		if ( isset( $_POST[ self::UPGRADE_INSECURE_CONTENT_OPTION ] ) ) { // WPCS: CSRF OK.
```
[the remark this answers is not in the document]
kienstra: Sure, [https://github.com/xwp/pwa-wp/pull/36/commits/af8f4c7cf658fbc88cac40d472e22b0c9786026c](https://github.com/xwp/pwa-wp/pull/36/commits/af8f4c7cf658fbc88cac40d472e22b0c9786026c) applies your suggestion.

kienstra:
Here's the current UI:

https-general-settings-page

kienstra:
**Question About Response Header**

Hi @westonruter,
Now that there's a method to [get_insecure_content()](https://github.com/xwp/pwa-wp/pull/36/files#diff-d46994826f028ee2342e91b535e70cb4R152), that could probably look at the response body that the loopback request in [WP_HTTPS_Detection::check_https_support()](https://github.com/xwp/pwa-wp/blob/4869e0e2fb616a78a705573cabdcbf79d1a45f57/wp-includes/class-wp-https-detection.php#L91) gets. 

But at the moment, `WP_HTTPS_Detection::check_https_support()` doesn't get the full HTML document, as [verify_https_check](https://github.com/xwp/pwa-wp/blob/4869e0e2fb616a78a705573cabdcbf79d1a45f57/wp-includes/class-wp-https-detection.php#L148) calls `wp_die()` on the `parse_query` action. 

Ideally, that would return the full HTML document, and [get_insecure_content()](https://github.com/xwp/pwa-wp/pull/36/files#diff-d46994826f028ee2342e91b535e70cb4R152) could find its insecure URLs.

What do you think about returning a response header [instead of calling wp_die()](https://github.com/xwp/pwa-wp/blob/4869e0e2fb616a78a705573cabdcbf79d1a45f57/wp-includes/class-wp-https-detection.php#L148) with the query var?

1. The loopback still makes a request [with self::REQUEST_QUERY_VAR](https://github.com/xwp/pwa-wp/blob/4869e0e2fb616a78a705573cabdcbf79d1a45f57/wp-includes/class-wp-https-detection.php#L95)
2. Instead of [looking for](https://github.com/xwp/pwa-wp/blob/4869e0e2fb616a78a705573cabdcbf79d1a45f57/wp-includes/class-wp-https-detection.php#L146) that query var on the `parse_query` action, look for it on `wp_headers`
3. If it's present, add a header with a [hash of it](https://github.com/xwp/pwa-wp/blob/4869e0e2fb616a78a705573cabdcbf79d1a45f57/wp-includes/class-wp-https-detection.php#L148)
4. Then, look for that header in [check_https_support()](https://github.com/xwp/pwa-wp/blob/4869e0e2fb616a78a705573cabdcbf79d1a45f57/wp-includes/class-wp-https-detection.php#L110)
5. The response should have the body, so [get_insecure_content()](https://github.com/xwp/pwa-wp/pull/36/files#diff-d46994826f028ee2342e91b535e70cb4R152) could find its insecure URLs

Thanks, Weston!

kienstra:
**Current UI**

active-insecure-content

kienstra:
**Screencast**

Here's a [screencast](https://drive.google.com/file/d/18Vlu1eUO_dRc91xx94X2N1RUO1q4aZoR/view?usp=sharing) of this.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,293 @@
...
+				for ( $i = 0; $i < self::INITIAL_URLS_SHOWN; $i++ ) :
+					if ( empty( $all_insecure_urls[ $i ] ) ) :
+						continue;
+					endif;
+					?>
+					<li><a href="<?php echo esc_attr( $all_insecure_urls[ $i ] ); ?>"><?php echo esc_html( $all_insecure_urls[ $i ] ); ?></a></li>
+				<?php endfor; ?>
+			</ul>
+			<?php if ( $total_urls_count > self::INITIAL_URLS_SHOWN ) : ?>
+				<details style="max-width: 400px; max-height: 200px; overflow-y: scroll">
```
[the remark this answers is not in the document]
kienstra: Thanks, here's the current UI, which has the `<details>` removed via https://github.com/xwp/pwa-wp/pull/36/commits/5ac8619f5450197930b9418a95736e4534395433:
scroll-through-insecure-urls

kienstra:
There's now a singular form when there's only 1 insecure URL:

singular-url

kienstra:
(Still WIP, but should be ready for code review soon)

tests/test-class-wp-https-detection.php
```diff
@@ -32,7 +53,7 @@ class Test_WP_HTTPS_Detection extends WP_UnitTestCase {
 	public function setUp() {
 		parent::setUp();
 		$this->instance = new WP_HTTPS_Detection();
-		add_filter( 'wp_die_handler', array( $this, 'get_handler_to_prevent_exiting' ), 11 );
```
kienstra: This was a previous approach to validating the loopback request.

wp-includes/class-wp-https-detection.php
```diff
@@ -107,14 +133,93 @@ public function check_https_support() {
...
+	}
+
+	/**
+	 * Gets whether the body has a manifest <link>, with the proper href value.
+	 *
+	 * @param string $body The body of the response.
+	 * @return boolean Whether the body has a proper manifest.
+	 */
+	public function has_proper_manifest( $body ) {
+		$libxml_previous_state = libxml_use_internal_errors( true );
```
kienstra: This prevents errors, like in [parsing HTML5 elements](http://php.net/manual/es/domdocument.loadhtml.php#118107).

wp-includes/class-wp-https-detection.php
```diff
@@ -25,77 +32,96 @@ class WP_HTTPS_Detection {
...
 	 */
-	public function update_option_https_support() {
-		$https_support = $this->check_https_support();
-		if ( is_bool( $https_support ) ) {
-			update_option( self::HTTPS_SUPPORT_OPTION_NAME, $https_support );
+	public function update_https_support_options() {
+		$https_support_response = $this->check_https_support();
+		update_option(
+			self::HTTPS_SUPPORT_OPTION_NAME,
+			! is_wp_error( $https_support_response ) && 200 === wp_remote_retrieve_response_code( $https_support_response )
```
kienstra: A `WP_Error` is possible if the request fails completely.

wp-includes/class-wp-https-detection.php
```diff
@@ -25,77 +32,96 @@ class WP_HTTPS_Detection {
...
+		);
+
+		if ( ! is_wp_error( $https_support_response ) ) {
```
kienstra: If the request failed, there's not going to be insecure content to parse.

kienstra:
**Steps To Use**

1. Install and activate the Insecure URLs plugin below (already done on the [staging site](https://content-xwp-pwa-testing.pantheonsite.io/))
2. Ensure the WordPress Address (URL) and Site Address (URL) are HTTP:
wp-address

3. If possible, visit the home page using HTTPS, like on the [staging site](https://content-xwp-pwa-testing.pantheonsite.io/). 

There should be console errors and an icon from mixed content that the Insecure URLs plugin outputs:
mixed-content-errors


4. On the 'General Settings' page, check 'Upgrade to secure connection':
check-this

5. You should see the insecure URLs above, and clicking 'Show more' should display 5 more at a time.

6. Visit the home page with HTTPS, like on the [staging site](https://content-xwp-pwa-testing.pantheonsite.io/)

7. There shouldn't be insecure content errors, because the `upgrade-insecure-requests` header upgrades them. They still cause 404 responses, because they don't actually exist and they're only for testing.
[PWAWP-17-no-insecure-content.pdf](https://github.com/xwp/pwa-wp/files/2315875/PWAWP-17-no-insecure-content.pdf)



**Insecure URLs Plugin For Testing**
```php
<?php
/**
 * Plugin Name: Insecure URLs
 * Description: To test the HTTPS upgrade feature in the PWA plugin, this enqueues insecure resources.
 */

add_action( 'wp_enqueue_scripts', function() {
	wp_enqueue_script( 'iu-foo-a', 'http://example.com/foo-a' );
	wp_enqueue_script( 'iu-foo-b', 'http://example.com/foo-b' );
	wp_enqueue_script( 'iu-foo-c', 'http://example.com/foo-c' );
	wp_enqueue_script( 'iu-foo-d', 'http://example.com/foo-d' );
	wp_enqueue_script( 'iu-foo-e', 'http://example.com/foo-e' );
	wp_enqueue_script( 'iu-bar-f', 'http://example.com/bar-f' );
	wp_enqueue_script( 'iu-baz-g', 'http://example.com/baz-g' );

	wp_enqueue_style( 'iu-example-a', 'http://example.com/example-a' );
	wp_enqueue_style( 'iu-example-b', 'http://example.com/example-b' );
	wp_enqueue_style( 'iu-example-c', 'http://example.com/example-c' );
	wp_enqueue_style( 'iu-example-d', 'http://example.com/example-d' );
	wp_enqueue_style( 'iu-example-e', 'http://example.com/example-e' );
	wp_enqueue_style( 'iu-example-f', 'http://example.com/example-f' );
	wp_enqueue_style( 'iu-example-g', 'http://example.com/example-g' );
	wp_enqueue_style( 'iu-example-h', 'http://example.com/example-h' );
} );

```

kienstra:
**Request For Review**

Hi @hellofromtonya,
Could you please review this PR? This plugin is deployed, up to the latest commit (https://github.com/xwp/pwa-wp/pull/36/commits/f7c3bb16411d143f38fb69582b8426fd5df01f1f), to:
http://content-xwp-pwa-testing.pantheonsite.io/

All [loopback requests](https://github.com/xwp/pwa-wp/blob/f7c3bb16411d143f38fb69582b8426fd5df01f1f/wp-includes/class-wp-https-detection.php#L122) have been returning `WP_Error` instances in my local VVV, and not only with this plugin. Ideally, you won't have this problem 😄It doesn't seem to exist on the [staging URL above](https://content-xwp-pwa-testing.pantheonsite.io/).

But if you do, here's a method I've added to `WP_HTTPS_Detection` to populate the insecure URLs (added to the `init` action):

```php

/**
 * Sets mock insecure content, only for development.
 *
 */
public function set_mock_insecure_content() {
	update_option(
		self::INSECURE_CONTENT_OPTION_NAME,
		array(
			'http://example.com/foo-passive',
			'http://example.com/bar-passive',
			'http://example.com/baz-passive',
			'http://example.com/var-passive',
			'http://example.com/something-passive',
			'http://example.com/example-passive',
			'http://example.com/this-passive',
			'http://example.com/mab-passive',
			'http://example.com/foo-active',
			'http://example.com/bar-active',
			'http://example.com/baz-active',
			'http://example.com/var-active',
			'http://example.com/ex-active',
			'http://example.com/something-active',
			'http://example.com/example-active',
			'http://example.com/this-active',
			'http://example.com/mab-active',
			'http://example.com/zoom-active',
			'http://example.com/this-active',
			'http://example.com/won-active',
		)
	);
}
```

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,342 @@
...
+
+		$insecure_urls_option = get_option( WP_HTTPS_Detection::INSECURE_CONTENT_OPTION_NAME );
+		$all_insecure_urls    = $insecure_urls_option;
+		$total_urls_count     = count( $all_insecure_urls );
+		$insecure_content_id  = 'insecure-content';
+		$show_more_button_id  = 'view-urls';
+		$insecure_urls_class  = 'insecure-urls';
+
+		// Exit if there is no insecure URL to display.
+		if ( ! $total_urls_count ) {
```
hellofromtonya: I'd suggest moving this guard clause up to line 177.  If there are no insecure URLs, bail out; else, then set the other `insecure_content_id` and other variable values and continue.

hellofromtonya:
>All loopback requests have been returning WP_Error instances in my local VVV, and not only with this plugin. Ideally, you won't have this problem 😄It doesn't seem to exist on the staging URL above.

@kienstra Same problem when using VVV and Local.  Hmm, interesting....

wp-includes/class-wp-https-detection.php
```diff
@@ -25,77 +32,95 @@ class WP_HTTPS_Detection {
...
+		'script',
+		'link',
+		'iframe',
+	);
 
 	/**
 	 * Inits the class.
 	 */
 	public function init() {
 		add_action( 'wp', array( $this, 'schedule_cron' ) );
```
hellofromtonya: @kienstra Is the idea here to _actively continue checking_ after the cron time regardless if HTTPS is enabled or not? Curious....

wp-includes/class-wp-https-ui.php
kienstra: Sure, https://github.com/xwp/pwa-wp/pull/36/commits/fc159a274426fae4d906fe596b39df86d0f4eb8a this `return` statement up.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,342 @@
...
+			(function ( $ ) {
+				// Move this UI under the Site Address (URL) <tr> on the General Settings page.
+				$( 'input[name=<?php echo esc_attr( self::UPGRADE_HTTPS_OPTION ); ?>]' ).parents( 'tr' ).insertAfter( $( 'label[for=home]' ).parents( 'tr') );
+			})( jQuery );
+		</script>
+		<?php
+
+		$insecure_urls_option = get_option( WP_HTTPS_Detection::INSECURE_CONTENT_OPTION_NAME );
+		$all_insecure_urls    = $insecure_urls_option;
+		$total_urls_count     = count( $all_insecure_urls );
```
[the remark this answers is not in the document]
kienstra: Good catch. This https://github.com/xwp/pwa-wp/pull/36/commits/fc159a274426fae4d906fe596b39df86d0f4eb8a addresses this.

tests/test-class-wp-https-detection.php
```diff
@@ -80,60 +138,102 @@ public function test_schedule_cron() {
...
+
+		// If one of these options has an HTTP URL, the method should return false.
+		update_option( 'siteurl', self::HTTPS_URL );
+		$this->assertFalse( $this->instance->is_currently_https() );
+
+		// If both of these options have an HTTPS URL, the method should return true.
+		update_option( 'siteurl', self::HTTPS_URL );
+		update_option( 'home', self::HTTPS_URL );
+		add_filter( 'set_url_scheme', array( $this, 'convert_to_https' ) );
+		$this->assertTrue( $this->instance->is_currently_https() );
```
hellofromtonya: @kienstra We need another set of assertions to validate having a proper manifest.

wp-includes/class-wp-https-detection.php
kienstra: Good point. This does keep checking even when HTTPS is enabled. Let me work on that.

tests/test-class-wp-https-detection.php
```diff
@@ -80,60 +138,102 @@ public function test_schedule_cron() {
...
+		$this->instance->update_https_support_options();
+		$this->assertFalse( get_option( WP_HTTPS_Detection::HTTPS_SUPPORT_OPTION_NAME ) );
 		remove_filter( 'http_response', array( $this, 'mock_error_response' ) );
 
 		// The response is a 301, so the option value should be false.
 		add_filter( 'http_response', array( $this, 'mock_incorrect_response' ) );
-		$this->instance->update_option_https_support();
+		$this->instance->update_https_support_options();
 		$this->assertFalse( get_option( WP_HTTPS_Detection::HTTPS_SUPPORT_OPTION_NAME ) );
 		remove_filter( 'http_response', array( $this, 'mock_incorrect_response' ) );
```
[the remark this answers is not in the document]
kienstra: Sure, https://github.com/xwp/pwa-wp/pull/36/commits/54b68e0f05c71371a03120376600a8b5f6ab942b adds assertions for the insecure content option.

tests/test-class-wp-https-detection.php
kienstra: Hi Tonya, https://github.com/xwp/pwa-wp/pull/36/commits/0fb5b7ef260d7829eafa90fc587554452fd13c12 now adds an assertion for there being an incorrect manifest.

This [assertion](https://github.com/xwp/pwa-wp/blob/0fb5b7ef260d7829eafa90fc587554452fd13c12/tests/test-class-wp-https-detection.php#L156) tested a proper manifest being output, [by adding](https://github.com/xwp/pwa-wp/blob/0fb5b7ef260d7829eafa90fc587554452fd13c12/tests/test-class-wp-https-detection.php#L154) `mock_successful_response` as a callback.

wp-includes/class-wp-https-detection.php
kienstra: Hi @hellofromtonya, https://github.com/xwp/pwa-wp/pull/36/commits/b7b06f5de738b7f028a47792dfe8c0d12ad81b95 prevents running the cron hook to update the HTTPS options if the `home` and `siteurl` are already HTTPS. The UI that `WP_HTTPS_UI` adds wouldn't show in that case.

hellofromtonya: @kienstra What if we also wrap the cron hooks within a conditional check to see if 

```
if ( ! $this->is_currently_https() ) {
    // hooks here.
}
```

Then we only hook if HTTPS is not enabled.

kienstra: Thanks, good idea. Is https://github.com/xwp/pwa-wp/pull/36/commits/07ce3cf050fb4f65eeb0965b8a048a70355803d1 what you had in mind?

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,340 @@
...
+	 * Only add this if the site can support HTTPS,
+	 * but the home and siteurl option values are not HTTPS (WordPress Address and Site Address).
+	 * This UI would not apply if those URLs are already HTTPS.
+	 */
+	public function add_settings_field() {
+		/*
+		 * Todo: add ! before $this->is_currently_https(), as this is only for development.
+		 * It allows developing this while the siteurl and home values are HTTPS.
+		 */
+		if ( ! $this->wp_https_detection->is_currently_https() ) {
```
hellofromtonya: @kienstra Can we remove this `Todo` note?

kienstra: Good point, https://github.com/xwp/pwa-wp/pull/36/commits/8c0f09f4b65f00563aad1135e8d9dd0279e1dfac removes the `Todo`.

kienstra:
Hi @hellofromtonya,
>On our staging server, we [have this post](http://content-xwp-pwa-testing.pantheonsite.io/2018/08/23/test-post-with-images/) that has http for the img element's src. Why aren't these URLs are showing in the list of insecure content?

hellofromtonya:
@westonruter This PR is ready for your review and/or merge.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,336 @@
...
+
+	/**
+	 * Sanitization callback for the upgrade HTTPS option.
+	 *
+	 * @param string $raw_value The value to sanitize.
+	 * @return bool Whether the option is true or false.
+	 */
+	public function upgrade_https_sanitize_callback( $raw_value ) {
+		unset( $raw_value );
+		return isset( $_POST[ self::UPGRADE_HTTPS_OPTION ] ); // WPCS: CSRF OK.
```
westonruter: This is because there is a checkbox being used?

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,336 @@
...
+			) );
+			?>
+		</p>
+		<p>
+			<label><input name="<?php echo esc_attr( self::UPGRADE_HTTPS_OPTION ); ?>" type="checkbox" <?php checked( $upgrade_https_value ); ?> value="<?php echo esc_attr( self::OPTION_CHECKED_VALUE ); ?>"><?php esc_html_e( 'Upgrade to secure connection', 'pwa' ); ?></label>
+		</p>
+		<script>
+			( function ( $ ) {
+				// Move this UI under the Site Address (URL) <tr> on the General Settings page.
+				$( 'input[name=<?php echo esc_attr( self::UPGRADE_HTTPS_OPTION ); ?>]' ).parents( 'tr' ).insertAfter( $( 'label[for=home]' ).parents( 'tr') );
```
westonruter: Use of `esc_attr` isn't right here. It should be `wp_json_encode()`, kinda, but really it's not necessary.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,336 @@
...
+
+					// If there are no more insecure URLs that are hidden, hide the 'Show more' button.
+					if ( ! $( '.<?php echo esc_attr( $insecure_urls_class ); ?>.hidden' ).length ) {
+						$( this ).addClass( 'hidden' );
+					}
+				} );
+			} )( jQuery );
+		</script>
+		<style>
+			.<?php echo esc_attr( $insecure_urls_class ); ?> {
```
westonruter: HTML entity-encoding doesn't work in CSS. Probably better to use `sanitize_key()` if really needed. But since these are the strings defined in the method itself, it's not essential to escape.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,336 @@
...
+				$( 'input[type=checkbox][name="<?php echo esc_attr( self::UPGRADE_HTTPS_OPTION ); ?>"]' ).on( 'change', function() {
+					$( '#<?php echo esc_attr( $insecure_content_id ); ?>' ).toggleClass( 'hidden' );
+				} );
+
+				$( '#<?php echo esc_attr( $show_more_button_id ); ?>' ).on( 'click', function( event ) {
+					event.preventDefault();
+					$( '.<?php echo esc_attr( $insecure_urls_class ); ?>.hidden' ).first().removeClass( 'hidden' );
+
+					// If there are no more insecure URLs that are hidden, hide the 'Show more' button.
+					if ( ! $( '.<?php echo esc_attr( $insecure_urls_class ); ?>.hidden' ).length ) {
```
westonruter: This should rather be:

```php
if ( ! $( <?php echo wp_json_encode( ".$insecure_urls_class.hidden" ); ?> ).length ) {
```

Use of `esc_attr()` inside of a `script` could have unexpected effects. But in the end, this variable is just a constant that you're using in the class.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	/**
+	 * Renders the HTTPS settings in /wp-admin on the General Settings page.
+	 */
+	public function render_https_settings() {
+		$upgrade_https_value = (bool) get_option( self::UPGRADE_HTTPS_OPTION );
+		$https_more_details  = sprintf(
+			'<a href="%s">%s</a>',
+			__( 'https://make.wordpress.org/support/user-manual/web-publishing/https-for-wordpress/', 'pwa' ),
```
westonruter: This should have `esc_url()` wrapping.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,336 @@
...
+
+		$insecure_content_id = 'insecure-content';
+		$show_more_button_id = 'view-urls';
+		$insecure_urls_class = 'insecure-urls';
+		$description         = sprintf(
+			/* translators: %s is a link for more details */
+			__( 'We found content on your site that wasn&#39;t loading correctly over HTTPS. While we will try to fix these links automatically, you might check to be sure your pages work as expected. %s', 'pwa' ),
+			sprintf(
+				'<a href="%s">%s</a>',
+				__( 'https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content/How_to_fix_website_with_mixed_content', 'pwa' ),
```
westonruter: This should have `esc_url()` wrapping.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,336 @@
...
+				__( 'More details', 'pwa' )
```
westonruter: Best to use `esc_html__()` here.

wp-includes/class-wp-https-ui.php
```diff
@@ -0,0 +1,336 @@
...
+
+	/**
+	 * Renders the HTTPS settings in /wp-admin on the General Settings page.
+	 */
+	public function render_https_settings() {
+		$upgrade_https_value = (bool) get_option( self::UPGRADE_HTTPS_OPTION );
+		$https_more_details  = sprintf(
+			'<a href="%s">%s</a>',
+			__( 'https://make.wordpress.org/support/user-manual/web-publishing/https-for-wordpress/', 'pwa' ),
+			__( 'More details', 'pwa' )
```
westonruter: Best to use `esc_html__()` here.

wp-includes/class-wp-https-ui.php
kienstra: Yes

wp-includes/class-wp-https-ui.php
kienstra: `$https_more_details` is [escaped inline with wp_kses_post()](https://github.com/xwp/pwa-wp/blob/8c0f09f4b65f00563aad1135e8d9dd0279e1dfac/wp-includes/class-wp-https-ui.php#L155):

```php
echo wp_kses_post( sprintf(
	/* translators: %s: a link for more details */
	__( 'HTTPS is essential to securing your WordPress site, we strongly suggest upgrading to it. %s', 'pwa' ),
	$https_more_details
) );
```

Should this use `esc_url()` and escape it inline?

westonruter: Yes as `wp_kses_post()` wont escape `&` to `&amp;`.

wp-includes/class-wp-https-ui.php
kienstra: Sure, https://github.com/xwp/pwa-wp/pull/36/commits/9e969c0475549792f504528a6e1041b9440a2262 removes the escaping here.

wp-includes/class-wp-https-ui.php
kienstra: Sure, https://github.com/xwp/pwa-wp/pull/36/commits/9e969c0475549792f504528a6e1041b9440a2262 removes this `esc_attr()`.

wp-includes/class-wp-https-ui.php
kienstra: Thanks, https://github.com/xwp/pwa-wp/pull/36/commits/2a87852716f186ea3e83d182cd3856837fb2c035 adds `esc_url()` here.

wp-includes/class-wp-https-ui.php
kienstra: Sure, https://github.com/xwp/pwa-wp/pull/36/commits/2a87852716f186ea3e83d182cd3856837fb2c035 adds `esc_url()`.

wp-includes/class-wp-https-ui.php
kienstra: Sure, https://github.com/xwp/pwa-wp/pull/36/commits/2a87852716f186ea3e83d182cd3856837fb2c035 adds `esc_html__()`.

wp-includes/class-wp-https-ui.php
kienstra: Sure, https://github.com/xwp/pwa-wp/pull/36/commits/2a87852716f186ea3e83d182cd3856837fb2c035 uses `esc_html__()`.

PR #37, opened 2018-07-27
[WIP] Add Workbox

westonruter:
See #5.

### DIFF
--- a/wp-includes/class-wp-service-workers.php
+++ b/wp-includes/class-wp-service-workers.php
@@ -55,12 +55,54 @@ class WP_Service_Workers extends WP_Scripts {
 	 * Initialize the class.
 	 */
 	public function init() {
+
+		$this->register(
+			'workbox-sw',
+			array( $this, 'get_workbox_script' ),
+			array()
+		);
+
 		/**
 		 * Fires when the WP_Service_Workers instance is initialized.
 		 *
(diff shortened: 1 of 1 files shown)


wp-includes/class-wp-service-workers.php
```diff
@@ -63,6 +70,43 @@ public function init() {
...
+	 * Get workbox script.
+	 *
+	 * @return string Script.
+	 */
+	public function get_workbox_script() {
+
+		// @todo The Workbox sources will probably need to be installed locally to avoid the external request(s).
+		$script = sprintf(
+			"importScripts( %s );\n",
+			wp_json_encode( 'https://storage.googleapis.com/workbox-cdn/releases/3.4.1/workbox-sw.js', 64 /* JSON_UNESCAPED_SLASHES */ )
```
jeffposnick: It's totally cool to prototype against the `storage.googleapis.com` hosted version of the code, but if/when you do get closer to deploying this more widely, we'd like to avoid a WordPress level of traffic for our CDN bucket 😄

westonruter: @jeffposnick Thanks. We'll be sure to `workbox copyLibraries` to copy the scripts into WordPress instead of using the CDN.

westonruter:
Closing in favor of #43.

PR #40, opened 2018-07-30
Removed php-notice from serviceworker

nico-martin:
In `wp-includes/class-wp-service-workers.php` there is a `_doing_it_wrong` function which creates a php notice.
In my opinion that is not necessary because there is also a `console.warn` right after that. But it violates the content type which should be JavaScript but the php notice then would be HTML/Markup. Can we remove it or would it be better to suppress the notice using `@`?


westonruter:
Good question. I suppose the benefit of the `_doing_it_wrong()` is that the PHP issue would be included in the PHP error log. That would allow for the error to be discovered more easily than monitoring the console for warnings. You're right that we'd want to suppress the output from being added to the service worker since it is HTML.

westonruter:
Alas, the error suppression also suppresses it from the error log.

westonruter:
But plugins that listen for `doing_it_wrong_run` actions will benefit.

PR #42, opened 2018-07-31
Bring back "Turbo" mode to Admin

### LINKED ISSUE #34: Use Service Workers (as grandchild of Gears) to bring back Turbo mode in admin
Prior to WordPress 3.0 there was a "Turbo mode" in the WordPress admin which was used to cache scripts and stylesheets in the admin to speed up the experience. This Turbo mode depended on Google Gears. When Gears was discontinued the Turbo mode was also discontinued, and core switched to improving performance by concatenating scripts via `load-scripts.php` and `load-styles.php`. This has not been without problems, however. If there are too many scripts included then the URL could become too long in some browsers (though these browsers surely don't support service workers). Also this concatenation breaks the ability to reuse the cache for individual scripts. Google Gears was discontinued in favor of AppCache, and then AppCache has been deprecated in favor of Service Workers. The service worker is now capable of picking up the mantle of its grandfather Google Gears to bring back Turbo mode, and in a standardized way that does not require any browser plugins.

For history of Turbo mode, see:

* [#6965](https://core.trac.wordpress.org/ticket/6965): Google Gears support (WordPress 2.6)
* [#11301](https://core.trac.wordpress.org/ticket/11301): Remove Gears (WordPress 3.0)
* [#20981](https://core.trac.wordpress.org/ticket/20981): Remove gears-manifest.php and manifest.php (WordPress 3.5)
* [WordPress.com post on Gears](https://en.blog.wordpress.com/2008/07/02/gears/)

For more, see [WordPress Trac tickets mentioning “Gears” or “Turbo”](https://www.google.com/search?q=site%3Acore.trac.wordpress.org%2Fticket%2F+%28gears+OR+turbo%29).

miina:
Fixes #34.

Using Workbox API Prototype for caching the admin assets separately and replacing the usage of `load-scripts.php` and `load-styles.php`.

Screenshots of the PR in use:

- Workbox log:
screen shot 2018-08-02 at 1 27 05 pm

- Network tab:
screen shot 2018-08-02 at 1 28 17 pm
- Output in the SW script:
screen shot 2018-08-02 at 1 29 38 pm


wp-includes/class-wp-service-workers.php
```diff
@@ -63,6 +129,93 @@ public function init() {
...
+		}
+
+		foreach ( $inc_files as $filename ) {
+			$ext = pathinfo( $filename, PATHINFO_EXTENSION );
+			if ( ! in_array( $ext, array( 'js', 'css' ), true ) ) {
+				continue;
+			}
+
+			// @todo This will cache both min.js and .js, however, not all the files have .min.js. Is it OK to cache all the files?
+			$routes[] = strstr( $filename, '/wp-includes' );
```
miina: Should we try to precache only the files which are visibly minified (ending with `.min.js`) or don't have a minified version separated?

wp-includes/class-wp-service-workers.php
```diff
@@ -51,10 +93,34 @@ class WP_Service_Workers extends WP_Scripts {
...
+		}
+
+		$this->register(
+			'workbox-sw',
+			array( $this, 'get_workbox_script' ),
+			array()
+		);
+
+		if ( SCRIPT_DEBUG || is_preview() ) {
+			$this->register_cached_route( '/(wp-admin|wp-includes)/.*\.(?:js|css)/', self::STRATEGY_NETWORK_ONLY );
```
westonruter: This is missing a `regex` flag?

wp-includes/class-wp-service-workers.php
```diff
@@ -51,10 +93,34 @@ class WP_Service_Workers extends WP_Scripts {
...
+			require_once ABSPATH . 'wp-admin/includes/file.php';
+		}
+
+		$this->register(
+			'workbox-sw',
+			array( $this, 'get_workbox_script' ),
+			array()
+		);
+
+		if ( SCRIPT_DEBUG || is_preview() ) {
```
westonruter: Why `is_preview()`? How will this ever return `true` given that it is the service worker response?

wp-includes/class-wp-service-workers.php
westonruter: If there are two versions of a given URL, cache the min version. Otherwise, if there is no min version then cache the regular.

For example, jQuery is bundled with core and it is pre-minified so there is no minified version. The `jquery-migrate` script, however, does have a minified version:

https://github.com/WordPress/wordpress-develop/blob/c52e7c79b2fd6381ba89fb8c40624e6621284743/src/wp-includes/script-loader.php#L211-L212

Basically look for the presence of the `$suffix` variable in the registered URLs.

wp-includes/class-wp-service-workers.php
westonruter: Is this pattern missing image extensions?

wp-includes/class-wp-service-workers.php
```diff
@@ -51,10 +93,34 @@ class WP_Service_Workers extends WP_Scripts {
...
+	 * @var array
+	 */
+	public $registered_caching_routes = array();
+
 	/**
 	 * Initialize the class.
 	 */
 	public function init() {
+
+		if ( ! function_exists( 'list_files' ) ) {
```
westonruter: The `list_files()` function is slow. What about instead just iterating over of the registered scripts and styles in `wp_scripts()` and `wp_styles()`? Use the `ver` with each registered asset as the revision.

wp-includes/class-wp-service-workers.php
```diff
@@ -63,6 +129,117 @@ public function init() {
...
+		$routes = array_merge(
+			$this->get_routes_from_file_list( $admin_files, 'wp-admin' ),
+			$this->get_routes_from_file_list( $inc_files, 'wp-includes' )
+		);
+
+		if ( empty( $routes ) ) {
+			return;
+		}
+
+		$this->register_cached_route( $routes, self::STRATEGY_PRECACHE );
```
westonruter: It's going to be slow to as it is right now because it is going to fetch the file contents for all of the URLs to compute the hashes of each. Instead, we should use the current WordPress version as the revision, and get the list of files by looking at what is registered among default scripts and styles.

wp-includes/class-wp-service-workers.php
miina: Right💡 Will remove.

miina: Hmm, actually, at this moment `init` is called on any page if any route caching is registered, not only when serving the request of the registered service worker.

wp-includes/class-wp-service-workers.php
miina: Changed the logic to use `wp_scripts()` and `wp_styles()`, left in the `list_files()` for images for now within 024eab3.

wp-includes/class-wp-service-workers.php
```diff
@@ -51,10 +86,47 @@ class WP_Service_Workers extends WP_Scripts {
...
+		);
+
+		$this->register(
+			'caching-utils-sw',
+			PWA_PLUGIN_URL . '/wp-includes/js/service-worker.js',
+			array( 'workbox-sw' )
+		);
+
+		if ( SCRIPT_DEBUG ) {
+			$this->register_cached_route( '/(wp-admin|wp-includes)/.*\.(?:js|css|gif|png|svg)/', self::STRATEGY_NETWORK_ONLY, array(), true );
```
westonruter: Come to think of it, should not this condition be removed entirely? If using `networkOnly` then this is just passing through to the network and it would be better to just skip caching the route entirely. In other words, this `if` could become:

```php
if ( ! SCRIPT_DEBUG ) {
    $this->precache_admin_assets();
}
```

wp-includes/class-wp-service-workers.php
westonruter: Yes. It's what WordPress uses as the default `ver` as well: 

https://github.com/WordPress/wordpress-develop/blob/c52e7c79b2fd6381ba89fb8c40624e6621284743/src/wp-includes/script-loader.php#L64

wp-includes/class-wp-service-workers.php
```diff
@@ -63,6 +135,119 @@ public function init() {
...
+	protected function precache_admin_assets() {
+
+		$admin_dir    = ABSPATH . 'wp-admin/';
+		$admin_images = list_files( $admin_dir . 'images/' );
+		$inc_images   = list_files( ABSPATH . WPINC . '/images/' );
+
+		$routes = array_merge(
+			$this->get_routes_from_file_list( $admin_images, 'wp-admin' ),
+			$this->get_routes_from_file_list( $inc_images, 'wp-includes' ),
+			$this->get_admin_routes_from_dependency_list( wp_scripts()->registered ),
```
westonruter: We should only include assets that are in the `wp-admin` and `wp-includes` directories. We should not include plugin-registered assets by default, I don't think. Or maybe we should. My concern is a plugin updated which doesn't define the `ver`, and so when the plugin updates and a new version of the script is included, then the service worker will not update the cache.

westonruter: Sorry, I see you're doing this already in `get_admin_routes_from_dependency_list`!!

wp-includes/class-wp-service-workers.php
```diff
@@ -63,6 +135,119 @@ public function init() {
...
+	 */
+	protected function get_admin_routes_from_dependency_list( $dependencies ) {
+		$routes = array();
+		foreach ( $dependencies as $handle => $params ) {
+
+			// Only precache scripts from wp-admin and wp-includes.
+			if ( false === strpos( $params->src, 'wp-admin' ) && false === strpos( $params->src, 'wp-includes' ) ) {
+				continue;
+			}
+			$revision = false === $params->ver ? get_bloginfo( 'version' ) : $params->ver;
```
westonruter: We could potentially opt-in to using the service worker for theme/plugin assets if we detect that the script is located within `wp-content/themes` or `wp-content/plugins` and then grab the version for the theme or plugin to use if the `ver` is empty. Or even, the `ver` could be combined with the version of the plugin to use as the revision.

wp-includes/class-wp-service-workers.php
```diff
@@ -63,6 +135,119 @@ public function init() {
...
+	 *
+	 * @param array  $list List of file paths.
+	 * @param string $folder Folder -- either 'wp-admin' or 'wp-includes'.
+	 * @return array List of routes.
+	 */
+	protected function get_routes_from_file_list( $list, $folder ) {
+		$routes = array();
+		foreach ( $list as $filename ) {
+			$ext = pathinfo( $filename, PATHINFO_EXTENSION );
+			if ( ! in_array( $ext, array( 'png', 'gif', 'svg' ), true ) ) {
```
westonruter: I was going to ask about JPEG but then I saw that there are no JPEGS in core. For reference, the static assets are (via `git ls-files wp-includes/ wp-admin/ | sed 's/.*\.//' | sort | uniq -c | sort -nr`):

```
  99 png
  62 gif
  51 css
  10 scss
   7 txt
   6 svg
   3 woff
   3 ttf
   3 eot
   1 xml
   1 crt
```

It doesn't make sense to include `crt`, `eot` , `xml`, or `scss`. But `woff` would be very useful. Also, if this is used for themes then it would be good to include the JPEG extension(s).

miina: Thinking that maybe instead of using `list_files` for getting the 3 `woff` files we could list these out manually?

wp-includes/class-wp-service-workers.php
miina: Maybe precaching theme/plugin assets could be something that could be opted in from admin UI as a setting instead of doing this by default? Thinking that precaching theme and plugin files could cause issues (as you mentioned above) if the version is not changed when updating the code (either in WP repo or directly in the environment). Feels like it's better to leave the asset precaching by default for the admin interface only, at least for now. Thoughts?

wp-includes/class-wp-service-workers.php
miina: Added the three `.woff` files as a static list within 7132b74.

wp-includes/integrations/class-wp-service-worker-styles-integration.php
```diff
@@ -68,11 +76,15 @@ public function register( WP_Service_Worker_Cache_Registry $cache_registry ) {
 
 			$revision = false === $dependency->ver ? get_bloginfo( 'version' ) : $dependency->ver;
 
-			/** This filter is documented in wp-includes/class.wp-styles.php */
-			$url = apply_filters( 'style_loader_src', $url, $handle );
+			// @todo Look into why this fails with 'colors', note that $_wp_admin_css_colors is null at this moment.
+			if ( 'colors' !== $handle ) {
```
miina: For some reason `colors` style fails both in PHP and also in Workbox. In PHP it looks like it might be due to `$_wp_admin_css_colors` not being set when calling `style_loader_src` filter.

In Workbox the output in console is the following:
screen shot 2018-09-11 at 5 00 48 pm

wp-includes/service-workers.php
```diff
@@ -236,4 +236,14 @@ function wp_default_service_workers( $service_workers ) {
 				);
 		}
 	}
+
+	add_action( 'wp_admin_service_worker', array( $service_workers, 'precache_admin_assets' ), 9 );
```
miina: Not sure if that's the best place where to add precaching for admin assets. Thoughts, @westonruter?

wp-includes/class-wp-service-workers.php
```diff
@@ -286,6 +286,116 @@ protected function get_base_script() {
...
+	 * Add precaching for wp-admin and wp-includes .js and .css files.
+	 */
+	public function precache_admin_assets() {
+
+		// Don't continue if debug mode.
+		if ( SCRIPT_DEBUG ) {
+			return;
+		}
+
+		do_action( 'admin_init' );
```
miina: Looks like doing `admin_init` here fixes the issues with the `colors` script and the `$_wp_admin_css_colors` not being set, not sure if that's the best solution.

westonruter: This is an issue that I had when wanting to `do_action('admin_enqueue_scripts')` for the admin as it is doing `wp_enqueue_scripts()` for the frontend. This fails, however, when the user is not logged-in and thus the service worker would not be able to be installed at `wp-login.php`, for example.

wp-includes/integrations/class-wp-service-worker-styles-integration.php
```diff
@@ -12,6 +12,14 @@
  */
 class WP_Service_Worker_Styles_Integration extends WP_Service_Worker_Base_Integration {
 
+	/**
+	 * Scope this integration applies to.
+	 *
+	 * @since 0.2
+	 * @var int
+	 */
+	protected $scope = WP_Service_Workers::SCOPE_ALL;
```
westonruter: @felixarntz Note how this had to be added to the subclass to override the variable in the base class. Something doesn't seem quite right.

wp-includes/service-workers.php
westonruter: @felixarntz With the change to `WP_Service_Worker_Scripts_Integration` and `WP_Service_Worker_Styles_Integration` to add them to both the front and admin scopes, then the integration will run at `wp_admin_service_worker` priority 10:

https://github.com/xwp/pwa-wp/blob/57ebfa7652cd6e485056c7d11417acb726a7e87d/wp-includes/service-workers.php#L219-L221

This means that in order to set the `precache` flags on admin-specific scripts and styles that the flag has to be set at priority 9 of the`wp_admin_service_worker` action. Something is missing here.

wp-includes/default-filters.php
```diff
@@ -18,3 +18,8 @@
 add_action( 'wp_head', 'wp_add_error_template_no_robots' );
 add_action( 'error_head', 'wp_add_error_template_no_robots' );
 add_action( 'wp_default_service_workers', 'wp_default_service_workers' );
+
+// Disable contactenation of scripts and styles on admin pages.
+foreach ( array( 'wp_default_styles', 'wp_print_scripts' ) as $filter ) {
```
westonruter: I think this would be better done once at `admin_init` instead.

wp-includes/service-workers.php
```diff
@@ -236,4 +236,14 @@ function wp_default_service_workers( $service_workers ) {
...
+}
+
+/**
+ * Disables concatenating scripts to leverage caching the assets via Service Worker instead.
+ */
+function wp_disable_script_concatenation() {
```
westonruter: Something we need to look into is what to do for users who don't have service workers available. In their case, we actually need to retain the concatenation functionality.

The only option that comes to mind is that when the promise [returned](https://github.com/xwp/pwa-wp/blob/57ebfa7652cd6e485056c7d11417acb726a7e87d/wp-includes/service-workers.php#L110-L117) by `navigator.serviceWorker.register()` resolves:

```diff
diff --git a/wp-includes/service-workers.php b/wp-includes/service-workers.php
index 69a39fb..4d2029c 100644
--- a/wp-includes/service-workers.php
+++ b/wp-includes/service-workers.php
@@ -112,7 +112,9 @@ function wp_print_service_workers() {
 					navigator.serviceWorker.register(
 						<?php echo wp_json_encode( wp_get_service_worker_url( $name ) ); ?>,
 						<?php echo wp_json_encode( compact( 'scope' ) ); ?>
-					);
+					).then( function() {
+						document.cookie = 'wordpress_sw_installed=1; path=/; expires=Fri, 31 Dec 9999 23:59:59 GMT; secure; samesite=strict';
+					} );
 				<?php } ?>
 			} );
 		}
```

That we can set a cookie like `wordpress_sw_installed` which can then be used to check for whether to proceed with concatenation. 

So then in this `wp_disable_script_concatenation` it can first check if `isset( $_COOKIE['wordpress_sw_installed'] )` before proceeding.

Note that concatenation only applies in the admin, which entails that the user is authenticated, and no full-page caching is used. Thus there won't be concerns about reading the cookie since these pages won't be cached.

wp-includes/integrations/class-wp-service-worker-styles-integration.php
felixarntz: @westonruter I think it makes sense to have integrations define their scope. However, the implementation I made may not be obvious enough. Currently, the default scope is `SCOPE_FRONT`, via the property. So in order to change it to `SCOPE_ADMIN` or `SCOPE_ALL` you need to reset the property in the child class. Two ideas:
* It might make more sense to set the default scope to `SCOPE_ALL`.
* Or alternatively, maybe it's better to not have any default scope, but instead make the `WP_Service_Worker_Base_Integration::get_scope()` method abstract, thus require each integration to explicitly define it. That would also make it obvious that you as the interface developer _have_ to handle it.

wp-includes/class-wp-service-workers.php
miina: It seems to work when testing, the only thing actually needed is for `$_wp_admin_css_colors` to be populated properly so we could just call `register_admin_color_schemes()` here instead, will replace it.

wp-includes/service-workers.php
miina: @westonruter Thanks for bringing that issue out and also for the solution suggestion, good idea!

Added within ec569c4.

wp-includes/integrations/class-wp-service-worker-scripts-integration.php
```diff
@@ -12,6 +12,14 @@
  */
 class WP_Service_Worker_Scripts_Integration extends WP_Service_Worker_Base_Integration {
 
+	/**
+	 * Scope this integration applies to.
+	 *
+	 * @since 0.2
+	 * @var int
+	 */
+	protected $scope = WP_Service_Workers::SCOPE_ALL;
```
felixarntz: This can be removed, as it is no longer necessary after #72. The scripts and styles integrations are now already available in both scopes.

wp-includes/integrations/class-wp-service-worker-styles-integration.php
felixarntz: This can be removed, as it is no longer necessary after #72.

wp-includes/class-wp-service-workers.php
```diff
@@ -78,31 +78,19 @@ public function get_registry() {
 	/**
 	 * Get the current scope for the service worker request.
 	 *
-	 * @global WP $wp
-	 *
-	 * @return int Scope. Either SCOPE_FRONT, SCOPE_ADMIN, or if neither then 0.
+	 * @todo We don't really need this. A simple call to is_admin() is all that is required.
```
miina: @westonruter Not related to this specifically but a little bit related in terms of checking for `self::QUERY_VAR`:
Should we only `init` `WP_Service_Worker_Scripts` if the `self::QUERY_VAR` does exist? Otherwise wouldn't the scripts always be registered but only served if `self::QUERY_VAR` is set? Wouldn't the scripts need to be registered only in case of serving the service worker? Thoughts?

PR #43, opened 2018-07-31
Workbox API Prototype

### LINKED ISSUE #5: Propose a Workbox solution for WordPress

miina:
Fixes #5. Start of the prototype for PHP layer on top of Workbox.

Adds ~three~ two methods for registering caching strategies:
- `wp_register_route_caching_strategy()`
- `wp_register_routes_precaching()`.
- ~`wp_register_caching_strategy_by_route_pattern()`;~
- ~`wp_register_caching_strategy_by_route_url()`;~

The caching strategies are added to the end of the Service Worker file output.

You could register a caching strategy like this for example:

- Precaching:
```php
$routes = array(
    array(
        'url' => '/wp-admin',
        'revision' => `my-revision`,
    ),
);
wp_register_routes_precaching( $routes );
```
    
- Caching strategy:
```php
wp_register_route_caching_strategy(
	'/wp-admin/.*\.css',
	WP_Service_Workers::STRATEGY_STALE_WHILE_REVALIDATE,
	array(
		'cache_name' => 'Assets',
		'plugins'    => array(
			'expiration' => array(
				'maxEntries'    => 3,
				'maxAgeSeconds' => 1000,
			),
		),
	)
);
```


wp-includes/js/service-worker.js
```diff
@@ -0,0 +1,52 @@
+/* global workbox */
+
+/**
+ * Handle registering caching strategies.
+ */
+
+// @todo Perhaps some other variable makes more sense?
```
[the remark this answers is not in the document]
westonruter: In a service worker it is `self` that is the global.

wp-includes/class-wp-service-workers.php
```diff
@@ -37,6 +37,14 @@ class WP_Service_Workers extends WP_Scripts {
 	 */
 	const SCOPE_ALL = 3;
 
+	/**
+	 * Stale while revalidate caching strategy.
+	 *
+	 * @todo Add more strategies.
+	 * @var int
+	 */
+	const STRATEGY_STALE_WHILE_REVALIDATE = 1;
```
westonruter: Instead of `1` I think better to just use the workbox identifiers, for example `'staleWhileRevalidate'`. That will make it easier.

wp-includes/class-wp-service-workers.php
```diff
@@ -83,6 +148,55 @@ public function register( $handle, $src, $deps = array(), $scope = self::SCOPE_A
...
+	/**
+	 * Register caching strategy for route.
+	 *
+	 * @param string $route Route.
+	 * @param int    $strategy Strategy, can be self::STRATEGY_STALE_WHILE_REVALIDATE (@todo Add others).
+	 * @param array  $args Array of args, can be cache_name, max_age, max_entries.
+	 * @return string Script.
+	 */
+	protected function register_caching_strategy_for_route( $route, $strategy, $args ) {
+		$script = "wp.serviceWorker.addCachingStrategy( '" . $route . "', '" . $strategy . "'";
```
westonruter: Note that `wp_json_encode()` needs to be used here instead of manually including quotes. This will ensure that if a route contains quotes it won't break the JS, for example.

wp-includes/js/service-worker.js
```diff
@@ -0,0 +1,52 @@
...
+	/**
+	 * Add caching strategy. Missing processing conflicting routes.
+	 *
+	 * @param {string}      route Route.
+	 * @param {string}      strategy Strategy.
+	 * @param {string|null} cacheName Cache name.
+	 * @param {int}         maxAge Max Age.
+	 * @param {int}         maxEntries Max entries.
+	 */
+	addCachingStrategy: function ( route, strategy, cacheName, maxAge, maxEntries ) {
```
westonruter: I wonder if this abstraction would be necessary or if we'd be largely reinventing the wheel that Workbox already provides. In other words, what if `wp.serviceWorker` were just an alias to `workbox`? So then instead of calling `workbox.routing.registerRoute()` we could instead call `wp.serviceWorker.routing.registerRoute()`?

wp-includes/class-wp-service-workers.php
```diff
@@ -83,6 +148,55 @@ public function register( $handle, $src, $deps = array(), $scope = self::SCOPE_A
...
+		// @todo Add other strategies.
+		// @todo Add validation.
+		// @todo Logic for detecting conflicts.
+		switch ( $strategy ) {
+			default:
+				$strategy = 'staleWhileRevalidate';
+				break;
+		}
+
+		$this->caching_strategies .= $this->register_caching_strategy_for_route( $route, $strategy, $args );
```
westonruter: Instead of concatenating JS to then output later, it would be better to have an array in the class like `$registered_caching_routes` which can then be iterated over at the end to generate the JS. This would allow us to do better conflict detection.

wp-includes/service-workers.php
```diff
@@ -37,6 +37,17 @@ function wp_register_service_worker( $handle, $src, $deps = array(), $scope = WP
...
 }
 
+/**
+ * Register route and caching strategy.
+ *
+ * @param string $route Route.
+ * @param int    $strategy Strategy, can be WP_Service_Workers::STRATEGY_STALE_WHILE_REVALIDATE (@todo Add others).
+ * @param array  $args Array of args, can be cache_name, max_age, max_entries.
+ */
+function wp_register_caching_strategy( $route, $strategy = WP_Service_Workers::STRATEGY_STALE_WHILE_REVALIDATE, $args = array() ) {
```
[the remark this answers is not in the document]
miina: Makes sense, I'll change it to `wp_register_route_caching_strategy` for now and then this can be changed / removed later if needed.

wp-includes/js/service-worker.js
miina: Could be. Added this abstraction mainly thinking of having all the direct usage of Workbox in one JS file and to only use the abstraction in PHP files in case for some reason core would move away from Workbox. Then just this JS file would need to be modified to replace the logic while keeping the abstraction method the same. However, I might be overthinking, and it's true that using abstraction like this might be redundant. In this case we could for example remove the `service-worker.js` file and define `wp.serviceWorker` within `WP_Service_Workers::get_workbox_script()`. And then use `wp.serviceWorker.routing.registerRoute()` within `WP_Service_Workers::register_caching_strategy_for_route()` instead. Thoughts?

miina: Removing the abstraction layer to use `wp.serviceWorker` as an alias to `workbox` for now.

jeffposnick: Just a quick heads-up in case anything catches you out: the `workbox.*` namespace is, for the most part, backed by a [`Proxy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) that will [dynamically load](https://github.com/GoogleChrome/workbox/blob/096e44897e9c348f85bb51927c1113de4e339897/packages/workbox-sw/controllers/WorkboxSW.mjs#L59-L72) (via `importScripts()`) the underlying `.js` on demand.

One wrinkle is that calls to `importScripts()` need to happen as part of the initial, synchronous execution of a service worker script—as opposed to, e.g., inside of a callback registered with `addEventListener()`.

That's all to say: there's a chance that you might run into unexpected errors that are hard to diagnose if you end up making the initial call to, e.g., a `wp.serviceWorker.strategies.*` method inside of, e.g., a custom `fetch` event listener. I wanted to explicitly call this out to head off any really confusing debugging 😄 

(There's more context, including some workarounds for synchronously preloading the required scripts, at https://github.com/GoogleChrome/workbox/issues/1504)

westonruter: @jeffposnick thank you for that helpful info.

In the PR now there is a `wp.serviceWorker = workbox` statement to achieve an alias. However, the `workbox` name still remains and it is important to remain currently because the imported scripts all extend that global object from what I understand. 

To recap, we discussed creating a wrapper API around Workbox so that in the future we could potentially swap out Workbox with another implementation later, by however unlikely chance we’d need to. This was done in WordPress with the introduction of CodeMirror by creating a `wp.codeEditor` abstraction. Now with the Gutenberg there is the `wp.element` namespace which exports the React functionality; if WordPress moves away from React in the future, then `wp.element.createElement` could be easily swapped out with another implementation that has the same behaviors. So we’d like to do something similar with Workbox in the service worker, perhaps making Workbox available as `wp.serviceWorker` as seen here now.

All of this to say, it doesn’t seem that the `workbox copyLibraries` command makes this possible to use a different export namespace than the `workbox.*` global object. Is that something which would need to be added, or is it possible to do now using an alternative means?

wp-includes/class-wp-service-workers.php
```diff
@@ -83,6 +187,149 @@ public function register( $handle, $src, $deps = array(), $scope = self::SCOPE_A
...
+];';
+		}
+
+		$args_script = $script;
+
+		$script .= '
+wp.serviceWorker.routing.registerRoute(';
+
+		// Check if the route is regex or string literal.
+		$is_regex_string = @preg_match( $route, '' ); // phpcs:ignore Generic.PHP.NoSilencedErrors.Discouraged
```
miina: Is there a better way how to check if it's a `string` or regex?

wp-includes/js/service-worker.js
jeffposnick: We don't allow for customizing the top-level namespace when using `workbox copyLibraries`, but `wp.serviceWorker = workbox` seems like a reasonable approach to indirection.

We do ship `.mjs` files (i.e. JavaScript modules) with explicit exports of each of the underlying classes/static methods, without a namespace. It's possible to create a source file that references only the exports you need, pulling them all into a distinct namespace that you control, and then running that through `webpack` or `Rollup` to create the final bundle. That's really the alternative that we offer for folks who need more than our default `workbox.*`-based dynamic loading experience.

If you're happy with the `wp.serviceWorker = workbox` approach, especially when prototyping, I don't think there's any need to switch.

wp-includes/class-wp-service-workers.php
```diff
@@ -123,6 +371,27 @@ public function serve_request( $scope ) {
 		echo $this->output; // phpcs:ignore WordPress.XSS.EscapeOutput, WordPress.Security.EscapeOutput
 	}
 
+	/**
+	 * Add logic for routes caching to the request output.
+	 *
+	 * @todo Handle conflicts between routes.
```
miina: @westonruter Any thoughts on what would be the goal with detecting conflicts? Should we just report it in the console, for example as Ryan suggested here: https://github.com/xwp/pwa-wp/issues/5#issuecomment-401248790? Or should we report it upon the plugin activation that causes conflicts? Or should we try to handle these in some way, for example if conflicting routes have the same strategy then that's actually not a conflict, however, if there is a conflict, should we dismiss both routes? Or should we leave detecting conflicts out of the prototype for now?

Thoughts?

wp-includes/js/service-worker.js
westonruter: The only concern with the `wp.serviceWorker = workbox` approach is that third party code could start to directly access `workbox`, which we'd want to avoid.

Also, since we can't anticipate the functionality that third party code will leverage we need to have the dynamic imports, instead of creating a build with Webpack.

What we have is fine for now, but once we get to the point of proposing for WordPress core merge then this issue will come up again.

wp-includes/class-wp-service-workers.php
westonruter: @miina I don't think we can do it in PHP, since we won't be able to programmatically know if two patterns will match without the URL they are matching against. We need to do it at runtime instead via JS. In particular, we can create a `CustomRouter` that overrides `_findHandlerAndParams()` as Jeff suggests in https://github.com/xwp/pwa-wp/issues/5#issuecomment-403082892

wp-includes/class-wp-service-workers.php
westonruter: I think we need to explicitly indicate whether a given route is a regex. For example, in `$args` there could be `'regex' => true`.

Alternatively, we could automatically treat a `$route` that begins and ends with `/` as a regex, and then `trim('$route, /')` when passing into the `RegExp` constructor. But this isn't good either because, of course, URLs often begin and end with `/` 😄 

So I think explicitly indicating the route type via a flag is the way to go.

wp-includes/class-wp-service-workers.php
miina: @westonruter Saw Jeff's comment and was looking into doing it via JS with the `CustomRouter`, I should have asked the question in general comments thread for clarity, sorry for that. Saw how Workbox works by default and takes the first matching route. Considering creating a custom router that overrides `_findHandlerAndParams()` wondering what would be the goal with detecting conflicts and if we should include this into the prototype.
For example should we:
- just leave the default behavior, report the conflicts;
- Dismiss any conflicting logic, report the conflicts;

In case of including it to the prototype, perhaps we could start with leaving the default behavior and reporting conflicts? Thoughts?

wp-includes/class-wp-service-workers.php
miina: Added within c16b6b5, defaulting to `false`.

wp-includes/class-wp-service-workers.php
miina: Note: added custom router initial implementation within fc751ad.

wp-includes/js/service-worker.js
```diff
@@ -0,0 +1,148 @@
...
+				matchResult = undefined;
+			}
+
+			return {
+				matchedRoute,
+				params: matchResult,
+				handler: matchedRoute.handler
+			};
+
+			// @todo Confirm approach for conflicting routes: for example perhaps we shouldn't consider multiple matches but same strategy a conflict. Maybe we should still use the first matching route and warn about others.
```
miina: Currently this approach is skipping all the routes where more than one match was found and logging the warning like this:
screen shot 2018-08-06 at 2 41 11 pm

Thinking that instead of this approach maybe we should take the first match as it is in the default behavior and report the conflicts. However, perhaps if the strategy is the same (e.g. two matching registered routes with the same strategy) then this is actually not a conflict and not report this. So basically report all the matching routes where the strategy differs. Just realized that there could still be conflicts when the expiration time configuration differs though. Thoughts, @westonruter?

westonruter: Just realized I commented on this todo in another PR: https://github.com/xwp/pwa-wp/pull/48#discussion_r208068200

>  I think it's a good idea to let the first match handle the request, and then to warn about the conflicting handlers.

Your right that we need to report any time that more than one match occurs, since different caches could be used or different plugins employed.

miina:
@westonruter / @hellofromtonya:
Is there anything else you see that could be necessary / useful to add to the Workbox API Prototype? If this seems to be covering the basic use cases for now then I'd do a more thorough testing and then remove the `WIP` status. And then we could tweak the API accordingly within the feature PR-s.

wp-includes/service-workers.php
```diff
@@ -37,6 +37,64 @@ function wp_register_service_worker( $handle, $src, $deps = array(), $scope = WP
...
+ * @param array  $strategy_args {
+ *     An array of strategy arguments.
+ *
+ *     @type string $cache_name Cache name.
+ *     @type array  $plugins    Array of plugins with configuration. The key of each plugin in the array must match the plugin's name.
+ *                              See https://developers.google.com/web/tools/workbox/guides/using-plugins#workbox_plugins.
+ * }
+ */
+function wp_register_caching_strategy_by_route_url( $route, $strategy = WP_Service_Workers::STRATEGY_STALE_WHILE_REVALIDATE, $strategy_args = array() ) {
+	wp_service_workers()->register_cached_route_url( $route, $strategy, $strategy_args );
```
westonruter: I just recalled that in WordPress the rewrite API and the REST API both only support regular expressions. That being the case, we could just opt to not expose registering non-regex routes in Workbox. As [the docs say](https://developers.google.com/web/tools/workbox/guides/route-requests#matching_a_route_with_a_string):

> Matching a route with a string is the easiest to understand, but also the least flexible option.

So if we just support regular expressions, then this would simplify the PHP API while also being more aligned with functions like `register_rest_route()` and `add_rewrite_rule()`.

wp-includes/class-wp-service-workers.php
```diff
@@ -83,6 +187,185 @@ public function register( $handle, $src, $deps = array(), $scope = self::SCOPE_A
...
+			if ( ! isset( $route['revision'] ) ) {
+				$route['revision'] = get_bloginfo( 'version' );
+			}
+
+			$routes_list[] = $route;
+		}
+		if ( empty( $routes_list ) ) {
+			return '';
+		}
+		return sprintf( "wp.serviceWorker.precaching.precacheAndRoute( %s );\n", wp_json_encode( $routes_list ) );
```
westonruter: Should the `precacheAndRoute` call here include the `directoryIndex: null` option? In a WordPress context, it doesn't make sense that Workbox would look for `index.html`.

https://developers.google.com/web/tools/workbox/modules/workbox-precaching#directory_index

The same goes for the `cleanUrls` option. I think it should be disabled too.

wp-includes/class-wp-service-workers.php
```diff
@@ -491,59 +482,64 @@ protected function remove_url_scheme( $schemed_url ) {
 	 * Get validated path to file.
 	 *
 	 * @param string $url Relative path.
-	 * @param bool   $allow_content_only If to allow path from content directory only. Defaults to true.
 	 * @return null|string|WP_Error
 	 */
-	protected function get_validated_file_path( $url, $allow_content_only = true ) {
```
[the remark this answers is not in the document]
jeffposnick: Sorry, just chiming in here in case folks weren't aware of this context: the location of the service worker file that gets registered is [pretty important](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#Why_is_my_service_worker_failing_to_register). It needs to be located at either at the top-level of the eventual service worker scope, or if that's not possible, then you need to configure your web server to include the `Service-Worker-Allowed` response header specifying what the scope should be.

In practice, getting folks to configure their web servers "in the wild" is a bit of a pain, so serving the service worker file from a top-level URL ends up being the easiest thing to do. So if you're going to want the service worker to control all pages on a given origin, hosting it at `/service-worker.js` (rather than `/wp-content/service-worker.js`) is important.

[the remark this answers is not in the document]
miina: The service worker itself will always be served from the top-level URL, this specific method is about where the scripts that are registered via the API are allowed from. The API combines together different scripts which will always be served from the top-level URL, however, the combined script files themselves can be in a different place. This note was about the location of the scripts which will be combined into the service worker.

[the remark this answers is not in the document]
jeffposnick: Awesome! Sorry for the noise 😄

PR #48, opened 2018-08-06
Add ability to serve the default offline page when the user or site is offline.

### LINKED ISSUE #45: Add ability to serve the default offline page when the user or site is offline
As an administrator site owner, I expect the selected default offline page to be served when either the user or site is offline.

- [ ] AC1 Serve the default offline page when the requested page is not available offline.
- [ ] AC2 Add `is_offline` to the query to denote. [See Weston's comment](https://github.com/xwp/pwa-wp/pull/28#discussion_r205304263). 

Issue #23 added the ability to select the default offline page.  Issue 23 includes conversations on how to serve the default offline page:

See [Weston's comment](https://github.com/xwp/pwa-wp/issues/23#issuecomment-404276424):
>Something important to note here I just realized: this offline page could be served not only when someone's internet is offline but also when the server itself experiences an internal server error (a 500 response). We should try to design the offline page so that it could be used for both use cases. There would need to be some client-side logic involved here, however, since the client would know if the page fails to be fetched because of a network error or a server error.

Important related issues in [Issue #9](https://github.com/xwp/pwa-wp/issues/9) including [this comment](https://github.com/xwp/pwa-wp/issues/9#issuecomment-403262531):
>We could then serve it only if requesting with the query var as you suggested. If no offline page exists, then the response could be treated as is_404() to serve the not found message, which I think would be better then no offline page at all. (Alternatively, I wonder also if the template hierarchy should be extended so that is_offline() would be a special case under is_404().)
>
>There are thousands of themes out there and it would take a long time for them to add an offline.php template. I don't think that should be a requirement before serving the offline page.
>
>Requiring a special offline template would also not align with the template hierarchy expectation either. A minimal theme is only required to include an index.php template which gets used then to serve responses to all queries, including is_404(). A theme should also be able to use this generic index template to serve an offline response as well (is_offline()?). That would seen the most logical from a theme developer's perspective.

And Weston [commented on PR 28](https://github.com/xwp/pwa-wp/pull/28):
>The serving of the offline page should probably incorporate Workbox (#5) and also it will probably involve precaching (#25) since we'll need to make sure the offline page and its required assets get added to the cache.

miina:
Fixes #45 .

* Precache and route the offline page URL, along with its featured image, in the frontend service worker.
* Serve offline page when unable to navigate to a URL due to do a server error or connection error.
* Precache and route the theme's header image, background image, and custom logo (with URLs in `srcset`).
* Precache and route all enqueued scripts and styles.
* Add runtime caching with stale-while-revalidate strategy for any images and fonts inside of themes (such as referenced in the stylesheets).
* Override `wp.serviceWorker.routing` with instance of `WPRouting` and frozen public API on the same way Workbox does.
* Add `is_offline()` template conditional.


wp-includes/class-wp-service-workers.php
```diff
@@ -63,6 +134,72 @@ public function init() {
 		do_action_ref_array( 'wp_default_service_workers', array( &$this ) );
 	}
 
+	/**
+	 * Configure offline page based on the saved offline page value.
+	 *
+	 * @return string Script.
+	 */
+	protected function configure_offline_page() {
```
westonruter: This should ideally tie-in to a precaching API (e.g. `register_precaching_for_routes`) instead of having its own separate logic.

wp-includes/class-wp-service-workers.php
```diff
@@ -83,6 +220,170 @@ public function register( $handle, $src, $deps = array(), $scope = self::SCOPE_A
...
+			self::STRATEGY_CACHE_ONLY,
+			self::STRATEGY_NETWORK_FIRST,
+			self::STRATEGY_NETWORK_ONLY,
+			self::STRATEGY_PRECACHE,
+		), true ) ) {
+			_doing_it_wrong( __METHOD__, esc_html__( 'Strategy must be either WP_Service_Workers::STRATEGY_STALE_WHILE_REVALIDATE, WP_Service_Workers::STRATEGY_CACHE_FIRST, WP_Service_Workers::STRATEGY_NETWORK_FIRST, WP_Service_Workers::STRATEGY_CACHE_ONLY, or WP_Service_Workers::STRATEGY_NETWORK_ONLY.', 'pwa' ), '0.2' );
+			return;
+		}
+
+		if ( self::STRATEGY_PRECACHE === $strategy ) {
```
westonruter: Instead of including `precache` in this method, I think it should probably be a separate method like `register_precached_route`. This then can be a straight wrapper for `workbox.precaching.precache()`: https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox.precaching#.precache

wp-includes/class-wp-service-workers.php
```diff
@@ -83,6 +220,170 @@ public function register( $handle, $src, $deps = array(), $scope = self::SCOPE_A
...
+
+	/**
+	 * Register precaching for routes. Gets the hashes of files' contents and adds as revision for each route.
+	 *
+	 * @param array $routes Array of routes.
+	 * @return string Precaching logic.
+	 */
+	protected function register_precaching_for_routes( $routes ) {
+
+		$routes_list = '[';
```
westonruter: It would be better to construct an array here, like `$routes_list = array()`. And then for each `$route` do:

```php
$routes_list[] = array(
    'url' => $base_url . $route,
    'revision' => $revision,
);
```

And then at the end just do:

```php
return sprintf( "wp.serviceWorker.precaching.precacheAndRoute( %s );\n", wp_json_encode( $routes_list ) );
```

wp-includes/class-wp-service-workers.php
```diff
@@ -83,6 +220,170 @@ public function register( $handle, $src, $deps = array(), $scope = self::SCOPE_A
...
+
+		if ( isset( $args['cache_name'] ) ) {
+			$script .= 'const args = {
+	cacheName: ' . wp_json_encode( $args['cache_name'] ) . '
+};';
+		}
+		if ( isset( $args['max_entries'] ) || isset( $args['max_age'] ) ) {
+			$script .= 'args.plugins = [
+	new wp.serviceWorker.expiration.Plugin({';
+			if ( isset( $args['max_age'] ) ) {
```
westonruter: The the `max_age` shouldn't be hard-coded here in `$args` I don't think. Instead, `$args` should have a `plugins` array like:

```php
array(
    'plugins' => array(
        'expiration' => array(
            'maxAgeSeconds' => 123,
            'maxEntries' => 45,
        ),
    )
)
```

This would then allow _any_ plugin to be referenced and passing all of the args that the plugins accept. The `$args` would be whatever the caching strategy `options` are accepted, for example: https://developers.google.com/web/tools/workbox/reference-docs/latest/workbox.strategies.CacheFirst#constructor_1

So `$args` should be renamed `$strategy_args` for clarity.

westonruter: That being the case, I think the `$regex` should not be passed in the `$args` now. Perhaps there should be a separate methods like `register_cached_route_pattern()` and `register_cached_route_url()` to differentiate between the two types of args allowed?

miina: Made the general Workbox related changes within 6c2592f in #43.

Makes sense that the plugins shouldn't be hardcoded and it should be possible to add other plugins, too. Made these changes within 6c2592f.

On regex: what do you think of just adding an additional param to the method so that if a developer would switch from between regex/url then only a param would require changing instead of the method? For example: `register_cached_route( $route, $strategy, $strategy_args = array(), $is_regex = false )`.
Thinking that since the only difference it makes when registering routes to `self::$registered_caching_routes` would be a change if `is_regex` is `true/false` for the relevant route, then maybe an additional param would be enough.

Or are you thinking that maybe it's more straightforward and better structured to have two separate methods? Thoughts?

(Added `$is_regex` param as a temporary change within 6c2592f to move this out of the `$strategy_args`, will it change it accordingly.)

wp-includes/class-wp-service-workers.php
miina: Updated the logic within dec04fb to use `register_precached_routes` for the precaching part.

Wondering if we should rename `register_precaching_for_routes` to `get_precaching_for_routes_script` or something similar, otherwise maybe it's too similar with `register_precached_routes`. Also, one is registering the route for caching and the other is generating and returning the script. Thoughts?

westonruter: Yes, good idea.

wp-includes/class-wp-service-workers.php
miina: Makes sense. Split the method into two within 0fd9320.

wp-includes/class-wp-service-workers.php
miina: Updated within a71a0cd.

wp-includes/class-wp-service-workers.php
```diff
@@ -117,6 +117,31 @@ public function init() {
 			array( 'workbox-sw' )
 		);
 
+		$offline_page_id = (int) get_option( WP_Offline_Page::OPTION_NAME, 0 );
+		if ( $offline_page_id ) {
+
+			// Cache relevant theme assets.
+			$this->register_cached_route( '/wp-content/.*\.(?:css|gif|png|jpg|jpeg)', self::STRATEGY_STALE_WHILE_REVALIDATE );
```
westonruter: This will be problematic because it will store all uploaded files in the cache. We'll need to limit this to the actual theme directory URL.

wp-includes/class-wp-service-workers.php
```diff
@@ -125,6 +150,56 @@ public function init() {
 		do_action_ref_array( 'wp_default_service_workers', array( &$this ) );
 	}
 
+	/**
+	 * Configure offline page based on the saved offline page value.
+	 *
+	 * @return string Script.
+	 */
+	protected function configure_offline_page() {
+		$offline_page_id = (int) get_option( WP_Offline_Page::OPTION_NAME, 0 );
```
westonruter: We should short-circuit if this is not a valid page.

wp-includes/class-wp-service-workers.php
```diff
@@ -125,6 +150,56 @@ public function init() {
...
+
+	/**
+	 * Get routes for images found in the content of offline post and for the featured image.
+	 *
+	 * @param WP_Post $post Post object.
+	 * @return array Array of routes.
+	 */
+	protected function get_offline_post_image_routes( $post ) {
+		$routes = array();
+		preg_match_all( '/< *img[^>]*src *= *["\']?([^"\']*)/i', $post->post_content, $matches );
```
westonruter: I think we need to actually extract the attachment image IDs, as we'll need to also obtain the images defined in the `srcset` and `sizes` attributes (via `wp_get_attachment_image_srcset()` and `wp_calculate_image_sizes()` respectively).

wp-includes/class-wp-service-workers.php
```diff
@@ -125,6 +150,56 @@ public function init() {
...
+				);
+			}
+		}
+
+		$featured_img_url = get_the_post_thumbnail_url( $post->ID );
+		if ( false !== $featured_img_url ) {
+			$routes[] = array(
+				'url'      => $this->remove_url_scheme( $featured_img_url ),
+				'revision' => $post->post_modified,
+			);
```
westonruter: Per above, we'll also need to calculate the image URLs for the `srcset` and `sizes`, as the `post-thumbnail` image may not actually be the size used on the offline page.

wp-includes/js/offline-page-handling.template.js
```diff
@@ -0,0 +1,26 @@
+/* global OFFLINE_PAGE_URL */
+( () => {
```
westonruter: My old JS habits are carrying through here. We only need to wrap in a `{` block since the `const` variables are scoped lexically. No need for IIFE.

wp-includes/class-wp-service-workers.php
```diff
@@ -117,6 +117,31 @@ public function init() {
...
+			$this->register_cached_route( '/wp-content/.*\.(?:css|gif|png|jpg|jpeg)', self::STRATEGY_STALE_WHILE_REVALIDATE );
+			$this->register(
+				'offline-sw',
+				array( $this, 'configure_offline_page' ),
+				'workbox-sw'
+			);
+
+			// Register precaching for offline page route.
+			$offline_post         = get_post( $offline_page_id );
+			$content_asset_routes = $this->get_offline_post_image_routes( $offline_post );
```
westonruter: We can include the offline page here as well.

wp-includes/js/offline-page-handling.template.js
```diff
@@ -0,0 +1,26 @@
...
+
+	// Add custom offline / error response serving.
+	const networkFirstHandler = wp.serviceWorker.strategies.networkFirst( {
```
westonruter: I think we need to re-use `precache` as the `cacheName` for this so that it will use what was already cached.

wp-includes/class-wp-service-workers.php
```diff
@@ -117,6 +117,31 @@ public function init() {
...
+			$this->register_precached_routes( array_merge(
+				array(
+					array(
+						'url'      => $this->remove_url_scheme( get_the_permalink( $offline_page_id ) ),
```
westonruter: I think actually we won't need `remove_url_scheme` because all of the URLs should be getting served as HTTPS anyway.

PR #53, opened 2018-08-18
Remove user-specified offline page; add static theme-overridable offline/500 templates

### LINKED ISSUE #50: Idea: Add support for custom theme template file to be used for offline page
Although you can use a regular page as the offline page and choose a page template it might be desirable to let theme authors create a template file specifically for offline use as it could be quite different, eg. with no menus etc.

There are a couple of suitable options for this:

- `503.php` - service unavailable, also send the same status code? Not sure this is technically accurate to use, as it could be either the site or user who is offline
- `offline.php` - not status code specific and self explanatory

### LINKED ISSUE #52: Serve static offline/500 template as opposed to using specially-designated page
We had quite a bit of back-and-forth in #9 about whether to serve a static `offline.php` template or a serve a designated offline page, but went ahead with going down the route of the latter: letting the user to select the offline page. We merged the ability to select the offline page (#23 and #45) and to be able to serve this page when offline (#48).

Nevertheless, through the implementation of this ability we've discovered some problems:

* The `page.php` template could be including assets that would not be precached, and would thus be broken.
* A user could supply content (e.g. a video) that would not be precached, and thus be broken.
* Reliably precaching the header image which would be served on the offline page can be a challenge, especially if the random header option is selected.
* The designated offline page could be an existing page, and in that case should it be removed from nav menus (#46)? Should the selected offline page be excluded from search results and being added to nav menus in the first place? What if they want to use the contact page as the offline page?

These complications have circled things back to what @nico-martin recommended in https://github.com/xwp/pwa-wp/issues/9#issuecomment-403241008 as part of the offline page:

> I think the `offline.php` template would be the better solution than asking the user to pick an offline-page.

Basically this debate comes down to two separate precedences in core for serving a template: do we follow the precedence of the privacy page, or do we follow the precedence of the 404 template? It should be noted that while case could be made for allowing a custom page to be designated to be served on the 404 template, core does not provide that ability (though it could be provided by a plugin). So this feature plugin providing a way to select a custom offline page now seems strange in this way.

If going the static template route, we have another challenge if we want to preserve the theme's design as much as possible. If we look at an example [`404.php`](https://github.com/WordPress/wordpress-develop/blob/4.9.8/src/wp-content/themes/twentyseventeen/404.php) template, we can see it is not as simple as just:

```php
<?php get_header(); ?>
<p>Not found!</p>
<?php get_footer(); ?>
```

No, there are multiple levels of wrapper elements that are required by the theme in order to get the hard-coded “Not Found” content to be laid out and styled properly. So we wouldn't be able to introduce an `offline.php` that could be faithful to the theme's required markup structure. But even if we did, there would be a chance that content in the header or footer that would fail to precache anyway.

One of the arguments I had about going with a user-selected page is that it would allow us to fully use the template hierarchy, where a theme could still fully be served with a single `index.php`. In such a theme, the 404 case can be built in by having a standard `if ( ! have_posts() )` condition. If a user-selected page were used, then it would be included in The Loop™ like any other content, and there wouldn't be the need for an additional hard-coded condition like `elseif ( is_offline() )`.

Nevertheless, a theme that only contains an `index.php` template is not exclusively using this template on a site. Enter the [`theme-compat` directory](https://github.com/WordPress/wordpress-develop/tree/master/src/wp-includes/theme-compat), and in particular [`embed.php`](https://github.com/WordPress/wordpress-develop/blob/master/src/wp-includes/theme-compat/embed.php). When calling `locate_template()` it will first check if a given template exists in the child theme and if not it checks the parent theme. If it doesn't exist in either of those locations, it then checks the `theme-compat` directory. And this is where the post embed template is served from in most WordPress themes, since `embed.php` is normally not implemented. And this is a very similar situation to where themes will be once an offline page is introduced.

So the best solution for proposed core merge is to design a good-looking generic `offline.php` (and `500.php`) be located in the `theme-compat` directory to be served when the user/site is offline. If a theme defines a `offline.php` then they will have to do additional work to make sure that any assets used on this page are included among those that are precached. Additionally, **a plugin should be able to override which URL is used for the offline page so that they can provide the ability to select a custom page**.

So in the end a theme would not have to define an `offline.php` to see the site-specific offline page, but they should define one if they want to have the offline page styled according to their theme, instead of a generic one.

westonruter:
Fixes #52. Fixes #50.

For an example of how to supply custom `500.php` and `offline.php` templates in a theme, see this PR for a child theme of Twenty Seventeen: https://github.com/westonruter/twentyseventeen-westonson/pull/1

For an example of how the AMP plugin can extend, see https://github.com/Automattic/amp-wp/pull/1261

The code in this PR is organized to mirror where the code would be incorporated into a core patch.

* Removes code for user-selected offline page from #28 and #48. It was determined in #52 that this logic is better suited to a plugin rather than core. For the proposed core merge, only static templates will be served instead (similar to `404.php`).
* Adds `error.php` as a root in the template hierarchy (similar to `embed.php`), with `offline.php` and `500.php` as more specific templates.
* Add default styles for the `error.php` template to match the style of `wp_die()`.
* Includes default `error.php`, `offline.php`, and `500.php` to a `theme-compat` directory so that they will be used if no template file exists in the theme. There are `header-error.php` and `footer-error.php` templates to be used for this template, similar again to how there are `header-embed.php` and `footer-embed.php`; these new header files have `error_head` and `error_footer` actions to correspond with the `embed_head`/`wp_head` and `embed_footer`/`wp_footer` actions respectively. 
* Adds `is_500()` to accompany `is_offline()`. Set `$wp_query->is_500` and `$wp_query->is_offline`.
* Adds `get_offline_template()` and `get_500_template()` to follow the pattern of other such `get_*_template()` functions in core.
* Adds `wp_error_template` query var and serves the appropriate error template when `?wp_error_template=500` or `?wp_error_template=offline` are in the URL.
* Add filters `wp_offline_error_precache_entry` and `wp_server_error_precache_entry` for modifying the default URLs (and their revisions) for the offline/server error templates used on the frontend, respectively.
* Adds `error` to the body class in addition to `error500` (to follow `error404`) and `offline` when `is_500()` or `is_offline()` are true respectively.
* Sets the document title to include “Internal Server Error“ or “Offline” when `is_500()` or `is_offline()` respectively.
* Add no-robots `meta` tag to error template for good measure.
* In addition to the theme-controlled offline/server error templates, also introduce an admin-specific offline/server error screen that is not overridable by the theme. This will allow for admin-specific functionality to be served by the service worker, including a place where to list which administrative actions can be performed while offline.
* Displays any error message returned on the internal server error response inside of a `details` element on the `500.php` template and server error admin screen. A `wp_print_service_worker_error_details_script()` PHP function is introduced to facilitate passing error details to the template.
* Introduce `wp_front_service_worker` and `wp_admin_service_worker` actions for when serving front/admin works, respectively. Also introduce `wp_service_worker` action to run as well when either worker is being served.
* Deprecate `WP_Service_Workers::register()` in favor of `WP_Service_Workers::register_script()`.
* Remove redundant `wp_register_routes_precaching()` function. 
* Deprecate `WP_Service_Workers::register_precached_routes( $entries )` in favor of `WP_Service_Workers::register_precached_route( $url, $options )`. The `$options` array can have a `revision` and `cache` keys, where the `cache` can be “precache” or “runtime” but defaulting to the former. If “runtime” then the `revision` key is not currently allowed.
* Introduce `WP_Service_Workers::register_precached_scripts()` and `WP_Service_Workers::register_precached_styles()` methods to facilitate precaching already-registered scripts and styles. 
* Introduce `WP_Service_Workers::register_precached_custom_background()`, `WP_Service_Workers::register_precached_custom_header()`, `WP_Service_Workers::register_precached_custom_logo()`, and `WP_Service_Workers::register_precached_site_icon()` helper methods.
* Enqueue frontend scripts when serving front service worker ~and enqueue admin scripts when serving admin service worker~ (cannot since it not run in admin context; this may need to be revisited for Turbo mode). Doing so ensures that the scripts (and styles) will exist.
* Add support for `precache` argument for registered scripts and styles, and automatically precache them at `wp_front_service_worker` action.
* Add runtime caching of Google Fonts by default via `WP_Service_Workers::register_cached_font_routes()`.
* Precache custom header, logo, background, site icon, and precache-flagged scripts/styles by default. Allow disabling via a call like:
```php
add_action( 'wp_default_service_workers', function( $sw ) { 
    remove_action( 'wp_front_service_worker', array( $sw, 'register_precached_custom_header' ) ); 
} );
```
* Postpone serving worker from `parse_request` to `parse_query` to give opportunity for more scripts to be enqueued.
* Improve API for enabling navigation preload: use `wp_service_worker_navigation_preload` filter name, and pass current scope as second param.
* Discontinue registering required base scripts in favor of calling them directly.

# Screenshots

## Frontend Offline Page Template

image

## Frontend Internal Server Error Template

image

# Todo

* [x] Add some default styling to the error template.
* [x] Precache the offline and 500 template URLs.
* [x] Serve the offline and 500 responses when appropriate.
* [x] Populate the error details on `500.php` with any information available.
* [x] Opt to remove 'ver' in favor of having arg included among ignoreUrlParametersMatching.
* [x] Declare script/style preloading via `wp_scripts()->add_data( 'foo', 'precache', true );`
* [x] Uncaught (in promise) DOMException: Failed to register a ServiceWorker: The script has an unsupported MIME type ('text/html'). 
* [x] Use md5 of 500/offline template path in revision.
* [ ] Prevent requesting frontend offline/server error templates with credentials?
* [ ] Use networkFirst for precached assets when WP_DEBUG enabled.
* [ ] Add tests.


westonruter:
As of now, this PR results in Lighthouse giving PWA a perfect score:

image

westonruter:
Ready for initial review. 

Here's a handy little plugin for simulating PHP errors, so you can just throw `?internal_server_error` onto any URL, whether on the frontend or the admin: https://gist.github.com/westonruter/9b7330fffd24dd61aa1519004a2608b1

wp-includes/class-wp-service-workers.php
```diff
@@ -149,87 +154,95 @@ public function get_current_scope() {
...
-	 *
-	 * @todo It may be unsafe to use staleWhileRevalidate caching here. It is needed for caching background images in stylesheets. It may be safer to do networkFirst, at least when SCRIPT_DEBUG is enabled.
-	 */
-	protected function add_offline_page_caching() {
-		$offline_page = $this->get_offline_page();
-		if ( ! $offline_page ) {
-			return;
+		$revision = sprintf( '%s-v%s', get_template(), wp_get_theme( get_template() )->Version );
+		if ( get_template() !== get_stylesheet() ) {
+			$revision .= sprintf( ';%s-v%s', get_stylesheet(), wp_get_theme( get_stylesheet() )->Version );
```
hellofromtonya: Is there a need to call `get_template()` and `get_stylesheet()` multiple times here?  Each time we call it, it's passing through the filters and consuming a bit of processing time.  Could we instead store it into a variable and then reuse the variable here?

wp-includes/class-wp-service-workers.php
```diff
@@ -521,6 +384,209 @@ public function register_precached_routes( $routes ) {
...
 	}
 
+	/**
+	 * Register scripts which are pre-cached.
+	 *
+	 * @todo Consider storing the script $handles to refer to later by the offline page for what to dequeue.
+	 * @since 0.2
+	 * @param array $handles Script handles.
+	 */
+	public function register_precached_scripts( $handles ) {
```
hellofromtonya: @westonruter For these new method, do you think we should provide functionality, i.e. function or hook, to make it easier for theme developers to register their scripts and styles.  I see confusion around _when_ to do the registration process.  For example, should the assets be enqueued and then registered?

Or better yet, could we _add_ configuration parameter to mark scripts and styles for precaching?

wp-includes/class-wp-service-workers.php
westonruter: Yeah, good call. Fixed in 2aed1a9f2a19ac329fa256dff47df8d063df82ae.

wp-includes/class-wp-service-workers.php
```diff
@@ -271,6 +271,13 @@ public function get_workbox_script() {
 		);
 		$script .= sprintf( "workbox.setConfig( %s );\n", wp_json_encode( $options, 64 /* JSON_UNESCAPED_SLASHES */ ) );
 
+		$cache_name_details = array(
+			'prefix' => 'wordpress',
+			'suffix' => 'v1',
+		);
```
hellofromtonya: 👍 I like this.  Good job!

wp-includes/class-wp-service-workers.php
hellofromtonya: Thanks for the link to how you envision it being used.  I knew you were working on an example theme implementation.

>I really like like the idea of a new script/style attribute like “preload”.

>And then the service worker could automatically look for any such styles that have this precache flag to include it, and then this list of styles/scripts would then be used to obtain the list of script/style handles to pass to this method automatically.

Right, it seems to fit into how theme developers enqueue.  At the point where you enqueue, tell us if you want to pre-cache or not. Then the plugin (and later core) handles it for you.

westonruter:
@hellofromtonya In regards to https://github.com/xwp/pwa-wp/pull/53#discussion_r212405703 I've switched to using `precache` data on the given registered scripts and styles. That makes it much cleaner. See updated PR: https://github.com/westonruter/twentyseventeen-westonson/pull/1

wp-includes/class-wp-service-workers.php
```diff
@@ -436,11 +443,18 @@ public function register_precached_scripts( $handles ) {
...
-	 * @todo Consider storing the style $handles to refer to later by the offline page for what to dequeue.
 	 * @since 0.2
-	 * @param array $handles style handles.
+	 * @param array $handles style handles. If empty, then registered scripts with precache enabled will be used.
 	 */
-	public function register_precached_styles( $handles ) {
+	public function register_precached_styles( $handles = array() ) {
+		if ( empty( $handles ) ) {
+			foreach ( wp_styles()->registered as $handle => $dependency ) {
+				if ( ! empty( $dependency->extra['precache'] ) ) {
```
hellofromtonya: @westonruter 👍

wp-includes/class-wp-service-workers.php
```diff
@@ -399,11 +399,18 @@ public function register_precached_routes( $routes ) {
...
-	 * @todo Consider storing the script $handles to refer to later by the offline page for what to dequeue.
 	 * @since 0.2
-	 * @param array $handles Script handles.
+	 * @param array $handles Script handles. If empty, then registered scripts with precache enabled will be used.
 	 */
-	public function register_precached_scripts( $handles ) {
+	public function register_precached_scripts( $handles = array() ) {
+		if ( empty( $handles ) ) {
+			foreach ( wp_scripts()->registered as $handle => $dependency ) {
+				if ( ! empty( $dependency->extra['precache'] ) ) {
```
hellofromtonya: @westonruter 👍

hellofromtonya:
>In regards to 53 (comment) I've switched to using precache data on the given registered scripts and styles. That makes it much cleaner. See updated example PR: westonruter/twentyseventeen-westonson#1

@westonruter I like that solution much better. Good job 👍

westonruter:
@hellofromtonya I think we should probably merge this PR now as is and iterate more in other PRs. It's getting too big to manage now. I feel pretty good about it as a mergeable state given that there is a theme and plugin successfully using the API.

wp-includes/class-wp-service-workers.php
```diff
@@ -522,30 +480,293 @@ public function register_precached_routes( $routes ) {
...
 
-			$routes_list[] = $route;
+		$attachment = get_post( get_option( 'site_icon' ) );
+		if ( ! $attachment ) {
+			return 0;
 		}
-		if ( empty( $routes_list ) ) {
-			return '';
+
+		$image_urls = array_unique( array(
```
hellofromtonya: @westonruter Why just these sizes?  Do we want to provide a filter that allows a theme developer to modify the site icon sizes, i.e. an array of sizes?  Using that strategy, we could then iterate over sizes to get the URL, register the precached route, and increment the count.

wp-includes/template-loader.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @since 0.2
+ *
+ * @param string $template Template.
+ * @return string Template to include.
+ */
+function _pwa_filter_template_include( $template ) {
+	$located_template = null;
+	if ( is_offline() ) {
+		$located_template = pwa_locate_template( array( 'offline.php', 'error.php' ), false );
```
hellofromtonya: @westonruter Right now, we are repeating which templates are associated with `is_offline` and `is_500`.  Would it be better to centralize each and then use that mechanism (constant or function) to get the collection?

wp-includes/template-loader.php
```diff
@@ -0,0 +1,31 @@
...
+	} elseif ( is_500() ) {
+		$located_template = pwa_locate_template( array( '500.php', 'error.php' ), false );
```
hellofromtonya: Same questions as above.

wp-includes/class-wp-service-workers.php
```diff
@@ -239,206 +264,83 @@ protected function get_offline_page_serving_script() {
...
+	 * @link https://github.com/GoogleChrome/workbox
 	 *
 	 * @return string Script.
 	 */
-	public function get_workbox_script() {
+	protected function get_base_script() {
 
-		$workbox_dir = 'wp-includes/js/workbox-v3.4.1/';
+		$current_scope = $this->get_current_scope();
+		$workbox_dir   = 'wp-includes/js/workbox-v3.4.1/';
```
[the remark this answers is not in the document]
westonruter: It actually won't be referenced anywhere else than this method. Since we define `modulePathPrefix` then the library will henceforth know how to load its own modules. Themes and plugins should not know or care where Workbox is located. So I don't think we need a constant or method in this case.

wp-includes/class-wp-service-workers.php
westonruter: Good question. I'll add a note explaining that the URLs here are those which are used in `wp_site_icon()`. Inside that function, there is also a `site_icon_meta_tags` filter that applies where for plugins to add new icon sizes potentially. But that will then require us to (1) do output buffering to capture, and (2) parse out the `href` attributes to obtain the URLs. There's not a clean way to just get the actual URLs. So I'll add a comment indicating that this is something we could do in the future, but for now I think this is sufficient.

Side note: it seems Chrome doesn't even respect the service worker's cache of the site icon when I go offline.

wp-includes/template-loader.php
westonruter: Yes, good call. We need a `get_offline_template()` and `get_500_template()` to coincide with the other `get_*_template()` functions.

wp-includes/class-wp-service-workers.php
westonruter: Comment added in 4193e499b5d313cc2c944851c51867b31a642ad5

wp-includes/class-wp-query.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+/**
+ * Parse query for an error template request.
+ *
+ * @param WP_Query $query Query.
+ */
+function pwa_parse_query_for_error_template( WP_Query $query ) {
+	$error_template = $query->get( 'wp_error_template' );
+	if ( in_array( $error_template, array( 'offline', '500' ), true ) ) {
```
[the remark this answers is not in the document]
westonruter: Done in 4193e499b5d313cc2c944851c51867b31a642ad5

wp-includes/template-loader.php
westonruter: Done in c9e6a1b52d92e1b299861430c48334f16b26cf0e

wp-includes/template-loader.php
westonruter: See c9e6a1b52d92e1b299861430c48334f16b26cf0e

PR #54, opened 2018-08-27
Conditionally redirect front-end requests to HTTPS

kienstra:
**Request For Review**

Hi @hellofromtonya, could you please review this follow-up to #17?

Lighthouse reported that this does not redirect HTTP to HTTPS:
redirect-to-https

So this redirects to HTTPS:
* Only if the user has checked the 'Upgrade' checkbox, and the last loopback request to check HTTPS support succeeded.
* If `! is_admin()` , so if the SSL certificate expires somehow, admins won't be locked out of `/wp-admin`.

### DIFF
--- a/wp-includes/class-wp-https-ui.php
+++ b/wp-includes/class-wp-https-ui.php
@@ -82,6 +82,7 @@ class WP_HTTPS_UI {
 		add_action( 'admin_init', array( $this, 'init_admin' ) );
 		add_action( 'init', array( $this, 'filter_site_url_and_home' ) );
 		add_action( 'init', array( $this, 'filter_header' ) );
+		add_action( 'wp_loaded', array( $this, 'conditionally_redirect_to_https' ) );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


kienstra:
Thanks, @hellofromtonya!

wp-includes/class-wp-https-ui.php
```diff
@@ -333,4 +334,27 @@ public function upgrade_insecure_requests( $headers ) {
...
+
+	/**
+	 * Conditionally redirects HTTP requests to HTTPS.
+	 *
+	 * Does not apply if is_admin().
+	 * So if the SSL certificate expires somehow, the admin won't be locked out of wp-admin.
+	 */
+	public function conditionally_redirect_to_https() {
+		$do_redirect = (
+			isset( $_SERVER['REQUEST_SCHEME'], $_SERVER['REQUEST_URI'] ) && 'https' !== $_SERVER['REQUEST_SCHEME']
```
westonruter: @kienstra Is `$_SERVER['REQUEST_SCHEME']` right here? Should it not be using `! is_ssl()`?

kienstra: Good question, @westonruter. I'll get back to you in several hours if that's alright.

kienstra:
**Response regarding canonical_redirect()**

>@kienstra I thought that redirect_canonical() should have handled this redirection automatically. Maybe the filters aren't happening early enough?

Hi @westonruter,
Good point. I haven't seen `redirect_canonical()` redirect HTTP URLs to HTTPS locally on [VVV](https://github.com/Varying-Vagrant-Vagrants/VVV), either when using this plugin, or by manually changing the `home` and `siteurl` values to `HTTPS`.

Some sites do redirect based on the `home` and `siteurl` values being HTTPS (when not using this plugin). 

For example:
http://paired-ampconfdemo.pantheonsite.io/
redirects to:
https://paired-ampconfdemo.pantheonsite.io/

But locally on [VVV](https://github.com/Varying-Vagrant-Vagrants/VVV), if I manually set the `home` and `siteurl` to HTTPS, a HTTP URL (like the homepage) doesn't redirect to HTTPS (without this PR).

wp-includes/class-wp-https-ui.php
kienstra: Hi @westonruter, PR #55 changes this from `$_SERVER['REQUEST_SCHEME']`to `! is_ssl()`. If you have any more suggestions, I could also apply them there.

PR #57, opened 2018-09-05
Improve service worker classes responsibility

felixarntz:
This PR aims at improving the organization and clarifying responsibilities of service worker classes. `WP_Service_Workers` currently handles way too many different parts:
* the general entry point to service worker functionality, via script registration and service worker request serving methods
* registration of cached and precached routes
* implementation of default integrations that use the above APIs

As a start, it would be great to get these three things separated. Since version 0.2 has not been released yet, there shouldn't be backward-compatibility issues here. I made sure that all changes are backward-compatible with the 0.1 version (most of it is new anyway though).

What this PR does exactly:
* [x] Introduce a `WP_Service_Worker_Registry` class. This class contains methods to register additional functionality that can then be used by service workers.
* [x] Introduce a `WP_Service_Worker_Integration` interface and, based on that, classes for all default integrations, which replace the methods that were previously part of the `WP_Service_Workers` class. Having such an interface makes it more obvious how service worker integrations can easily be built. Furthermore, the separate classes ensure that the implementations remain separate from the actual API.

Tests for the `WP_Service_Worker_Registry` changes are present, and all existing changes pass. Let me know if we should also add tests for all the default integrations.


felixarntz:
I have now implemented all default integrations, based on a `WP_Service_Worker_Integration` interface and a `WP_Service_Worker_Base_Integration` abstract class. Some notes:
* Each integration has a `register()` method (based on the interface), which receives the `WP_Service_Worker_Registry` instance.
* A new `service-worker-loader.php` file contains a `wp_default_service_workers()` function that is hooked into `wp_default_service_workers`.
* In that function, some basic `WP_Service_Workers` properties are set, similar as in core's `wp_default_scripts()` function.
* In that function, integrations are loaded. The default integrations are filtered by a new `wp_service_worker_integrations`, which allows to easily remove default integrations or add custom ones.

As a summary, all functionality remains the same compared to before this PR. However, the code being more organized and having separated concerns now will help provide a better foundation for the future.

westonruter:
> A new service-worker-loader.php file contains a wp_default_service_workers() function that is hooked into wp_default_service_workers.

I moved `wp_default_service_workers()` into `service-workers.php` since this is the file that contains other service worker functions. I didn't see a need for another file that just contained one function.

wp-includes/class-wp-service-workers.php
```diff
@@ -886,34 +536,36 @@ public function serve_request( $scope ) {
...
+			 *         unset( $integrations['wp-custom-header'] );
+			 *         return $integrations;
 			 *     } );
 			 *
 			 * @since 0.2
-			 * @param WP_Service_Workers $this
+			 * @param WP_Service_Worker_Registry $registry Instance to register service worker behavior with.
 			 */
-			do_action( 'wp_front_service_worker', $this );
+			do_action( 'wp_front_service_worker', $this->registry );
```
westonruter: @felixarntz I think it is actually better to continue passing `$this`. Any themes or plugins can then just access `$service_workers->registry`. Plugins that are hooking into this may not care about the registry anyway, just wanting to manual low-level do a `$service_workers->register_script()` call.

felixarntz: Hmm, I prefer passing a more scoped object. A reason for me introducing the registry class was to have a more simple class which only contains methods that should actually be called by `wp(_front|_admin)?_service_worker` callbacks. Currently this would be for cached and precached routes.

Registering a script already requires passing a scope anyway, and continues to be accessible via a more generic `wp_service_workers()->register_script()`. I'd argue that registering scripts should be the responsibility of `WP_Service_Workers`, which makes sense because that extends `WP_Scripts`.

I prefer to keep passing the registry, supporting functionality as folows:
* Service worker scripts should preferably be added via the `wp_default_service_workers` hook, calling `register_script()` on the passed instance (`WP_Service_Workers`).
* Service worker cached routes should preferably be added on one of the targeted `wp(_front|_admin)?_service_worker` hooks, calling `register_cached_route()` on the passed instance (`WP_Service_Worker_Registry`).
* Service worker precached routes should preferably be added on one of the targeted `wp(_front|_admin)?_service_worker` hooks, calling `register_precached_route()` on the passed instance (`WP_Service_Worker_Registry`).

What do you think?

westonruter: The `wp_front_service_worker` action (and others) are triggered inside of `WP_Service_Workers` not inside of `WP_Service_Worker_Registry`, so it makes more sense to me for the former to be passed as the argument than the latter.

> A reason for me introducing the registry class was to have a more simple class which only contains methods that should actually be called by wp(_front|_admin)?_service_worker callbacks. Currently this would be for cached and precached routes.

Right, but that's not necessarily the case. Someone may want to call `wp_service_workers()->register_script()` at the `wp_front_service_worker` action. True they could also do this at the `wp_service_worker` action if they then then pass the `register_script()` method's `$scope=SCOPE_FRONT` argument as well, as they could also do this at the `wp_default_service_workers` action. (Maybe we have too many entry points here.) 

But inside of the `wp_front_service_worker` action, if someone _does_ want to register a script then they'd have to reach out to the global `wp_service_workers()` function at present, since the `$registry` object doesn't have a reference to the `WP_Service_Workers` instance. This isn't necessarily un-WordPressy, but it would seem that if we can restrict ourselves to interacting with the explicitly-passed parameters then it would be better.

I'm not certain one way or the other.

felixarntz: I agree that the actions should receive an object that provides access to all methods which are intended to be called from there. However, the `WP_Service_Workers` class at the moment still has too many responsibilities and contains many methods that should never be called from those hooks. So I think neither of the two objects is a truly solid solution.

Maybe we could do the following:
* Introduce a concept of different registries that all only handle one type of resource. So there would be a `WP_Service_Worker_Cache_Registry`, `WP_Service_Worker_Precache_Registry` (splitting the current `WP_Service_Worker_Registry` functionality into these two), and a `WP_Service_Worker_Script_Registry`. The latter could internally call `wp_service_workers()->register_script()`, which would be ugly, but having this in place would allow us to refactor in the future.
* Each registry should follow a uniform infrastructure, with similar method names (for example each of the (above three) registry classes would have a `register()` and a `get_all()` method for now).
* Then we would introduce an aggregate registry / a container of multiple registries. That could then actually carry the name of `WP_Service_Worker_Registry`. It would have methods that provide access to the individual registries such as `cache()`, `precache()`, `script()`. This registry class would be the object that is passed to all the service worker hooks (as `$registry`). A hook callback could then do `$registry->cache()->register( ... )`, `$registry->precache()->register( ... )`, or `$registry->script()->register( ... )`. The aggregate registry would easily support further registries that may be needed in the future, without cluttering it with too much functionality.
* An even more memorable solution it would be if we could find a way to have the same signatures for all `register()` methods (in which case we should build an interface for it). Maybe a `$handle` and `$args` for all of them, or only `$args`. _If_ we can figure something like this out, it would be extra great, but is not absolutely necessary otherwise.

The only downside I see with the above approach is that `WP_Service_Workers` still contains the script registration functionality, but I think for backward-compatibility reasons we shouldn't change them _yet_. IMO this would be a necessary evil though, and by recommending explicitly to always go through the aggregate registry passed to the actions, we build a path to change that later.

Let me know your thoughts. Also, if there should be a new issue to go about this. :)

westonruter: @felixarntz As we chatted about earlier today, I like what you have proposed. Moving the service worker request/response handling could also be moved to a `WP_Service_Worker_Server` class which could be modeled after `WP_REST_Server`. I look forward to your PR 😄

felixarntz: @westonruter I'll work on it Tuesday or Wednesday.

PR #58, opened 2018-09-05
Introduce hook to short-circuit installation of service workers

felixarntz:
While service workers are typically installed via JavaScript and `navigator.serviceWorker`, this is not always applicable and possible. When using AMP for example, the `amp-install-serviceworker` tag should be used instead.

Therefore this PR introduces a `wp_install_service_workers` filter that allows to short-circuit the default implementation. In the hook callback, then a custom implementation can run.

With this filter in place, we can easily add support for service workers to the https://github.com/Automattic/amp-wp plugin.

### DIFF
--- a/wp-includes/service-workers.php
+++ b/wp-includes/service-workers.php
@@ -104,6 +104,27 @@ function wp_print_service_workers() {
 		return;
 	}
 
+	/**
+	 * Filters whether service workers should be installed.
+	 *
+	 * This filter can be used to install service workers through another means than the default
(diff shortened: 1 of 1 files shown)


felixarntz:
@westonruter Ah, sounds good then. Maybe it would require less extra code if we could hook into the process more closely. But it appears that the logic in the AMP PR is already far progressed, so feel free to close this.

westonruter:
We can always revisit later if it turns out short-circuiting needs to be more of a first-class citizen. But since we don't want the service worker to be short-circuited by default, we shouldn't make it too easy to do something that we don't want devs to be doing generally.

PR #65, opened 2018-09-11
Improve HTTPS detection and opt-in

westonruter:
* Always do HTTPS loopback tests even when HTTPS is supposedly enabled.
* Always show checkbox, even when HTTPS is currently enabled. (As it could be enabled temporarily.)
* When doing HTTPS loopback requests, store any errors and show them next to the checkbox.
* Allow HTTPS checkbox to be submitted even when there are errors. This is critical for self-signed certificates as in a development environment.
* Do HTTPS redirection at `template_redirect` instead of `wp_loaded`.

image

Also:

* Prevent HTML entities from being served in name, short_name, and description fields in web app manifest.

### DIFF
--- a/wp-includes/class-wp-https-detection.php
+++ b/wp-includes/class-wp-https-detection.php
@@ -57,7 +57,7 @@ class WP_HTTPS_Detection {
 	 * Initializes the object.
 	 */
 	public function init() {
-		add_action( 'wp', array( $this, 'schedule_cron' ) );
+		self::schedule_cron();
 		add_action( self::CRON_HOOK, array( $this, 'update_https_support_options' ) );
 		add_filter( 'cron_request', array( $this, 'conditionally_prevent_sslverify' ), PHP_INT_MAX );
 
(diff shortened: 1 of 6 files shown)


wp-includes/class-wp-https-ui.php
```diff
@@ -157,6 +149,17 @@ public function render_https_settings() {
 			) );
 			?>
 		</p>
+
+		<?php if ( is_wp_error( $https_support ) ) : ?>
+			<?php foreach ( $https_support->get_error_messages() as $error_message ) : ?>
+				<div class="notice notice-error inline">
+					<p>
+						<?php echo esc_html( $error_message ); ?>
```
kienstra: It's great how this displays the error message(s):


error-message-j

wp-includes/class-wp-https-detection.php
```diff
@@ -69,7 +69,7 @@ public function init() {
 	 * Schedules a cron event to check for HTTPS support.
 	 */
 	public function schedule_cron() {
-		if ( ! wp_next_scheduled( self::CRON_HOOK ) && ! $this->is_currently_https() ) {
```
kienstra: ensure-secure-connection

wp-includes/class-wp-https-detection.php
```diff
@@ -57,7 +57,7 @@ class WP_HTTPS_Detection {
 	 * Initializes the object.
 	 */
 	public function init() {
-		add_action( 'wp', array( $this, 'schedule_cron' ) );
+		self::schedule_cron();
```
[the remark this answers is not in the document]
westonruter: You're right. This should not be a static call.

wp-includes/class-wp-https-detection.php
westonruter: Maybe we should just have it always be “Force secure connections” regardless of whether HTTPS is on.

kienstra: That sounds good, as it will cover whether HTTPS is enabled or not.

PR #69, opened 2018-09-12
Update Web App Manifest endpoint to /wp/v2/web-app-manifest

adrianocalvitto:
Just looked at a couple of projects I am following, they actually register new endpoints to WP Api using `app` as part of the namespace.

Wouldn't be `app` more suitable as general namespace for theme-registered endpoints?

Wouldn't be `pwa` (official plugin slug) more appropriate the plugin's rest namespace? 

Adriano

### DIFF
--- a/tests/test-class-wp-web-app-manifest.php
+++ b/tests/test-class-wp-web-app-manifest.php
@@ -36,7 +36,7 @@ class Test_WP_Web_App_Manifest extends WP_UnitTestCase {
 	 *
 	 * @var string
 	 */
-	const EXPECTED_ROUTE = '/app/v1/web-manifest';
+	const EXPECTED_ROUTE = '/pwa/v1/web-manifest';
 
 	/**
 	 * Image mime_type.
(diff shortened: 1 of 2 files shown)


westonruter:
Since this is a feature plugin with the intention of being merged into core, we'd want to use a namespace that should be used in core. Maybe that means we should use the `wp/v2` namespace instead.

adrianocalvitto:
> Since this is a feature plugin with the intention of being merged into core, we'd want to use a namespace that should be used in core. Maybe that means we should use the `wp/v2` namespace instead.

@westonruter makes way more sense :)

westonruter:
Would you like to amend your PR with that change?

westonruter:
The `REST_ROUTE` should probable also change to `web-app-manifest`:

https://github.com/xwp/pwa-wp/blob/f7fb7e8c20ce4af8520bc3a2437b6864c7d5cc37/wp-includes/class-wp-web-app-manifest.php#L34

adrianocalvitto:
@westonruter sure I have some time later 👍

adrianocalvitto:
@westonruter sorry, any clue why that test is failing? No idea...

westonruter:
> sorry, any clue why that test is failing? No idea...

It's related to changes in #65. The test fails randomly. I restarted the job.

PR #70, opened 2018-09-12
Update Service Worker automatically with optional opt-out.

### LINKED ISSUE #35: Facilitate notifying user of updated service worker
An updated service worker defers activating until all existing client windows have gone away (e.g. via reloading browser). There should perhaps be a way to notify the user that updates are available without requiring them to have DevTools open. Such an update prompt is common on web apps, and having a standardized notice in WordPress would probably be a good idea.

In the admin the notice could be shown as an admin notice. On the frontend maybe there is no standard way to show the notice, other than to display something in the admin bar. But this isn't shown to logged-out users.

Read more about the lifecycle at https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle#waiting

miina:
Fixes #35.

This PR aims to:
- Add admin notification if service worker can be updated;
- Add notification to admin bar if service worker can be updated;
- Updates the service worker and reloads the page in case of clicking on the action link in the notification;
- Does all this for logged in user only.

TODO:
- [ ] Design for the frontend notification;
- [ ] Review and if needed refactor the code to match the WP structure better, perhaps use class methods.
- [ ] Test.
- [ ] Update `AMP_Service_Worker` to enable `clientsClaim` and `skipWaiting`.


miina:
Currently the notifications look like this:
- Admin:
screen shot 2018-09-13 at 12 36 23 pm

- Frontend desktop:
screen shot 2018-09-13 at 12 57 49 pm

- Frontend mobile:
screen shot 2018-09-13 at 12 57 40 pm

 As can be seen the frontend notification is quite similar to the admin notice at this moment, is this OK or should we create something different? Thoughts, @westonruter?

westonruter:
What about _inside_ the admin bar itself? Or is the thought that the notice should be shown even when the admin bar is not shown?

westonruter:
Also, I realized that if the service worker does `skipWaiting` that we could potentially avoid having to show the notice in the default case:

https://github.com/xwp/pwa-wp/blob/1cbbaac06d15ba3e02bae03c6648eee2ae9e8a96/wp-includes/class-wp-service-workers.php#L245-L246

Note that the boilerplate service worker for AMP has this by default:

https://github.com/ampproject/amp-by-example/blob/d5aafed56545c4f45ec7cb7bf9115aac99fa09bb/boilerplate-generator/templates/files/serviceworkerJs.js#L3-L4

Maybe it should also be the default for WordPress? It could then be disabled with a filter, similar to how we're handling navigation preload.

westonruter:
> Are you thinking that we could just `skipWaiting` / update SW without displaying a notice at all and instead allow disabling the automatic update by filtering? And the notification would be displayed only when the update is disabled by filter?

Yes, that's right.

> Or that maybe we shouldn't have a default notice at all and a custom one could be added via filtering?

That's a good question. I think it comes down to the prevalence of sites which would need to opt-out of `skipWaiting`. But if `skipWaiting` would be included by default in core, then yes I think we could actually having an update notice. Or at least, we could limit the notice to just the admin bar.

> On the notice's location: It's inside the admin bar in terms of markup and would show only when the admin bar is shown, moved it out with CSS to make it noticeable, however, I'm fine either way. Do you think it would be better to just add it to Admin bar (matching the existing fonts, etc.)?

Yes. I think it would be better in the admin bar as the default. This will ensure it doesn't conflict with other theme elements. If a site disables `skipWaiting` then they could also show their own update notice.

There could be an admin bar item added which is hidden by default, and then is revealed when there is an update. This is similar to the “Customize” admin bar item which is hidden by default (the `hide-if-no-customize` class name) and then shown if JS is available (via `wp_customize_support_script()`).

This would provide an easy way for sites to replace that update notice with their own, since they could just just remove the node from the admin bar.

miina:
@westonruter I added some changes within b33e7d3. Now `updatedSw.postMessage( { action: 'skipWaiting' } );` is called by default, however, it can be opted out by filter `pwa.serviceWorker.skipWaiting`. Also, now `$wp_admin_bar->add_node()` is used for adding the update notice, note that it has no CSS modifications at this moment and displays just as a default link text, the text and CSS likely need improvements.

Thoughts?

wp-includes/js/service-worker.js
```diff
@@ -148,3 +148,10 @@ self.addEventListener( 'fetch', event => {
 		event.respondWith( responsePromise );
 	}
 } );
+
+// Skip the waiting phase for the Service Worker.
+self.addEventListener( 'message', function ( event ) {
+	if ( 'skipWaiting' === event.data.action ) {
```
westonruter: Is sending a message the right way to do this? Should it not be `self.skipWaiting` unconditionally?

For example, in PHP-generated JS that is added to the service worker:

```php
if ( apply_filters( 'wp_service_worker_skip_waiting', true ) ) {
    echo 'self.skipWaiting();';
}
```

I see this similar to how we're managing navigation preload.

wp-includes/service-workers.php
```diff
@@ -141,8 +141,14 @@ function wp_print_service_workers() {
 								// If new service worker is available, show notification.
 								if ( 'installed' === updatedSw.state ) {
 									if ( navigator.serviceWorker.controller ) {
-										var notification = document.getElementById( 'pwa-sw-update-notice' );
-										jQuery( notification ).removeClass( 'hidden' );
+
+										// Allows opting out from updating the service worker automatically.
+										if ( wp.hooks.applyFilters( 'pwa.serviceWorker.skipWaiting', true ) ) {
```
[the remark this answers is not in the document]
westonruter: Correction: this filter should be done in PHP that generates the service worker to begin with per above.

wp-includes/service-workers.php
```diff
@@ -126,13 +126,43 @@ function wp_print_service_workers() {
...
+								// If new service worker is available, show notification.
+								if ( 'installed' === updatedSw.state ) {
+									if ( navigator.serviceWorker.controller ) {
+
+										// Allows opting out from updating the service worker automatically.
+										if ( wp.hooks.applyFilters( 'pwa.serviceWorker.skipWaiting', true ) ) {
+											updatedSw.postMessage( { action: 'skipWaiting' } );
+										} else {
+											var notification = document.getElementById( 'wp-admin-bar-pwa-sw-update-notice' );
+											jQuery( notification ).removeClass( 'hidden' );
```
westonruter: We should not rely on jQuery. Just do `notification.hidden = false;`

wp-includes/service-workers.php
```diff
@@ -126,13 +126,43 @@ function wp_print_service_workers() {
...
+							updatedSw = reg.installing;
+							updatedSw.addEventListener( 'statechange', () => {
+
+								// If new service worker is available, show notification.
+								if ( 'installed' === updatedSw.state ) {
+									if ( navigator.serviceWorker.controller ) {
+
+										// Allows opting out from updating the service worker automatically.
+										if ( wp.hooks.applyFilters( 'pwa.serviceWorker.skipWaiting', true ) ) {
+											updatedSw.postMessage( { action: 'skipWaiting' } );
```
westonruter: I see this being useful if the user clicks on the update link. But the PHP filter `wp_service_worker_skip_waiting` filter should manage whether or not the notification is displayed at all, right? If the filter returns true then it should just `skipWaiting` whenever there is an update?

wp-includes/js/service-worker.js
miina: Changed this within f1ca81a.

wp-includes/components/class-wp-service-worker-configuration-component.php
```diff
@@ -78,8 +78,21 @@ public function get_script() {
...
+		 * For optioning out from skipping waiting and displaying a notification to update instead, you could do:
+		 *
+		 *     add_filter( 'wp_service_worker_skip_waiting', '__return_false' );
+		 *
+		 * @param bool $skip_waiting Whether to skip waiting for the Service Worker and update when an update is available.
+		 */
+		$skip_waiting = apply_filters( 'wp_service_worker_skip_waiting', true );
+		if ( true === $skip_waiting ) {
+			$script .= "workbox.skipWaiting();\n";
+			$script .= "workbox.clientsClaim();\n";
```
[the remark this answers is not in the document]
miina: In [this guide](https://developers.google.com/web/tools/workbox/guides/generate-complete-sw#skip_waiting_and_clients_claim) updating seems to use these together, thought that this would be the standard way since it doesn't mention the cases for using `skipWaiting` separately, although it is possible.

According to Jeff Posnick's [answer here](https://stackoverflow.com/questions/51715127/what-are-the-downsides-to-using-skipwaiting-and-clientsclaim-with-workbox):
> `clientsClaim` ensures that all uncontrolled clients (i.e. pages) that are within scope will be controlled by a service worker immediately after that service worker activates. If you don't enable it, then they won't be controlled until the next navigation.

Also, on [this documentation page](https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle#clientsclaim) it mentions: 
> If you use your service worker to load pages differently than they'd load via the network, clients.claim() can be troublesome, as your service worker ends up controlling some clients that loaded without it.

and also
> Note: I see a lot of people including clients.claim() as boilerplate, but I rarely do so myself. It only really matters on the very first load, and due to progressive enhancement the page is usually working happily without service worker anyway.

I guess it is better to have the option to filter those separately!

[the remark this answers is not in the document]
miina: Updated within 8cbc81f.

PR #72, opened 2018-09-14
Rely on `WP_Scripts`, introduce components and improve infrastructure

### LINKED ISSUE #71: Rely on `WP_Scripts` for all plugin functionality and introduce components
@westonruter and I chatted for a bit about the code structure of the plugin and how to reduce complexity while making the functionality more straightforward and flexible at the same time.

Here's my suggestion as a takeaway from this discussion:
* Let's split `WP_Service_Workers` into two different classes:
    * `WP_Service_Worker_Scripts` will contain the script registry part and will extend `WP_Scripts`.
    * `WP_Service_Workers` will contain the `serve()` method and handle everything related to actually serving the service worker for frontend or admin.
* The `wp_service_worker` action should be removed. In terms of timing, it is problematic to have it run independently of the `wp_front_service_worker` / `wp_admin_service_worker` action respectively. A plugin wanting to register scripts for both scopes could simply hook the same callback into both actions.
* The instance of `WP_Service_Worker_Scripts` should be passed to the `wp_front_service_worker` / `wp_admin_service_worker` actions.
* An interface `WP_Service_Worker_Component` should be introduced. Components take care of core service worker functionality, for example outputting configuration via the base script (currently part of `WP_Service_Workers::get_base_script()`, would then move into a `WP_Service_Worker_Configuration_Component`) or outputting the script for handling precaching URLs (would move into a `WP_Service_Worker_Precaching_Routes_Component`). The interface should have a method `serve( WP_Service_Worker_Scripts $scripts )` and a method `get_priority()`. The `WP_Service_Workers` class will be aware of all available components and hook each `serve()` method into both the frontend and admin service worker actions, at the priority returned by `get_priority()`.
* An interface `WP_Service_Worker_Registry` should be introduced. This is a general interface defining a unified behavior for any registerable content. It should be implemented by `WP_Service_Worker_Scripts`, as well as all registries that belong to a specific component, for example the precache component allows other code to register precache URLs (see more below). The interface should have a method `register( $handle, $args = array() )` and a method `get_all()`.
* An interface `WP_Service_Worker_Registry_Aware` should be introduced, with a single `get_registry()` method. This should be implemented by components that allow other code to integrate with themselves. For example, `WP_Service_Worker_Precaching_Routes_Component` would implement this interface in combination with its accompanying `WP_Service_Worker_Precaching_Routes` (which implements `WP_Service_Worker_Registry`, as already hinted above).
* The `WP_Service_Worker_Scripts` class would be instantiated from the `WP_Service_Workers` class, and it would (in addition to its own methods for adding scripts) provide access to all component registries, i.e. all registries returned by the `get_registry()` method of a component (which can easily be detected via the `WP_Service_Worker_Registry_Aware` interface that they would implement. This would allow plugins and other integrations to easily register everything from a centralized point, as the `WP_Service_Worker_Scripts` instance is passed to the service worker actions. A plugin could then for example call `$scripts->precaching_routes()->register( ... )`.
* The integrations (see `WP_Service_Worker_Integration`) as they exist now would almost remain the same. The only thing they would receive in addition would be a `get_priority()` method, in order to determine the priority they are hooked into the actions. The code would also be adjusted to account for the removal of the `wp_service_worker` action (mentioned earlier).

--
These changes will introduce the following clear responsibilities:
* `WP_Service_Workers` as main entry point to the service worker functionality.
* `WP_Service_Worker_Scripts` as the main registry that plugin developers would interact with, via the `wp_front_service_worker` / `wp_admin_service_worker` actions.
* Components handle crucial core functionality for service workers. These are hard-coded and it is not possible for plugins to register additional components. Those are purely for solid code organization and to keep the `WP_Service_Workers` and `WP_Service_Worker_Scripts` clean and preserve their single responsibility. Anything that happens in a component could theoretically be added as a script directly, components just make their respective functionality much easier to use than if one had to manually register a custom script for it.
* Component registries allow to modify components that support registration of their own data.
* Integrations are an alternative object-oriented way of integrating with the plugin's service worker(s). It is simply a layer around hooking a callback into the `wp_front_service_worker` / `wp_admin_service_worker` actions directly. All default integrations (such as for the Custom Logo or Custom Header core features) should use the integration approach, while plugin developers can use whatever of the two approaches they prefer.

Let me know if you have any suggestions or concerns about this. I'll move forward with a PR for these changes tomorrow. With that in place, it might also be easier to see how all of the above comes together.

felixarntz:
Fixes #71.

Please read the issue for in-depth information on what the PR does. Since it is a massive PR due to the structural changes it introduces, for reviewing it might be helpful to follow the individual commits, for which I made sure to follow a logical iterative structure.

With this PR, plugins can integrate with the service worker in the following ways:
* Callbacks for registering service worker functionality must be hooked into either `wp_front_service_worker`, `wp_admin_service_worker`, or both. Alternatively, a plugin can create its own `WP_Service_Worker_Integration` implementation and register it via the `wp_service_worker_integration` filter.
* Custom service worker scripts can be registered with `wp_register_service_worker_script()`, which is the same as `wp_service_workers()->get_registry()->register()`, and also the same as `$scripts->register()` when called from code that is hooked into one of the above two actions. The latter is the recommended way to do so.
* A plugin may also register precaching and caching routes. The registries for those are accessible though `wp_service_workers()->get_registry()`, which is (as indicated above) the same as the `$scripts` instance passed to the above actions. Since those registries are both part of a component, they are passed into the `$scripts` instance's constructor and can be accessed via a magic method (`precaching_routes()` and `caching_routes()` respectively).

As the `serve_request()` method now simply prints all registered scripts, even the very foundational core scripts are registered (via components). The base configuration script is registered at hook priority -999999 to be really early, and the error response script is registered at hook priority -99999 to be really early as well. This allows plugin developers to theoretically hook something in before if they want to do something very custom and they know what they're doing, but generally "ensures" these scripts are printed before anything else. Similarly, the precaching script is registered at hook priority 99999, and the caching script is registered at hook priority 999999, to "ensure" these are printed after anything else.

As the code is now relatively well-structured compared to how WordPress core typically approaches enhancements and new features, we need to keep in mind to have solid reasons and arguments in case this brings up questions at the point of a merge proposal. Examples:
* WordPress has historically not used interfaces, but in this case they help following a coherent structure in all service worker code that makes sense to be similar. Furthermore, using an abstract class would not work, as `WP_Service_Worker_Scripts` for example needs to already extend `WP_Scripts`, and PHP does not allow extending multiple classes.
* Functionality that somewhat belongs to the same area is typically bundled into a single class, causing those to quickly become large and hardly maintainable. With the approach of using individual, small classes that implement a common interface (which is how both components and integrations work) provides a much better overview of what a piece of code is responsible for, and furthermore allows plugin developers to leverage those interfaces as well for their own code.

While the object-oriented structure may indeed be a bit overwhelming for a new developer who wants to actually contribute to the project, this shouldn't present an issue for regular consumption by plugins, as I believe that actually using the underlying APIs is just as simple as using the Customizer API for example.

Regarding tests, so far I only ensured that the existing tests pass, and accounted for the structural changes. Once this PR has been reviewed and there is consensus that we can move forward with it, I'm happy to provide comprehensive tests for good coverage.
It might still be worth merging the PR before that because this allows other PRs that might require changes based on it to move on quicker. In that case, a follow-up issue and PR could focus on improving coverage.


wp-includes/integrations/class-wp-service-worker-base-integration.php
```diff
@@ -18,18 +18,49 @@
...
 
+	/**
+	 * Gets the priority this integration should be hooked into the service worker action with.
+	 *
+	 * @since 0.2
+	 *
+	 * @return int Hook priority. A higher number means a lower priority.
+	 */
+	public function get_priority() {
+		return 10;
```
miina: With the default priority being `10` for styles and scripts wouldn't it mean that if anyone would want to use the `wp_front_service_worker` or `wp_admin_service_worker` for opting in for precaching registered scripts they wouldn't be able to use the filters without knowing to use a priority lower that 10? Since that would be the way how to flag scripts and styles for precaching perhaps it would be better to do it so that it could be used without knowing to set a priority? Thoughts?

miina:
@felixarntz Could you also update the relevant part under `Service Workers` in the `readme`?

When this gets merged we'll also need a more thorough documentation for developers before the next plugin release.

wp-includes/integrations/class-wp-service-worker-base-integration.php
felixarntz: @miina Good catch. At the moment it would indeed be necessary to use `wp_front_service_worker` / `wp_admin_service_worker` at a priority lower than 10 to flag scripts and styles for precaching because after that the `WP_Service_Worker_Scripts_Integration` / `WP_Service_Worker_Styles_Integration` logic will already have run.

On the other hand, I'm not sure though whether that is a big issue. I would think that eventually, scripts and styles should be flagged for being precached when they are registered/enqueued, which is `wp_enqueue_scripts` / `admin_enqueue_scripts`, and this should have executed before the `wp_front_service_worker` / `wp_admin_service_worker` hooks anyway.

Just to play safe, I think it would still make sense to have these two integrations run at a much higher priority value so that they are executed very late. This would mitigate the issue IMO.

felixarntz: I changed the scripts and styles integrations to run at priority 10000 to address this.

felixarntz:
Something we should discuss: Currently all `register()` methods accept a handle string and an array of arguments, per the interface definition. However, I also added support for passing a string as second parameter instead. If someone does that, ...
* for scripts it is interpreted as the `$src` argument
* for precaching routes it is interpreted as the `$revision` argument
* for caching routes it is interpreted as the `$strategy` argument

Do we want this? While it's not that great architecturally, it allows writing simple statements for simple registrations (e.g. `$scripts->register( 'foo', '/path/to/file.js' )` instead of `$scripts->register( 'foo', array( 'src' => '/path/to/file.js' ) )`). This is more typically WordPress, and I think this "trick" allows us to support simple method calls for the most basic use-case, while still allowing advanced functionality via the actual interface. Plus, at the same time we will never run into the issue where we add 5 positional arguments of which it is then important to remember the order (see e.g. `wp_enqueue_script()`) - an arguments array is easier to handle in that case.

wp-includes/class-wp-service-worker-scripts.php
```diff
@@ -0,0 +1,245 @@
...
+	 *
+	 * @since 0.2
+	 *
+	 * @param string $handle Handle of the script.
+	 * @param array  $args   {
+	 *     Additional script arguments.
+	 *
+	 *     @type string|callable $src  Required. URL to the source in the WordPress install, or a callback that
+	 *                                 returns the JS to include in the service worker.
+	 *     @type array           $deps An array of registered item handles this item depends on. Default empty array.
```
westonruter: The type here would be updated to `array|string|callable`, no?

felixarntz: The type for the parameter? Indeed, that would need to be adjusted, also for `WP_Service_Worker_Precaching_Routes::register()` and `WP_Service_Worker_Caching_Routes::register()` (for those it would be `array|string`). I did implement the logic, but the doc-block still has only `array` there at the moment.

PR #77, opened 2018-09-20
Update Workbox to v3.6.1 and explicitly disable navigation preload when not enabled

westonruter:
See https://github.com/GoogleChrome/workbox/issues/1648


westonruter:
I need to fix the unrelated failure of `Test_WP_HTTPS_Detection::test_schedule_cron`.

Issue #56, opened 2018-08-28
Display notice if HTTPS is not enabled, use 301 code in redirect


kienstra:
As a follow-up to #17:
* Display a notice if the `siteurl` and `home` (WordPress Address and Site Address) are HTTP, and HTTPS is not enabled (like if there's no certificate)
<img width="934" alt="display-notice" src="https://user-images.githubusercontent.com/4063887/44751240-3fd1ab80-aadd-11e8-9d05-db2d2bbdc7ff.png">

* Change the code in the redirect from `302` to `301`:

https://github.com/xwp/pwa-wp/blob/e1431eae92f75a09742c7cdd3dc438011f53fb9d/wp-includes/class-wp-https-ui.php#L358

westonruter:
This burn-in period sounds similar to what we were thinking of for HSTS. Maybe adding HSTS could be the final step in the burn-in period, with exponential backoff of the TTL. So maybe:

* 1 week of 302 redirects, followed by 301
* then 1 week of HSTS at 60 minute TTL,
* then 1 month of HSTS at 24 hour TTL
* then permanently setting HSTS to an optimal TTL (e.g. 1 month or longer)

kienstra:
**Redirect Code, HTTPS Notice**

Hi @westonruter and @igrigorik,
(a) Good ideas for phasing in a `301` redirect code. 


(b) Here's some possible copy for an admin notice if HTTPS is not possible, like if there's no certificate:

>HTTPS is not enabled on your site, possibly because there's no SSL certificate, or an invalid one. We strongly suggest enabling it for better security. [More details](https://make.wordpress.org/support/user-manual/web-publishing/https-for-wordpress/)

<img width="1333" alt="https-notice" src="https://user-images.githubusercontent.com/4063887/44808093-efb42100-ab90-11e8-8779-3a7f232d41fc.png">

kienstra:
**Link For 'More details'**

Hi @igrigorik,
>For the "more details" link, are there any other pages that we could link to in the codex, that focus on the user+site owner benefits of HTTPS instead of the technical guidance? I don't think that page does a good job of motivating the claim we're proposing in the notice.

Good question. I couldn't find a page on `wordpress.org` that better describes the benefits of HTTPS.

igrigorik:
@kienstra @westonruter could we link to the Web Fundamentals article then? Are there any best practices / requirements on links from WP dashboard?

kienstra:
[...]

PR #82, opened 2018-10-02
[WIP] Conditionally add HSTS header

### LINKED ISSUE #18: Add support for HTTP Strict Transport Security (HSTS)
As a site admin, I should have a simple way to enable HSTS on my site.

- [ ] AC1: Assuming HTTPS checks all clearly show I probably can enable the HSTS header, I should be presented with this option in the HTTPS dialogue in #17.
- [ ] AC2: If HSTS activation fails during a loopback request, warn the user and prevent them from activating this setting.

---

A best practice for HTTPS is to also use [HTTP Strict Transport Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) (HSTS):

> If a website accepts a connection through HTTP and redirects to HTTPS, visitors may initially communicate with the non-encrypted version of the site before being redirected, if, for example, the visitor types http://www.foo.com/ or even just foo.com. This creates an opportunity for a man-in-the-middle attack. The redirect could be exploited to direct visitors to a malicious site instead of the secure version of the original site.
> 
> The HTTP Strict Transport Security header informs the browser that it should never load a site using HTTP and should automatically convert all attempts to access the site using HTTP to HTTPS requests instead.

If we detect HTTPS support (#6) and the user opts-in to use it (#17) then in addition to redirecting from HTTP to HTTPS then we should also serve the HSTS header.

See also core ticket [#28520](https://core.trac.wordpress.org/ticket/28520).

kienstra:
* Begins the work of adding an HSTS header
* Follows @westonruter's [suggested](https://github.com/xwp/pwa-wp/issues/56#issuecomment-417024175) 'burn-in' period for HSTS 

@todos include:
* Maybe expiring HSTS if a certificate expires, probably by setting the `max-age` to 0:
>Should it be necessary to disable Strict Transport Security, setting the max-age to 0 (over a https connection) will immediately expire the Strict-Transport-Security header, allowing access via http. ([source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#How_the_browser_handles_it))

Closes #18.

### DIFF
--- a/tests/test-class-wp-https-detection.php
+++ b/tests/test-class-wp-https-detection.php
@@ -65,6 +65,7 @@ class Test_WP_HTTPS_Detection extends WP_UnitTestCase {
 	public function test_init() {
 		$this->assertEquals( 10, has_action( WP_HTTPS_Detection::CRON_HOOK, array( $this->instance, 'update_https_support_options' ) ) );
 		$this->assertEquals( PHP_INT_MAX, has_filter( 'cron_request', array( $this->instance, 'conditionally_prevent_sslverify' ) ) );
+		$this->assertEquals( 10, has_action( 'update_option_' . WP_HTTPS_UI::UPGRADE_HTTPS_OPTION, array( $this->instance, 'conditionally_reset_successful_https_check' ) ) );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


kienstra:
**Question About Closing PR**

Hi @westonruter,
Thanks, that sounds good to have HSTS only added programmatically.

What do you think about closing this PR, and letting developers add it on their own with something like:

```php
/**
 * Adds an HSTS header to the response.
 *
 * @param array $headers The headers to filter.
 */
add_filter( 'wp_headers', function( $headers ) {
	$headers['Strict-Transport-Security'] = 'max-age=3600'; // Or another max-age.
} );
```

Thanks, Weston!

postphotos:
@kienstra, I like @westonruter's solution and think it's a good suggestion. We should discuss this in tomorrow's meeting!

westonruter:
@kienstra yes, good suggestion. We can close this PR for now and instead opt to resolve #18 by means of documentation and adding the HSTS header via a filter as you suggested.

kienstra:
Thanks, @westonruter!

Issue #81, opened 2018-09-23
Use alternative to BroadcastChannel for sending server error to client 500 page


westonruter:
When a 500 error occurs, a 500 error template is served by the service worker. If there was an error message displayed on the response, the HTML is sent to the served 500 error page via `BroadcastChannel`:

https://github.com/xwp/pwa-wp/blob/2359ae5a890e199a00b12172297dac2a6af6c7f4/wp-includes/js/service-worker-error-response-handling.js#L11-L31

However, this is not supported by browsers other than Chrome and Firefox: https://caniuse.com/#feat=broadcastchannel

We'll need to find an alternative. (Also the approach taken should be validated.)

Aside: We also need to prevent serving the 500 template when WordPress is actually rendering a response with a `500` status code, per #74.

westonruter:
Actually, we don't need to send the error in a message at all. We can instead write the error message directly into the 500 error template response. 

This would eliminate the need for `wp_print_service_worker_error_details_script()` and most of:

```php
<?php if ( ! ( function_exists( 'is_amp_endpoint' ) && is_amp_endpoint() ) ) : ?>
	<details id="error-details" hidden>
		<summary><?php esc_html_e( 'More details', 'twentyseventeen-westonson' ); ?></summary>
		<iframe srcdoc=""></iframe>
		<script>
		function renderErrorDetails( data ) {
			if ( data.bodyText.trim().length ) {
				const details = document.getElementById( 'error-details' );
				details.querySelector( 'iframe' ).srcdoc = data.bodyText;
				details.hidden = false;
			}
		}
		</script>
		<?php wp_print_service_worker_error_details_script( 'renderErrorDetails' ); ?>
	</details>
<?php endif; ?>
```

Instead, there could be some token that is added to the template which is then replaced with the error when the service worker serves the response. There could be a token for encoding the error in an HTML attribute (for the `iframe@srcdoc`). Perhaps the token should be replaced with the entire populated `details` element, rather than just updating the `iframe`. We may not need to allow the markup to be modified to a large degree.

westonruter:
Closing in favor of #86.

PR #85, opened 2018-10-11
Add service worker streaming

westonruter:
Works best with AMP due to the fact that all scripts in AMP are asynchronous (https://github.com/Automattic/amp-wp/pull/1503). See changes in theme: https://github.com/westonruter/twentyseventeen-westonson/pull/2.

This pull request explores what it would look like to improve performance of navigating a site through the use of streaming. See @jeffposnick's article [Beyond SPA](https://developers.google.com/web/updates/2018/05/beyond-spa) for some key inspiration here. The goal is to speed up the perceived performance for navigating a site by showing a meaningful interface and loading indicator immediately. In short, this should reduce the first contentful paint time when navigating an AMP site on origin.

The proposed way that this would work is as follows:

0. The user opts in to streaming via theme support flag (`add_theme_support( 'service_worker_streaming' )`) and adds a `WP_Service_Worker_Navigation_Routing_Component::do_stream_boundary()` call to their theme's `header.php`. This function outputs a loading message/progressbar to be displayed while the body is loading.
1. A new query var called `wp_stream_fragment` is added which can take one of two values: `header` and `body` (for example, https://example.com/?wp_stream_fragment=header and https://example.com/?wp_stream_fragment=body). This argument can be added to any URL, and when `header` is supplied the AMP plugin returns the document up until the `do_stream_boundary()` call and when `body` is supplied the document after the `do_stream_boundary()` call is returned.
2. The `header` fragment is precached by the service worker. The `title` in this header fragment is filtered to say “Loading...”. This header fragment is also served with an inline `script` containing am `wpStreamCombine()` function which is called later in the body fragment.
3. When the service worker handles a navigation request, it will immediately serve the header fragment which will be displayed while the browser is fetching the body fragment. While this is happening the loading indicator is shown.
4. WordPress serves the body fragment with data extracted from the `head` children and `body` attributes which would have normally been included in an entire response. Output buffering is used to capture the `header` for that given request and it is loaded into a `DOMDocument`. A JSON representation of this data included in this response and into the `wpStreamCombine` function (defined in the header fragment) at the very beginning of the body fragment which reconciles the differences between the precached header fragment. In Jeff's example linked above, he used this approach for reconciling the `document.title`. 
5. If the request to obtain the body fragment results in a redirect, then the resultant URL is set in the user's browser via `location.replaceState()`.
6. The `wpStreamCombine` function deletes the element containing loading indicator.
7. If the body fragment request fails, then a precached offline page body fragment would be served instead. Eventually, the body fragment responses should use a stale-while-revalidate strategy once there is a mechanism to communicate an update back to the user (see https://github.com/ampproject/amphtml/issues/18615#issuecomment-428006471)

The logic for obtaining the elements in the `head` can be deferred for later so another plugin (like AMP) can obtain the elements after it has sanitized it: https://github.com/Automattic/amp-wp/pull/1503.

# Abbreviated example header fragment response

```html
<!DOCTYPE html>
<html lang="en-US" class="no-js" amp="">
	<head>
		<meta charset="UTF-8">
		...
	</head>
	<body class="home">
		...
		<div id="wp-stream-fragment-boundary">Loading...</div>
		<script id="wp-stream-combine-function">
			function wpStreamCombine( data ) { /* ... */ }
		</script>
```

# Abbreviated example body fragment response

```html
	<script id="wp-stream-combine-call">
	wpStreamCombine( {
		"head_nodes": [ /* ... */ ],
		"body_attributes": { /* ... */ }
	} );
	</script>
	<div id="primary" class="content-area">
	...
	</div>
</body>
</html>
```

# Demo

See child theme PR for Twenty Seventeen that implements support: https://github.com/westonruter/twentyseventeen-westonson/pull/2

Video: https://youtu.be/y5J69wLstUM

Live: https://streaming-westonruter.pantheonsite.io/

# Todo

- [x] Given that the header is dynamically-generated by WordPress it becomes very easy for the precached header fragment to become stale. Theme authors may need to add the boundary before the nav menu is output to help mitigate this. Otherwise, a filter like `wp_offline_error_precache_entry` will be needed with a `revision` that includes all variability that will be included in the header.
- [x] When the body request returns in a redirect, we'll need to cause the browser to redirect via a `location.replace()` call.


wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -8,7 +8,16 @@ wp.serviceWorker.routing.registerRoute( new wp.serviceWorker.routing.NavigationR
 
 		const handleResponse = ( response ) => {
 			if ( response.status < 500 ) {
-				return response;
+				if ( response.redirected ) {
+					const redirectedUrl = new URL( response.url );
+					redirectedUrl.searchParams.delete( STREAM_HEADER_FRAGMENT_QUERY_VAR );
+					const script = `<script>history.replaceState( {}, '', ${ JSON.stringify( redirectedUrl.toString() ) } );</script>`;
+					return response.text().then( ( body ) => {
+						return new Response( script + body );
```
westonruter: @jeffposnick here's how I ended up dealing with the challenge of handling a streamed body that redirects.

The “Try Redirect” nav menu item points to an old post URL that redirects, so you can see the URL go from "old-post" to "new post".

westonruter: https://streaming-westonruter.pantheonsite.io/

jeffposnick: 😍

wp-includes/components/class-wp-service-worker-navigation-routing-component.php
```diff
@@ -0,0 +1,461 @@
...
+		$is_header_buffered = (
+			false !== has_action( 'template_redirect', 'WP_Service_Worker_Navigation_Routing_Component::start_output_buffering_stream_fragment' )
+			&&
+			ob_get_level() > 0
+		);
+		if ( $is_header_buffered ) {
+			$header_html       = ob_get_clean();
+			$libxml_use_errors = libxml_use_internal_errors( true );
+			$header_html       = preg_replace( '#<noscript.+?</noscript>#s', '', $header_html ); // Some libxml versions croak at noscript in head.
+			$dom               = new DOMDocument( $header_html );
```
westonruter: Remove `$header_html` as an arg. Not correct.

wp-includes/components/class-wp-service-worker-navigation-routing-component.php
```diff
@@ -0,0 +1,461 @@
...
+			 *     @type string $url      URL to streaming header fragment.
+			 *     @type string $revision Revision for the entry. Care must be taken to keep this updated based on the content that is output before the stream boundary.
+			 * }
+			 */
+			$streaming_header_precache_entry = apply_filters( 'wp_streaming_header_precache_entry', $streaming_header_precache_entry );
+
+			if ( $streaming_header_precache_entry ) {
+				$scripts->precaching_routes()->register( $streaming_header_precache_entry['url'], isset( $streaming_header_precache_entry['revision'] ) ? $streaming_header_precache_entry['revision'] : null );
+
+				add_filter( 'wp_service_worker_navigation_preload', '__return_false' ); // Navigation preload and streaming don't mix!
```
westonruter: Verify this is needed.

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -0,0 +1,102 @@
...
+				}
+			} catch ( error ) {
+				return sendOfflineResponse();
+			}
+		}
+
+		const canStreamResponse = () => {
+			const url = new URL( event.request.url );
+			return ! (
+				/\.php$/.test( url.pathname ) ||
```
westonruter: Instead of doing this, why not add `*.php` files to the blacklist patterns?

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -0,0 +1,102 @@
...
+
+		if ( isStreamingResponses && canStreamResponse() ) {
+			const streamHeaderFragmentURL = STREAM_HEADER_FRAGMENT_URL;
+			const precacheStrategy = wp.serviceWorker.strategies.cacheFirst({
+				cacheName: wp.serviceWorker.core.cacheNames.precache,
+			});
+
+			const url = new URL( event.request.url );
+			url.searchParams.append( STREAM_HEADER_FRAGMENT_QUERY_VAR, 'body' );
+			const request = new Request( url.toString(), {...event.request} );
```
westonruter: Verify that `{...event.request}` is working.

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -0,0 +1,102 @@
...
+
+			const stream = wp.serviceWorker.streams.concatenateToResponse([
+				precacheStrategy.makeRequest({ request: streamHeaderFragmentURL }), // @todo This should be able to vary based on the request.url. No: just don't allow in paired mode.
+				fetch( request )
+					.then( handleResponse )
+					.catch( sendOfflineResponse ),
+			]);
+
+			// @todo Handle error case.
```
westonruter: This was done

wp-includes/js/service-worker-stream-combiner.js
```diff
@@ -0,0 +1,139 @@
...
+ */
+function wpStreamCombine( data ) { /* eslint-disable-line no-unused-vars */
+	const processedHeadNodeData = new WeakSet();
+	const processedHeadElements = new WeakSet();
+
+	/**
+	 * Determine if a given element matches the nodeData coming from the body fragment.
+	 *
+	 * Returns true if the element name is the same and its attributes are equal.
+	 *
```
westonruter: todo: Add jsdoc for params.

wp-includes/js/service-worker-stream-combiner.js
```diff
@@ -0,0 +1,139 @@
...
+	// Populate body attributes.
+	for ( const key in data.body_attributes ) {
+		document.body.setAttribute( key, data.body_attributes[ key ] );
+	}
+
+	/* Purge all traces of the stream combination logic to ensure the AMP validator doesn't complain at runtime. */
+	const removedElements = [
+		STREAM_COMBINE_INVOKE_SCRIPT_ID,
+		STREAM_COMBINE_DEFINE_SCRIPT_ID,
+		STREAM_FRAGMENT_BOUNDARY_ELEMENT_ID
```
westonruter: Also include the `history.replaceState()` script?

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -0,0 +1,102 @@
...
+					redirectedUrl.searchParams.delete( STREAM_HEADER_FRAGMENT_QUERY_VAR );
+					const script = `<script>history.replaceState( {}, '', ${ JSON.stringify( redirectedUrl.toString() ) } );</script>`;
+					return response.text().then( ( body ) => {
+						return new Response( script + body );
+					} );
+				} else {
+					return response;
+				}
+			}
+			const channel = new BroadcastChannel( 'wordpress-server-errors' );
```
amedina: Not a blocker at this point, but it would be needed to define the scenario for browsers that do not support BroadcastChannel (https://caniuse.com/#search=broadcastchannel)

westonruter: Yes, this is captured in #81.

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -0,0 +1,102 @@
...
+				const response = await event.preloadResponse;
+				if ( response ) {
+					return handleResponse( response );
+				}
+			} catch ( error ) {
+				return sendOfflineResponse();
+			}
+		}
+
+		const canStreamResponse = () => {
```
amedina: Add a brief comment here on the need to avoid PHP files. As discussed, this could be moved to the `black listing` logic.

wp-includes/js/service-worker-stream-combiner.js
```diff
@@ -0,0 +1,139 @@
+/*global STREAM_COMBINE_INVOKE_SCRIPT_ID, STREAM_COMBINE_DEFINE_SCRIPT_ID, STREAM_FRAGMENT_BOUNDARY_ELEMENT_ID */
+
+"use strict";
+/* This JS file will be added as an inline script in a stream header fragment response. */
+/* This file currently uses JS features which are compatible with Chrome 40 (Googlebot). */
+
+/**
+ * Apply the stream body data to the stream header.
+ *
+ * @param {Array}  data Data.
```
amedina: Perhaps expand the expected content of the `data` argument.

PR #92, opened 2018-10-18
Offline commenting

### LINKED ISSUE #87: Implement Offline Commenting
**As as WordPress developer, I would like to be able to allow site visitors and admin users add comments to posts even when offline**

- [x] _AC 1:_ **Build comment offline sync functionality into PWA plugin.** When viewing a page or post with commenting enabled, a user will be able to make a comment and have that comment saved and posted once the user is back online
- [x] _AC 2:_ **The user interface for offline commenting will be easy to use and include appropriate messaging.** By default the plugin should show an offline page where submitting a comment when offline, with a message like “you are offline, but we will post your comment will be posted when your device comes back online.” This should integrate with the offline template to show custom messages similar to how the `500` page can show custom messages.   (See related issue in the twentyseventeen-weston repo, [Issue 4 Optimize the design of the Offline/500 page to be more engaging for the user](https://github.com/westonruter/twentyseventeen-westonson/issues/4))

- [x] _AC 3:_ A theme must be able to override the offline template and messaging for commenting failures.
- [ ] _AC 4:_ Site on Using our AMP and PWA plugins, enable this offline commenting support via a theme declaration.
- [ ] _AC 5:_ Intercept the amp-form action-xhr request (to submit the form) and when failed return a custom success message to be shown below the form

**Offline Routing and messaging**
- [ ] _AC 6:_  If AMP is enabled, The amp-live-list should continue to update in real time, showing the new comment when background sync happens or if the comment was submitted successfully right away. The error message should probably appear in the context of the posted comment.
- [x] _AC 7:_ If AMP is not enabled, route the user to the site's offline page and allow a theme developer to serve the offline message from AC2/AC3. Make sure the the ability for service worker to populate content of the offline template is added.

For reference, here was the original <amp-live-list> PR from the AMP plugin:
https://github.com/Automattic/amp-wp/issues/1028

miina:
Fixes #87.

Adds default Offline Commenting support:
- In case of submitting a comment while offline the default offline template with a message is displayed letting the user know the comment will be posted offline later;
- Adds filter for being able to customize the default messages;
- Implements background sync with queue `wpPendingComments`;

Temporarily adds Offline Commenting support for AMP Live List Commenting:
- In case of being offline displays the relevant message under the commenting form;
- Implements background sync with queue `amp-wpPendingComments`;
- In case of successful sync and comment being submitted updates the live list;

Related issue to be implemented later: https://github.com/xwp/pwa-wp/issues/101


wp-includes/components/class-wp-service-worker-offline-commenting-component.php
```diff
@@ -0,0 +1,77 @@
...
+		$scripts->register(
+			'wp-offline-commenting',
+			array(
+				'src'  => array( $this, 'get_script' ),
+				'deps' => array( 'wp-base-config', 'wp-error-response' ),
+			)
+		);
+
+		if ( ! is_admin() ) {
+			$offline_template_url = add_query_arg( 'wp_error_template', 'offline', home_url( '/' ) );
```
miina: @todo These originally have filtering

miina: Probably there should be one method for getting the `offline_template_url` instead of duplicating from the error response handling logic.

wp-includes/theme-compat/error.php
```diff
@@ -16,6 +16,7 @@
 <main>
 	<h1><?php esc_html_e( 'Oops! It looks like you&#8217;re offline.', 'pwa' ); ?></h1>
 	<p><?php esc_html_e( 'Please check your internet connection, and try again.', 'pwa' ); ?></p>
+	<p><!-- WP_OFFLINE_COMMENT --></p>
```
miina: @todo This is a placeholder for testing -- the whole offline message part should be replaced with a template tag instead of just adding a new one for offline comment.

wp-includes/theme-compat/offline.php
```diff
@@ -16,6 +16,7 @@
 <main>
 	<h1><?php esc_html_e( 'Oops! It looks like you&#8217;re offline.', 'pwa' ); ?></h1>
 	<p><?php esc_html_e( 'Please check your internet connection, and try again.', 'pwa' ); ?></p>
+	<p><!-- WP_OFFLINE_COMMENT --></p>
```
westonruter: Instead of `WP_OFFLINE_COMMENT` or `WP_OFFLINE_ERROR_MESSAGE` as I have above, we could have a common string that is used for both offline and 500 error scenarios (per #81). So there could be just `<!--WP_SERVICE_WORKER_ERROR_MESSAGE-->` which is used in all templates regardless.

westonruter:
Dependent PR has been merged, so `master` can be merged back into this branch now.

wp-includes/components/class-wp-service-worker-navigation-routing-component.php
```diff
@@ -424,6 +424,7 @@ public function serve( WP_Service_Worker_Scripts $scripts ) {
 			'BLACKLIST_PATTERNS'               => wp_service_worker_json_encode( $this->get_blacklist_patterns() ),
 			'SHOULD_STREAM_RESPONSE'           => wp_service_worker_json_encode( $should_stream_response ),
 			'STREAM_HEADER_FRAGMENT_QUERY_VAR' => wp_service_worker_json_encode( self::STREAM_FRAGMENT_QUERY_VAR ),
+			'ERROR_MESSAGES'                   => wp_service_worker_json_encode( wp_service_worker_get_error_messages() ),
```
miina: @westonruter: Continued with the PR and addressed your comments, too.

Question about `wp_service_worker_get_error_messages()` -- do you think this method is necessary as a global method or we could just filter it here within Navigation Component?

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -1,121 +1,140 @@
-/* global console, ERROR_OFFLINE_URL, ERROR_500_URL, SHOULD_STREAM_RESPONSE, STREAM_HEADER_FRAGMENT_URL, ERROR_500_BODY_FRAGMENT_URL, ERROR_OFFLINE_BODY_FRAGMENT_URL, STREAM_HEADER_FRAGMENT_QUERY_VAR, BLACKLIST_PATTERNS */
+/* global console, ERROR_OFFLINE_URL, ERROR_500_URL, SHOULD_STREAM_RESPONSE, STREAM_HEADER_FRAGMENT_URL, ERROR_500_BODY_FRAGMENT_URL, ERROR_OFFLINE_BODY_FRAGMENT_URL, STREAM_HEADER_FRAGMENT_QUERY_VAR, BLACKLIST_PATTERNS, ERROR_MESSAGES */
+{
```
miina: Wrapped all this into curly brackets, are there any concerns with it?

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -1,121 +1,140 @@
...
+				// Close the channel if client did not request the message within 30 seconds.
+				setTimeout( () => {
+					channel.close();
+				}, 30 * 1000 );
+
+				return caches.match( canStreamResponse() ? ERROR_500_BODY_FRAGMENT_URL : ERROR_500_URL );
+			};
+
+			const sendOfflineResponse = () => {
+				if ( canStreamResponse() ) {
```
miina: Here starts the actual main change for the file, most of the other changes are related to wrapping it into curly brackets.

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -1,121 +1,140 @@
...
+					return caches.match( ERROR_OFFLINE_BODY_FRAGMENT_URL );
+				}
+
+				return caches.match( ERROR_OFFLINE_URL ).then( function( response ) {
```
miina: This mostly duplicates the handling within `service-worker-offline-commenting.js`, except for the `errorMessages.default` vs `errorMessage.comment`. Wondering if we should create a global method instead?

wp-includes/service-workers.php
```diff
@@ -334,3 +334,25 @@ function wp_service_worker_fragment_redirect_old_slug_to_new_url( $link ) {
 	}
 	return $link;
 }
+
+/**
+ * Get service worker error messages.
+ *
+ * @return array Array of error messages: default, comment.
+ */
+function wp_service_worker_get_error_messages() {
```
miina: As mentioned above too, I'm wondering if this method is necessary as a global method.

wp-includes/js/service-worker-offline-commenting.js
```diff
@@ -0,0 +1,55 @@
...
+							headers: request.headers,
+							mode: 'same-origin',
+							credentials: request.credentials,
+							referrer: request.referrer,
+							redirect: 'manual',
+							body: body
+						} );
+
+						// Add request to queue.
+						queue.addRequest( req );
```
miina: Per GoogleChrome/workbox#1710 `addRequest` is going to be replaced in v4.

westonruter: We could update to v4 now if that is helpful.

wp-includes/js/service-worker-navigation-routing.js
westonruter: That's a good idea.

wp-includes/components/class-wp-service-worker-navigation-routing-component.php
westonruter: Making it a method of the navigation routing component would probably be best.

wp-includes/service-workers.php
```diff
@@ -334,3 +334,25 @@ function wp_service_worker_fragment_redirect_old_slug_to_new_url( $link ) {
...
+			'comment' => __( 'Your comment will be submitted once you are back online!', 'pwa' ),
+		)
+	);
+}
+
+/**
+ * Display service worker error message template tag.
+ */
+function wp_service_worker_error_message_placeholder() {
+	echo wp_kses_post( '<p><!--WP_SERVICE_WORKER_ERROR_MESSAGE--></p>' );
```
westonruter: I don't think the `wp_kses_post()` is necessary here. Also, should the paragraphs be included?

miina: Not sure about the paragraphs -- I was thinking at first that maybe not but then the developer would always have to add the markup instead of just overriding the text message. Without markup it would look like this:
screen shot 2018-10-23 at 11 15 47 am

So basically I was thinking that maybe for the default template we should add paragraph and try to avoid "breaking" how the default offline template looks like.

However, I'm fine removing it.

westonruter: OK, for now let's just do:

```php
echo '<p><!--WP_SERVICE_WORKER_ERROR_MESSAGE--></p>' ;
```

(No need for `wp_kses_post()`.)

I think I was thinking of a case where there is no message. But I suppose there would always be one.

wp-includes/components/class-wp-service-worker-navigation-routing-component.php
```diff
@@ -473,8 +474,9 @@ public function get_priority() {
 	 * @return string Script.
 	 */
 	public function get_script() {
-		$script = file_get_contents( PWA_PLUGIN_DIR . '/wp-includes/js/service-worker-navigation-routing.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
-		$script = preg_replace( '#/\*\s*global.+?\*/#', '', $script );
+		$script  = file_get_contents( PWA_PLUGIN_DIR . '/wp-includes/js/service-worker-offline-commenting.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
```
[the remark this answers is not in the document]
miina: Originally it was implemented separately but then moved to here since it seemed to make sense due to sharing some logic (getting the URL-s to replace for example, `$offline_error_precache_entry`). Makes sense though that the default handler should probably have a late priority! Will move the commenting separately.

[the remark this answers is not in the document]
miina: Also getting the error message would be shared then so in case of moving it separately we could maybe still have `wp_service_worker_get_error_messages` as a global method.

wp-includes/js/service-worker-offline-commenting.js
```diff
@@ -0,0 +1,55 @@
...
+{
+	const queue = new wp.serviceWorker.backgroundSync.Queue( 'wpPendingComments' );
+	const errorMessages = ERROR_MESSAGES;
+
+	const commentHandler = ( { event } ) => {
+
+		const clone = event.request.clone();
+		return fetch( event.request )
+			.then( ( response ) => {
+				return response;
```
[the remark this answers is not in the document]
miina: Yes. These two scripts (navigation and commenting) would have duplicate code this way, should we move the `handleRequest` method to be a global method instead?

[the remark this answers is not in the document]
westonruter: How about adding a new script which is marked as a dependency for both offline commenting and navigation routing which defines these two functions for handling the response and offline response, and then we can re-use these functions in both places.

Hurray for script dependencies!

Perhaps these functions should be defined off of `wp.serviceWorker`. Nevertheless, beware:

https://github.com/xwp/pwa-wp/blob/c3b06fe6f7cfc78538c2b256448a41e4a050602d/wp-includes/js/service-worker.js#L134-L141

wp-includes/js/service-worker-offline-commenting.js
```diff
@@ -0,0 +1,82 @@
...
+
+		const clone = event.request.clone();
+		return fetch( event.request )
+			.then( ( response ) => {
+				if ( response.status < 500 ) {
+					return response;
+				}
+
+				// @todo Replace depending on BroadcastChannel.
+				const channel = new BroadcastChannel( 'wordpress-server-errors' );
```
miina: I added currently the logic here without creating global methods (yet) to see what's needed exactly. Also removed the streaming from here since it didn't seem to be necessary in case of `wp-comments-post`.

wp-includes/components/class-wp-service-worker-offline-commenting-component.php
```diff
@@ -0,0 +1,153 @@
...
+			$scripts->precaching_routes()->register( $offline_error_precache_entry['url'], isset( $offline_error_precache_entry['revision'] ) ? $offline_error_precache_entry['revision'] : null );
+		}
+		if ( $server_error_precache_entry ) {
+			$scripts->precaching_routes()->register( $server_error_precache_entry['url'], isset( $server_error_precache_entry['revision'] ) ? $server_error_precache_entry['revision'] : null );
+		}
+
+		$this->replacements = array(
+			'ERROR_OFFLINE_URL' => wp_service_worker_json_encode( isset( $offline_error_precache_entry['url'] ) ? $offline_error_precache_entry['url'] : null ),
+			'ERROR_500_URL'     => wp_service_worker_json_encode( isset( $server_error_precache_entry['url'] ) ? $server_error_precache_entry['url'] : null ),
+			'ERROR_MESSAGES'    => wp_service_worker_json_encode( wp_service_worker_get_error_messages() ),
```
miina: @westonruter Created a separate component -- above you can see that getting these 3 URLs is just a duplication of Routing Component. Perhaps we should move these out of the components to avoid duplication.

wp-includes/js/service-worker-offline-commenting.js
miina: If we're going to update it anyway then probably would be good to do it now.

wp-includes/js/service-worker-offline-commenting.js
westonruter: That's right. No need to do streaming in handling `wp-comments-post.php` requests. It should be redirecting, or serving the offline page in its entirety.

amp/amp-service-worker-offline-commenting.js
```diff
@@ -0,0 +1,44 @@
...
+							redirect: 'manual',
+							body: body
+						} );
+
+						// Add request to queue. @todo Replace when upgrading to Workbox v4!
+						queue.addRequest( req );
+					}
+				);
+
+				// @todo That's not actually working yet.
```
westonruter: I think you have to move this into the `catch` function, like so:

```js
const bodyPromise = clone.blob();
bodyPromise.then(
	function( body ) {
		const request = event.request;
		const req = new Request( request.url, {
			method: request.method,
			headers: request.headers,
			mode: 'same-origin',
			credentials: request.credentials,
			referrer: request.referrer,
			redirect: 'manual',
			body: body
		} );

		// Add request to queue. @todo Replace when upgrading to Workbox v4!
		queue.addRequest( req );

		const body = JSON.stringify( { 'error': errorMessages.comment } );
		return new Response( body, {} );
	}
);
```

You only want it to be sent back in the case of an error (`catch`) scenario. Also, you should wait to send it until the `bodyPromise` resolves, as otherwise I believe it the callback can be killed.

amp/amp-service-worker-offline-commenting.js
```diff
@@ -0,0 +1,44 @@
...
+	const commentHandler = ( { event } ) => {
+		const clone = event.request.clone();
+		return fetch( event.request )
+			.then( ( response ) => {
+				// @todo Make sure that 409 etc. error still work as expected.
+				return response;
+			} )
+			.catch( () => {
+				const bodyPromise = clone.blob();
+				bodyPromise.then(
```
westonruter: Also this I believe should be:

```js
return bodyPromise.then(
```

westonruter:
@miina I just realized that it would be important to _not_ unregister the normal comment request handler (as with #94) because on a paired mode site, comments could be submitted outside of AMP as well. So both routes need to remain registered. Perhaps then this shows that all of the custom router logic should be actually eliminated (in a separate PR) since there is a real use case for multiple routes to match:

https://github.com/xwp/pwa-wp/blob/c3b06fe6f7cfc78538c2b256448a41e4a050602d/wp-includes/js/service-worker.js#L15-L141

miina:
@westonruter In case of Offline Commenting we're actually not going to need to unregister the default commenting since at least according to testing then these two registered routes are not conflicting and also not causing a warning (one is strictly matching `/wp-comments-post.php` and the other matching the route with specific parameters.

It still seems to me that it would be good to display a warning in case of conflicting routes since if there are for example two scripts that address exactly the same route, for example the default commenting (`/wp-comments-post.php`), and it's not working then as a developer it would be good to get the information that it's due to conflicts. The custom router is not doing anything else except for just displaying a warning in develop mode.

Thoughts?

westonruter:
> (one is strictly matching `/wp-comments-post.php` and the other matching the route with specific parameters.

Ok, so when you register a route with a string`'/wp-comments-post.php'` then Workbox will not allow query params?

wp-includes/js/service-worker-offline-commenting.js
```diff
@@ -0,0 +1,82 @@
...
+						const body = text.replace( /<!--WP_SERVICE_WORKER_ERROR_MESSAGE-->/, errorMessages.comment );
+
+						return new Response( body, init );
+					} );
+				} );
+			} );
+	};
+
+	wp.serviceWorker.routing.registerRoute(
+		'/wp-comments-post.php',
```
westonruter: We'll need to make sure this works on subdirectory installs, where `/wp-comments-post.php` does not exist at the root.

miina:
> Ok, so when you register a route with a string'/wp-comments-post.php' then Workbox will not allow query params?

According to testing then no, it does not.

wp-includes/js/service-worker-offline-commenting.js
miina: Good point, we can replace this with regex instead.

amp/amp-service-worker-offline-commenting.js
```diff
@@ -0,0 +1,51 @@
...
+			} )
+			.catch( () => {
+				const bodyPromise = clone.blob();
+				return bodyPromise.then(
+					function( body ) {
+						const request = event.request;
+						const req = new Request( request.url, {
+							method: request.method,
+							headers: request.headers,
+							mode: 'same-origin',
```
westonruter: I think that `mode` will need to be set to `cors` since when submitting a comment from the AMP Cache it will need to connect to the origin.

amp/amp-service-worker-offline-commenting.js
```diff
@@ -0,0 +1,49 @@
...
+						method: event.request.method,
+						headers: event.request.headers,
+						mode: event.request.mode,
+						credentials: event.request.credentials,
+						referrer: event.request.referrer,
+						redirect: 'follow',
+						body: body
+					} );
+
+					// Add request to queue. @todo Replace when upgrading to Workbox v4!
```
westonruter: This todo can now be _todone_ due to https://github.com/xwp/pwa-wp/pull/98

amp/amp-service-worker-offline-commenting.js
```diff
@@ -0,0 +1,49 @@
+/* global ERROR_MESSAGES, SITE_URL */
+{
+	const queue = new wp.serviceWorker.backgroundSync.Queue( 'amp-wpPendingComments' );
+	const errorMessages = ERROR_MESSAGES;
+
+	const commentHandler = ( { event } ) => {
+		const clonedRequest = event.request.clone();
+		return fetch( event.request )
+			.then( ( response ) => {
+				return response;
```
westonruter: Is there any need for this `then` handler now? Should it not just do:

```js
fetch( event.request ).catch( () => {
```

In reality a success `response` should always be a 302 redirect, correct? That is, unless AMP live list commenting is being used in which case it would be a 200 or 202 response. 

In particular, there is the [`handleResponse`](https://github.com/xwp/pwa-wp/blob/e63139548142aa261584902dcd22129457121658/wp-includes/js/service-worker-navigation-routing.js#L16-L57) function. 

Given that the 302 response is irrelevant for streaming or app shell, we don't need any of that logic. But some 500 error handling would perhaps be nice, but not essential for now. So I think we can just eliminate the `then`.

amp/amp-service-worker-offline-commenting.js
miina: Looks like we'll have to wait until [the related PR]( GoogleChrome/workbox#1710 ) gets merged, too.

miina:
Merged the Offline Commenting Component to Navigation Routing Component within 544a747.

PR #98, opened 2018-10-27
Upgrade Workbox to v4.0.0-alpha.0

westonruter:
Extends #97.

Updating to v4.0.0-alpha.0 is good for testing the release, and it is also needed for offline commenting (#87/#92).

This PR also eliminates storing Workbox in a version-specific directory (as output by `workbox copyLibraries`). This will make diffing between Workbox versions more sane.

This needs to be fixed before it can be merged:

image


westonruter:
In regards to the [changes needed](https://github.com/xwp/pwa-wp/pull/98/files/49b76b463c4d5c03051f86ec58e0f08097eb8b48..f2f745a26068ad482270b23975752f4e5091cdfc) to update `WPRouter`, @jeffposnick notes that the back-compat breakage between Workbox v3 and v4 was intentionally done in https://github.com/GoogleChrome/workbox/pull/1682

He also noted that overriding the default router still requires a bit of work to get it to behave as we want here. Overriding the default router in Workbox (v3) was first proposed in https://github.com/xwp/pwa-wp/issues/5#issuecomment-403082892

This change will be documented as part of the release notes in v4.

PR #100, opened 2018-10-30
Add an entry in the readme files explaining how to add an HSTS header

### LINKED ISSUE #18: Add support for HTTP Strict Transport Security (HSTS)

kienstra:
**Request For Review**

Hi @westonruter,
Following up on our conversation in https://github.com/xwp/pwa-wp/pull/82#issuecomment-430828370, this adds an entry in the `readme` files about how to add an `HSTS` header.

I'm not sure if this what you had in mind for documentation, though.

Thanks, Weston!

Closes #18.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -164,9 +164,11 @@ You can optionally add an [HSTS header](https://developer.mozilla.org/en-US/docs
  * Adds an HSTS header to the response.
  *
  * @param array $headers The headers to filter.
+ * @return array $headers The filtered headers.
  */
 add_filter( 'wp_headers', function( $headers ) {
 	$headers['Strict-Transport-Security'] = 'max-age=3600'; // Or another max-age.
+	return $headers;
 } );
 </pre>
 
--- a/readme.txt
+++ b/readme.txt
@@ -157,6 +157,25 @@ HTTPS is a prerequisite for progressive web apps. A service worker is only able 
 
 At the moment the plugin provides an API to detection of whether a site supports HTTPS. Building on that it's intended that this can then be used to present a user with an opt-in to switch over to HTTPS, which will also then need to include support for rewriting URLs from HTTP to HTTPS. See [labeled GitHub issues](https://github.com/xwp/pwa-wp/issues?q=label%3Ahttps) and see WordPress core tracking ticket [#28521](https://core.trac.wordpress.org/ticket/28521).
 
+You can optionally add an [HSTS header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) (HTTP `Strict-Transport-Security`). This indicates to the browser to only load the site with HTTPS, not HTTP.
+
+<pre lang=php>
+/**
+ * Adds an HSTS header to the response.
+ *
+ * @param array $headers The headers to filter.
+ */
+add_filter( 'wp_headers', function( $headers ) {
+	$headers['Strict-Transport-Security'] = 'max-age=3600'; // Or another max-age.
+} );
+</pre>
+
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -157,6 +157,25 @@ HTTPS is a prerequisite for progressive web apps. A service worker is only able
...
+You can optionally add an [HSTS header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) (HTTP `Strict-Transport-Security`). This indicates to the browser to only load the site with HTTPS, not HTTP.
+
+<pre lang=php>
+/**
+ * Adds an HSTS header to the response.
+ *
+ * @param array $headers The headers to filter.
+ */
+add_filter( 'wp_headers', function( $headers ) {
+	$headers['Strict-Transport-Security'] = 'max-age=3600'; // Or another max-age.
```
westonruter: This needs to return the `$headers` array.

readme.txt
```diff
@@ -157,6 +157,25 @@ HTTPS is a prerequisite for progressive web apps. A service worker is only able
 
 At the moment the plugin provides an API to detection of whether a site supports HTTPS. Building on that it's intended that this can then be used to present a user with an opt-in to switch over to HTTPS, which will also then need to include support for rewriting URLs from HTTP to HTTPS. See [labeled GitHub issues](https://github.com/xwp/pwa-wp/issues?q=label%3Ahttps) and see WordPress core tracking ticket [#28521](https://core.trac.wordpress.org/ticket/28521).
 
+You can optionally add an [HSTS header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) (HTTP `Strict-Transport-Security`). This indicates to the browser to only load the site with HTTPS, not HTTP.
+
+<pre lang=php>
+/**
+ * Adds an HSTS header to the response.
+ *
+ * @param array $headers The headers to filter.
```
westonruter: Add a `@return` tag.

readme.txt
kienstra: That's a big omission 😄 
https://github.com/xwp/pwa-wp/pull/100/commits/dd49a54cf4009a2b77a74809e4a3c349d57223ca adds the `return` statemetn.

readme.txt
kienstra: Thanks, https://github.com/xwp/pwa-wp/pull/100/commits/dd49a54cf4009a2b77a74809e4a3c349d57223ca adds the `@return` tag.

PR #102, opened 2018-10-31
Use newly-public routes property on Router class instead of private one

westonruter:
Regression introduced in #98.

This will fix this error being raised when using production build of Workbox:

image

This error is caused by the production version of Workbox [munging](https://github.com/GoogleChrome/workbox/blob/b146d8d3c2db89fa51afc322049b9795894bb4f2/gulp-tasks/utils/rollup-helper.js#L46-L54) the `_routes` property of the `Router` class:

https://github.com/xwp/pwa-wp/blob/1efb6a9243f2e1a1eaab264d8f1fe82e3d029831/wp-includes/js/service-worker.js#L25

Workbox v4 will make this private `_routes` into public `routes`. When this happens upstream, this PR needs to be updated with the new build.


westonruter:
Depends on https://github.com/GoogleChrome/workbox/pull/1747

PR #103, opened 2018-10-31
Add developer documentation for 0.2

miina:
Adds general documentation about caching, precaching, offline templates, offline commenting; also lists out all actions and filters added by the feature plugin.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -154,49 +154,43 @@ See [labeled GitHub issues](https://github.com/xwp/pwa-wp/issues?q=label%3Aservi
 Service Workers in the feature plugin are using [Workbox](https://developers.google.com/web/tools/workbox/) to power a higher-level PHP abstraction for themes and plugins to indicate the routes and the caching strategies in a declarative way. Since only one handler can be used per one route then conflicts are also detected and reported in console when using debug mode.
 
 The API abstraction allows registering routes for caching and urls for precaching using the following two functions:
-
-1. `wp_register_service_worker_caching_route`: accepts the following two parameters:
-
+1. `wp_register_service_worker_caching_route()`: accepts the following two parameters:
 * `$route`: Route regular expression, without delimiters.
 * `$args`: An array of additional route arguments as `$key => $value` pairs:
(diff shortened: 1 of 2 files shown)


miina:
@westonruter I added a brief description about the features added meanwhile since the last update to Readme file and also listed out all actions and filters.

Let me know if there seems to be anything missing or if you think it would be good add some additional examples.

westonruter:
> However, (not right now but in the future) what do you think of keeping `readme.md` and `readme.txt` as separate files where `readme.md` would contain developer focused information and where `readme.txt` would also include screenshots for the plugins page in WP repo and more relevant information for the plugin users who won't necessarily be developers only? It feels like these two files have slightly different purposes.

The thing is that if there is a `readme.md` then GitHub will use it to display rather than the `readme.txt`, whereas WordPress.org only recognizes the latter. So by having it all in `readme.txt` then Markdown generator in dev-lib will automatically keep them in sync, with GitHub getting a nicely-formatted readme (in Markdown) and WordPress.org getting its required readme format as well.

miina:
> The thing is that if there is a readme.md then GitHub will use it to display rather than the readme.txt, whereas WordPress.org only recognizes the latter. So by having it all in readme.txt then Markdown generator in dev-lib will automatically keep them in sync, with GitHub getting a nicely-formatted readme (in Markdown) and WordPress.org getting its required readme format as well.


What I was thinking was that maybe `readme.md` could be more developer-focused (since it's displayed in Github) and `readme.txt` would be more WP.org-focused with the screenshots etc. and not necessarily only developer-focused but also considering non-technical plugin users. That perhaps they shouldn't be exact match and automatically synced?

westonruter:
@miina I tried running `npm run generate-readme` and it seems that all of the Offline Commenting 
section isn't moved over to `readme.txt` yet. I did so in 9fb310b. Let me know if I missed anything.

> What I was thinking was that maybe `readme.md` could be more developer-focused (since it's displayed in Github) and `readme.txt` would be more WP.org-focused with the screenshots etc. and not necessarily only developer-focused but also considering non-technical plugin users. That perhaps they shouldn't be exact match and automatically synced?

I think since this is a feature plugin that is intended for developers that it is fine and even preferred to have the developer information displayed on WordPress.org.

miina:
> I think since this is a feature plugin that is intended for developers that it is fine and even preferred to have the developer information displayed on WordPress.org.

Okay, makes sense!

Thanks for moving the Offline Commenting section over, too.

PR #104, opened 2018-10-31
Replace using BroadcastChannel by injecting error message into error templates

### LINKED ISSUE #86: Reconsider use of JavaScript to inject 500 error details
A `500.php` template currently needs to include the following to display any 500 error that was returned from the server:

```php
<details id="error-details" hidden>
	<summary><?php esc_html_e( 'More details', 'twentyseventeen-westonson' ); ?></summary>
	<iframe srcdoc=""></iframe>
	<script>
	function renderErrorDetails( data ) {
		if ( data.bodyText.trim().length ) {
			const details = document.getElementById( 'error-details' );
			details.querySelector( 'iframe' ).srcdoc = data.bodyText;
			details.hidden = false;
		}
	}
	</script>
	<?php wp_print_service_worker_error_details_script( 'renderErrorDetails' ); ?>
</details>
```

However, this currently depends on `BroadcastChannel` to send the message from the service worker to the client after the page has loaded:

https://github.com/xwp/pwa-wp/blob/57b295c646bfae63f78c3d735fbc37762709fca0/wp-includes/js/service-worker-error-response-handling.js#L11-L31

It would seem more straightforward to instead inject the error messages directly into the `details` element when the 500 error is being constructed. This could be done done with streams or more simply it could be done just by reading the body of the 500 error response and injecting it straight away.

miina:
Fixes #86.

Adds placeholder tag for error details and replaces using BroadcastChanel.

### DIFF
--- a/wp-includes/js/service-worker-navigation-routing.js
+++ b/wp-includes/js/service-worker-navigation-routing.js
@@ -2,18 +2,9 @@
 {
 	const isStreamingResponses = SHOULD_STREAM_RESPONSE && wp.serviceWorker.streams.isSupported();
 	const errorMessages = ERROR_MESSAGES;
-	const PWA_ESCAPE_MAP = {
-		'&': '&amp;',
-		'<': '&lt;',
-		'>': '&gt;',
-		'"': '&quot;',
-		"'": '&#39;'
-	};
 
 	wp.serviceWorker.routing.registerRoute( new wp.serviceWorker.routing.NavigationRoute(
 		async function ( { event } ) {
-			const { url } = event.request;
-
 			let responsePreloaded = false;
 
 			const canStreamResponse = () => {
@@ -65,10 +56,13 @@
 										return ''; // Remove the details from the document entirely.
 									}
 									const src = 'data:text/html;base64,' + btoa( errorText ); // The errorText encoded as a text/html data URL.
-									const srcdoc = errorText.replace( /[&<>]/g, function( char ) {
-										return PWA_ESCAPE_MAP[ char ];
-									} ); // The errorText escaped for use in an HTML attribute.
-									const iframe = `<iframe style="width:100%" src="${src}"  srcdoc="${srcdoc}"></iframe>`;
+									const srcdoc = errorText
+										.replace( /&/g, '&amp;' )
+										.replace( /'/g, '&#39;' )
+										.replace( /"/g, '&quot;' )
+										.replace( /</g, '&lt;' )
+										.replace( />/g, '&gt;' );
+									const iframe = `<iframe style="width:100%" src="${src}" data-srcdoc="${srcdoc}"></iframe>`;
 									details = details.replace( '{{{error_details_iframe}}}', iframe );
 									// The following are in case the user wants to include the <iframe> in the template.
 									details = details.replace( '{{{iframe_src}}}', src );
--- a/wp-includes/js/service-worker-navigation-routing.js
+++ b/wp-includes/js/service-worker-navigation-routing.js
@@ -2,6 +2,13 @@
 {
 	const isStreamingResponses = SHOULD_STREAM_RESPONSE && wp.serviceWorker.streams.isSupported();
 	const errorMessages = ERROR_MESSAGES;
+	const PWA_ESCAPE_MAP = {
+		'&': '&amp;',
+		'<': '&lt;',
+		'>': '&gt;',
+		'"': '&quot;',
+		"'": '&#39;'
+	};
 
 	wp.serviceWorker.routing.registerRoute( new wp.serviceWorker.routing.NavigationRoute(
 		async function ( { event } ) {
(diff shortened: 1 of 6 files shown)


wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -31,29 +31,32 @@
...
+
+						return errorResponse.text().then( function( text ) {
+							let init = {
+								status: errorResponse.status,
+								statusText: errorResponse.statusText,
+								headers: errorResponse.headers
+							};
+
+							let body = text.replace( /<!--WP_SERVICE_WORKER_ERROR_MESSAGE-->/, errorMessages.error );
+							body = body.replace( /<!--WP_SERVICE_WORKER_ERROR_DETAILS-->/, errorText );
```
miina: @westonruter Should we have a placeholder text here in case there are no details for the error? For example `No details received for the error` or something similar?

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -31,29 +31,32 @@
...
 
-				// Close the channel if client did not request the message within 30 seconds.
-				setTimeout( () => {
-					channel.close();
-				}, 30 * 1000 );
+				if ( canStreamResponse() ) {
+					return caches.match( ERROR_500_BODY_FRAGMENT_URL );
+				}
+
+				return response.text().then( function( errorText ) {
```
miina: This part is duplicated in the offline commenting section.

wp-includes/js/service-worker-navigation-routing.js
westonruter: That's a good question. I don't think so, since if the `details` element is used then it would be expanded to find no additional details. The problem seems to be that the container element needs to be omitted if there is no error message, right?

miina: Maybe, I had the thought that for consistency it would always be good to display the `details` even if there are no details -- to state explicitly there are no details so that it would be clear there were no details. Otherwise perhaps the user wouldn't know that no details were found.

If you think it's better to just not display the `details` at all then we can do that instead. Thoughts?

wp-includes/theme-compat/500.php
```diff
@@ -16,20 +16,10 @@
...
-		function renderErrorDetails( data ) {
-			if ( data.bodyText.trim().length ) {
-				const details = document.getElementById( 'error-details' );
-				details.querySelector( 'iframe' ).srcdoc = data.bodyText;
-				details.hidden = false;
-			}
-		}
-		</script>
-		<?php wp_print_service_worker_error_details_script( 'renderErrorDetails' ); ?>
+		<iframe style="width:100%;" srcdoc="<?php wp_service_worker_error_details_placeholder(); ?>"></iframe>
```
westonruter: The `wp_service_worker_error_details_placeholder()` won't work so well here because if trying to access the error page directly then you'd get:

```html
<iframe style="width:100%;" srcdoc="<!--WP_SERVICE_WORKER_ERROR_DETAILS-->"></iframe>
```

That seems like it would be an HTML parse error for some parsers.

wp-includes/js/service-worker-offline-commenting.js
```diff
@@ -12,30 +12,27 @@
...
+								statusText: errorResponse.statusText,
+								headers: errorResponse.headers
+							};
 
-				// Close the channel if client did not request the message within 30 seconds.
-				setTimeout( () => {
-					channel.close();
-				}, 30 * 1000 );
+							let body = text.replace( /<!--WP_SERVICE_WORKER_ERROR_MESSAGE-->/, errorMessages.error );
+							body = body.replace( /<!--WP_SERVICE_WORKER_ERROR_DETAILS-->/, errorText );
```
westonruter: I believe this would need to get encoded in order to be able to appear in an HTML attribute. If the `errorText` contains markup (as it is expected to) then it will likely break the `srcdoc` attribute.

wp-includes/theme-compat/500.php
westonruter: Something else to consider is that `srcdoc` is not supported by all browsers. So perhaps it would make more sense to encode the `errorText` as an HTML data URI and then populate the `src` instead.

wp-includes/js/service-worker-navigation-routing.js
westonruter: Maybe better for now to just omit the details entirely if there is no error message detected.

But I see that this presumes that the `details` isn't already present in the document.

wp-includes/service-workers.php
```diff
@@ -450,3 +451,10 @@ function wp_service_worker_get_error_messages() {
 function wp_service_worker_error_message_placeholder() {
 	echo '<p><!--WP_SERVICE_WORKER_ERROR_MESSAGE--></p>'; // WPCS: XSS OK.
 }
+
+/**
+ * Display service worker error details template tag.
+ */
+function wp_service_worker_error_details_placeholder() {
```
westonruter: @miina Here's an idea that could tie things together. What if this template tag tags arguments to construct the `details` element (or another element) to populate? This function could then also add a comment before and after the output element which could then facilitate removing the `details` entirely if there is no error message to display.

Consider this usage:

```php
wp_service_worker_error_details_template(
    // The following could be the default value.
    '<details id="error-details"><summary>' . esc_html__( 'More Details', 'pwa' ) . '</summary>{{{error_details_iframe}}}</details>' 
);
```

This could then output markup like this:

```html
<!--WP_SERVICE_WORKER_ERROR_TEMPLATE_BEGIN-->
<details id="error-details"><summary>More details</summary>{{{error_details_iframe}}}</details>
<!--WP_SERVICE_WORKER_ERROR_TEMPLATE_END-->
```

Then the service worker can do a regex replacement:

```js
text.replace(
    /(<!--WP_SERVICE_WORKER_ERROR_TEMPLATE_BEGIN-->)((?:.|\n)+?)(<!--WP_SERVICE_WORKER_ERROR_TEMPLATE_END-->)/, 
    ( matches ) => {
        if ( ! errorText ) {
            return ''; // Remove the details from the document entirely.
        }
        let details = matches[1];
        const src = '...'; // The errorText encoded as a text/html data URL.
        const srcdoc = '...'; // The errorText escaped for use in an HTML attribute.
        const iframe = `<iframe style="width:100%" src="${src}"  srcdoc="${srcdoc}"></iframe>`;
        details = details.replace( '{{{error_details_iframe}}}', iframe );
        // The following are in case the user wants to include the <iframe> in the template.
        details = details.replace( '{{{iframe_src}}}', src );
        details = details.replace( '{{{iframe_srcdoc}}}', srcdoc );
        return details;
    } 
);
```

miina: Made some changes within d89b536. Let me know if the data URL / escaped attributes seem to be correct.

westonruter: One key thing when doing HTML escaping is that the `&` has to be always replaced with `&amp;` first. Otherwise, you could end up with `<` escaped to `&lt;` and then double-escaped to`&amp;lt;`.

westonruter:
I've updated the theme to use these updated functions: https://github.com/westonruter/twentyseventeen-westonson/pull/14

PR #107, opened 2018-11-10
Improve support for plugins and themes implementing app shells

westonruter:
* Add `wp_service_worker_navigation_route` filter for themes and plugins to opt-in to using a given URL as the app shell.
* Ensure that the navigation routing logic used when app shell is used so that navigation caching strategy and error handling will apply in the same way as when app shell is not used.
* Add `wp_service_worker_navigation_route_blacklist_patterns` filter for themes and plugins to amend the `blacklist` array that is passed to the navigation route handler. Also prevent Customizer preview requests from being handled by the navigation router.
* Vary the offline, server error, and navigation route (app shell) template precache revisions by the items in the theme's registered nav menus, key options, the mods, and enqueued scripts/styles.
* Eliminate custom `Router` which reported registered route collisions since it resulted in too much duplicate code, and it also facilitated a fix to registered routes overriding the precache route handler.
* Fix unit tests when running offline.
* Fix `wp_service_worker_loaded` to only run at parse_query for expected query.
* Remove AMP-specific code to be moved over to AMP plugin instead in https://github.com/Automattic/amp-wp/pull/1519.

See https://github.com/Automattic/amp-wp/pull/1519


amp/class-amp-service-worker.php
```diff
@@ -39,6 +39,91 @@ public function init() {
...
+			// @todo There should be some query var that is used to disable navigation routing entirely so that there is no need to bypass for network.
+			// Prevent app shell from being served when requesting AMP version directly.
+			add_filter(
+				'wp_service_worker_navigation_route_blacklist_patterns',
+				function ( $blacklist_patterns ) {
+					$blacklist_patterns[] = '\?(.+&)*' . preg_quote( amp_get_slug(), '/' ) . '(=|&|$)';
+					return $blacklist_patterns;
+				}
+			);
+		}
```
westonruter: This I want to move this class over to the AMP plugin now, to be part of https://github.com/Automattic/amp-wp/pull/1519

wp-includes/js/service-worker.js
```diff
@@ -130,6 +130,63 @@ wp.serviceWorker = workbox;
...
+		 * matches).
+		 * @param {Array<RegExp>} [options.whitelist=[/./]] If any of these patterns
+		 * match the URL's pathname and search parameter, the route will handle the
+		 * request (assuming the blacklist doesn't match).
+		 * @return {workbox.routing.NavigationRoute} Returns the generated
+		 * Route.
+		 *
+		 * @alias workbox.routing.registerNavigationRoute
+		 */
+		registerNavigationRoute(cachedAssetUrl, options = {}) {
```
westonruter: This is copied straight from Workbox: https://github.com/GoogleChrome/workbox/blob/912080a1bf3255c61151ca3d0ebd0895aaf377e2/packages/workbox-routing/_default.mjs#L126-L194

Copying is needed unfortunately because we fork the Routing logic in order to customize how the route matching is done.

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -1,181 +1,249 @@
...
-			 * If navigation preload is enabled, use the preload request instead of doing another fetch.
-			 * This prevents requests from being duplicated. See <https://github.com/xwp/pwa-wp/issues/67>.
+	const blacklist = NAVIGATION_BLACKLIST_PATTERNS.map( ( pattern ) => new RegExp( pattern ) );
+	if ( navigationRouteEntry ) {
+		wp.serviceWorker.routing.registerNavigationRoute(
+			navigationRouteEntry.url,
+			{ blacklist }
+		);
+
+		class FetchNavigationRoute extends wp.serviceWorker.routing.Route {
```
westonruter: It would be nice if Workbox had a Whitelist/Blacklist route that this and `NavigationRoute` could extend. /cc @jeffposnick

wp-includes/js/service-worker.js
```diff
@@ -144,14 +201,12 @@ wp.serviceWorker = workbox;
...
-	if ( responsePromise ) {
-		event.respondWith( responsePromise );
-	}
-} );
+/*
+ * Note: The fetch event listener was moved from here to wp-includes/js/service-worker-add-routing-fetch-event-listener.js
+ * The reason for this is that the fetch event needs to be added _after_ the call to precaching.addRoute() is made, so
+ * that precached assets will be handled before falling back to other caching strategies for non-precached routes.
+ * If the caching strategy handler is added first, then the precache route would never handle it.
+ */
```
westonruter: @jeffposnick I wanted to raise this issue that we encountered. It seems that there is a conflict between the precaching and routing modules. It seems the precaching route should always come first, right? At the moment, however, the `fetch` for routing gets added immediately without the ability for something else to come before, right?

https://github.com/GoogleChrome/workbox/blob/b146d8d3c2db89fa51afc322049b9795894bb4f2/packages/workbox-routing/_default.mjs#L199-L206

jeffposnick: If you call (the wrapped equivalent of) `workbox.precache.precacheAndRoute()` earlier in the service worker file than you call `workbox.routing.registerRoute()`, the precaching should take precedence.

So yes, ordering does matter—we [just updated the docs](https://github.com/google/WebFundamentals/pull/6820/files#diff-a298f19f420180849fe4a3cde57504bfR72) to clarify that.

Are you blocked at something particular that you can't resolve via reordering?

amp/class-amp-service-worker.php
```diff
@@ -39,6 +39,91 @@ public function init() {
...
+								'blogname'        => get_option( 'blogname' ),
+								'blogdescription' => get_option( 'blogdescription' ),
+								'site_icon_url'   => get_site_icon_url(),
+								'theme_mods'      => get_theme_mods(),
+								'version'         => PWA_VERSION,
+							)
+						)
+					);
+
+					// Include all scripts and styles in revision.
```
westonruter: Move this to PWA

amp/class-amp-service-worker.php
```diff
@@ -39,6 +39,91 @@ public function init() {
...
+						$revision .= sprintf( ';%s-v%s', $stylesheet, wp_get_theme( $stylesheet )->Version );
+					}
+					$revision .= sprintf( ';user-%d', get_current_user_id() );
+
+					/*
+					 * Note that themes will need to vary the revision further by whatever is contained in the app shell.
+					 * In particular, the post_modified times of the nav menu items appearing in the header and footer should
+					 * be included in the revision.
+					 */
+					$revision .= ';' . md5(
```
amedina: The calculation of the revision here is redundant with the login in the PWA plugin.

amp/class-amp-service-worker.php
```diff
@@ -39,6 +39,91 @@ public function init() {
...
+						wp_json_encode(
+							array(
+								'blogname'        => get_option( 'blogname' ),
+								'blogdescription' => get_option( 'blogdescription' ),
+								'site_icon_url'   => get_site_icon_url(),
+								'theme_mods'      => get_theme_mods(),
```
westonruter: These might as well be added as part of the PWA plugin as well.

wp-includes/js/service-worker.js
westonruter: Thanks. I've fixed the ordering in 7572d0d15b75fa4e76bfe3038272ec8504a9312f. It was due in part to us using a custom router which logged a notice when multiple matches were made. I suppose I didn't have to eliminate the custom router as I did there to fix the ordering, but it seemed better to do for now since the custom router meant duplicating too much code.

PR #114, opened 2018-12-17
Clarify the plugin's purpose in the readme

### LINKED ISSUE #113: Create better installation/use path
This plugin does a great job at enabling developers to achieve PWA functionalities, but it doesn’t explain itself. This is primarily a documentation issue.

Looking at the docs, it’s easy to learn about PWAs—but it’s not clear what you get out of the box with this plugin, and what your next follow-up steps are.

Suggested: Add a top-level section to the plugin description that tells you what you get for free and what you need to do to reach “full PWA” status.

Some example questions people may have:
What else needs to be done after installation to have a properly-registered service worker? If nothing, where should devs access the already-installed one?
Where should a dev/user go to make changes to their manifest.json? Should those changes be made directly to a source file, or by editing WordPress values, or both?

Closing the educational gap for people will make this plugin much more usable.

I suspect @postphotos will agree heartily.

westonruter:
Fixes #113.


readme.md
```diff
@@ -38,6 +38,8 @@ In general a PWA depends on the following technologies to be available:
 
 This plugin serves as a place to implement support for these in WordPress with the intention of being proposed for core merge, piece by piece.
 
+☞ Please note that this feature plugin is _not_ intended to obsolete the other plugins and themes which turn WordPress sites into PWAs. Rather, this plugin is intended to provide the PWA building blocks and coordination mechanism for these themes and plugins to not reinvent the wheel and also to not conflict with each other. For example, a theme that implements the app shell model should be able to extend the core service worker while a plugin that provides push notifications should be able to do the same. Themes and plugins no longer should have to each create a service worker on their own, something which is inheritly problematic because only one service worker can be active at a time: only one service worker can win. This feature plugin intends to create win-win situations for themes and plugins that each leverage PWA features.
+
```
felixarntz: Maybe instead of the last sentence, more something like a CTA: "If you are developing a plugin or theme that includes a service worker, it's recommended to rely on this PWA plugin instead, or at least only use the built-in implementation as a fallback for when the PWA plugin is not available."

PR #117, opened 2019-01-16
Add theme-color theme support

### LINKED ISSUE #116: Add ability to customize the theme color without relying on custom background
In my use-case the client would like their header-color to be used as theme-color, so a filter would be nice here.

aristath:
When a theme has `custom-background` theme support, this is used as the theme color for the app. When the theme doesn't have a `custom-background` or the background color is not what is desired for the app, then a way to override it is needed without resorting to the fallback theme color. To this end, the `theme-color` theme support is introduced in this PR.

Closes #116.

### DIFF
--- a/wp-includes/class-wp-web-app-manifest.php
+++ b/wp-includes/class-wp-web-app-manifest.php
@@ -61,7 +61,7 @@ class WP_Web_App_Manifest {
 	public function manifest_link_and_meta() {
 		?>
 		<link rel="manifest" href="<?php echo esc_url( rest_url( self::REST_NAMESPACE . self::REST_ROUTE ) ); ?>">
-		<meta name="theme-color" content="<?php echo esc_attr( apply_filters( 'theme_color_meta', $this->get_theme_color() ) ); ?>">
+		<meta name="theme-color" content="<?php echo esc_attr( $this->get_theme_color() ); ?>">
 		<?php
 	}
 
@@ -85,7 +85,7 @@ class WP_Web_App_Manifest {
 			$theme_color = self::FALLBACK_THEME_COLOR;
 		}
 
-		return $theme_color;
+		return apply_filters( 'theme_color_meta', $theme_color );
 	}
 
 	/**
--- a/wp-includes/class-wp-web-app-manifest.php
+++ b/wp-includes/class-wp-web-app-manifest.php
@@ -74,18 +74,25 @@ class WP_Web_App_Manifest {
 	 * @return string $theme_color The theme color for the manifest.json file, as a hex value.
 	 */
 	public function get_theme_color() {
-		if ( current_theme_supports( 'custom-background' ) ) {
-			$background_color = get_background_color(); // This returns a hex value without the leading #, or an empty string.
-			if ( $background_color ) {
-				$theme_color = "#{$background_color}";
+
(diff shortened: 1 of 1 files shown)


wp-includes/class-wp-web-app-manifest.php
```diff
@@ -61,7 +61,7 @@ public function init() {
 	public function manifest_link_and_meta() {
 		?>
 		<link rel="manifest" href="<?php echo esc_url( rest_url( self::REST_NAMESPACE . self::REST_ROUTE ) ); ?>">
-		<meta name="theme-color" content="<?php echo esc_attr( $this->get_theme_color() ); ?>">
+		<meta name="theme-color" content="<?php echo esc_attr( apply_filters( 'theme_color_meta', $this->get_theme_color() ) ); ?>">
```
westonruter: I should think that this filter would go inside the `get_theme_color` method itself.

aristath: makes sense... moved.

westonruter:
@aristath so does the site not have `custom-background` support? I wonder if instead of introducing a new filter it would be better to integrate with the existing `get_background_color()` function:

```php
function get_background_color() {
	return get_theme_mod( 'background_color', get_theme_support( 'custom-background', 'default-color' ) );
}
```

In other words, if we got rid of the `if ( current_theme_supports( 'custom-background' ) )` check then we could allow users like yourself to continue using this method of supplying the them color via:

```php
add_filter( 'theme_mod_background_color', function( $color ) {
    if ( ! $color ) {
        $color = 'red';
    }
    return $color;
} );
```

This feels somewhat better because it doesn't mean introducing a new filter.

aristath:
The site _does_ have `custom-background` control.
However, they want to use some other color (not the background color) for their "app". My first thought was to add a filter on the `theme_mod` too, but I didn't like the idea because I have no way to only apply that filter for the webapp's manifest. So if I use the filter, the background for the whole site would change and we don't want that :smile: 
If they didn't have custom-background support, I agree, filtering the theme_mod` is the simplest, straight-forward solution. But what if it does?

westonruter:
OK, good info.

Instead of a filter, what about adding a new theme support flag, like `theme-color`? That would make it more aligned with `custom-background` theme mod. 

Alternatively, the `theme-color` could be a property of `custom-background`. In other words:

```php
	add_theme_support(
		'custom-background',
		array(
			'default-color' => 'e6e6e6',
			'theme-color' => 'ffffff',
		)
	);
```

I suppose that's not great either as it overloads the `custom-background`.

aristath:
I hadn't thought of a new `theme-support`... It would make perfect sense to have something like this to get the theme-color:
```php
if ( current_theme_supports( 'theme-color' ) ) {
	$theme_color = get_theme_color() // Or whatever we want to call this function.
} else {
	if ( current_theme_supports( 'custom-background' ) ) {
		// Sanitize function adds the hash to the get_background_color() function.
		$theme_color = sanitize_hex_color( get_background_color() );
	}
	return ( $theme_color ) ? $theme_color : self::FALLBACK_THEME_COLOR;
}
```

Adding inside the `custom-background` just doesn't feel right IMO, it becomes too overloaded

aristath:
Pushed another commit :+1:

PR #121, opened 2019-01-20
Update to wp-dev-lib 1.0.0 and fix phpcs issues

westonruter:
This implements the same changes that were made in the AMP plugin: https://github.com/ampproject/amp-wp/pull/1131


westonruter:
Just we bump the minimum version to 5.6 as well? If core is going to be bumped to 5.6 by the time we propose merge, then this would make sense to me.

PR #125, opened 2019-01-24
Separate service worker integrations from core and hide them behind feature flag

felixarntz:
The service worker integrations are, while being super-useful, not part of the service worker core code. As discussed in person earlier, let's separate them from the rest of the codebase and hide them behind a feature flag.

This PR does this, moving the related function and hook registration into a separate `service-worker-integrations.php` file. All files related to integrations are then only loaded conditionally, based on a `wp_service_worker_integrations_enabled` filter or a constant of the same name.

A brief documentation segment is also included. Plus there is a minor related change for some clean up, moving a few theme-related functions from `service-workers.php` into `template.php`.


pwa.php
```diff
@@ -132,6 +116,38 @@ function _pwa_incorrect_plugin_slug_admin_notice() {
...
+ *
+ * Instead of using this filter, you can also use a constant `WP_SERVICE_WORKER_INTEGRATIONS_ENABLED`.
+ *
+ * @since 0.2
+ *
+ * @param bool $enabled Whether or not service worker integrations are enabled.
+ */
+if ( apply_filters( 'wp_service_worker_integrations_enabled', defined( 'WP_SERVICE_WORKER_INTEGRATIONS_ENABLED' ) && WP_SERVICE_WORKER_INTEGRATIONS_ENABLED ) ) {
+	/** WP_Service_Worker_Integration Interface */
+	require_once PWA_PLUGIN_DIR . '/wp-includes/integrations/interface-wp-service-worker-integration.php';
```
westonruter: Can we move the entire `integrations` directory out of `wp-includes`? That will make it much clearer what is in scope for a merge proposal and core patch.

westonruter:
@aristath FYI: After this is merged, you may beed to add to your theme/plugin:

```php
add_filter( 'wp_service_worker_integrations_enabled', '__return_true' );
```

aristath:
@westonruter thank you for the heads-up :+1:

Issue #122, opened 2019-01-21
Error with isStreamingResponses in Safari


aristath:
I've had reports from clients that on Safari they can't access a website where PWA was enabled. The error they were getting is shown in the screenshot below:
![image823](https://user-images.githubusercontent.com/588688/51471711-26791680-1d80-11e9-8161-5a0d737a8895.png)

From what i can tell `isStreamingResponses` only exists here:
https://github.com/xwp/pwa-wp/blob/733563f46996ea2d5aaf52a576971f92c3748164/wp-includes/js/service-worker-navigation-routing.js#L6-L21

Getting this fixed was a bit urgent for me, so commenting out these lines that enqueued the script seemed to get the job done
https://github.com/xwp/pwa-wp/blob/733563f46996ea2d5aaf52a576971f92c3748164/wp-includes/components/class-wp-service-worker-navigation-routing-component.php#L475-L481

westonruter:
I recall now from @miina that Safari fails to properly handle `const` variables that are defined lexically in this way. If you switch to `let` or `var` does that make the problem go away?

aristath:
I currently don't have Safari to test, but @ilicfilip tested changing these 3 lines from `const` to `var` and that fixed the issue: https://github.com/xwp/pwa-wp/blob/dcdc5795f8d5f90b008a5db034ccb5357a9f5c11/wp-includes/js/service-worker-navigation-routing.js#L6-L8

aristath:
@westonruter pushed :+1:

PR #128, opened 2019-01-28
Fix isStreamingResponses error in Safari

aristath:
See #122

### DIFF
--- a/wp-includes/js/service-worker-navigation-routing.js
+++ b/wp-includes/js/service-worker-navigation-routing.js
@@ -3,9 +3,9 @@ ERROR_OFFLINE_URL, ERROR_500_URL, SHOULD_STREAM_RESPONSE, STREAM_HEADER_FRAGMENT
 ERROR_OFFLINE_BODY_FRAGMENT_URL, STREAM_HEADER_FRAGMENT_QUERY_VAR, NAVIGATION_BLACKLIST_PATTERNS, ERROR_MESSAGES */
 
 {
-	const isStreamingResponses = SHOULD_STREAM_RESPONSE && wp.serviceWorker.streams.isSupported();
-	const errorMessages = ERROR_MESSAGES;
-	const navigationRouteEntry = NAVIGATION_ROUTE_ENTRY;
+	var isStreamingResponses = SHOULD_STREAM_RESPONSE && wp.serviceWorker.streams.isSupported();
+	var errorMessages = ERROR_MESSAGES;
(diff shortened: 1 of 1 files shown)


wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -3,9 +3,9 @@ ERROR_OFFLINE_URL, ERROR_500_URL, SHOULD_STREAM_RESPONSE, STREAM_HEADER_FRAGMENT
 ERROR_OFFLINE_BODY_FRAGMENT_URL, STREAM_HEADER_FRAGMENT_QUERY_VAR, NAVIGATION_BLACKLIST_PATTERNS, ERROR_MESSAGES */
 
 {
-	const isStreamingResponses = SHOULD_STREAM_RESPONSE && wp.serviceWorker.streams.isSupported();
-	const errorMessages = ERROR_MESSAGES;
-	const navigationRouteEntry = NAVIGATION_ROUTE_ENTRY;
+	var isStreamingResponses = SHOULD_STREAM_RESPONSE && wp.serviceWorker.streams.isSupported();
+	var errorMessages = ERROR_MESSAGES;
+	var navigationRouteEntry = NAVIGATION_ROUTE_ENTRY;
```
westonruter: I just remembered where this was fixed before. In the AMP plugin an IIFE was used instead of switching to `var`: https://github.com/ampproject/amp-wp/commit/60b3bca47ad26eb8d82abb78a035cda72a6c1b51

Would you be willing to make a similar change to this file and the other JS files in this directory?

westonruter: Also, please add a comment before the IIFE to explain why it is used:

```js
// IIFE is used for lexical scoping instead of just a braces block due to bug with const in Safari.
( () => {
// ...
} )();
```

aristath: Done. Thanks for the link... IIFE does seem like a better approach. :heart:

westonruter:
@aristath can you apply the same change to:

https://github.com/xwp/pwa-wp/blob/103b5960602b0f2be407e9cbed24150d4239a94e/wp-includes/js/service-worker-offline-commenting.js#L2-L4

And perhaps:

https://github.com/xwp/pwa-wp/blob/103b5960602b0f2be407e9cbed24150d4239a94e/wp-includes/js/service-worker-precaching.js#L2-L3

aristath:
meh sorry, I commented on the ticket instead. All done. :+1:

Issue #118, opened 2019-01-16
Make app installable on iOS


aristath:
I don't have an iOS device to test with, but I think this is also needed:

```html
<meta name="apple-mobile-web-app-capable" content="yes">
```

aristath:
I know this looks like a scenario out a nightmare, but this is what it took to make the app installable on iOS:
```php
<meta name="apple-mobile-web-app-capable" content="yes" />
<meta name="mobile-web-app-capable" content="yes" />
<meta name="apple-touch-fullscreen" content="YES" />
<link rel="apple-touch-startup-image" href="<?php echo esc_url_raw( $icon_url ); ?>">
<link href="<?php echo esc_url_raw( $icon_url ); ?>" rel="apple-touch-icon">
<meta name="apple-mobile-web-app-title" content="<?php echo esc_attr( get_bloginfo( 'name' ) ); ?>">
<meta name='application-name' content='<?php echo esc_attr( get_bloginfo( 'name' ) ); ?>'>
```

westonruter:
> `<link href="<?php echo esc_url_raw( $icon_url ); ?>" rel="apple-touch-icon">`

I believe `apple-touch-icon` is handled by core in `wp_site_icon()`:

```php
	$icon_180 = get_site_icon_url( 180 );
	if ( $icon_180 ) {
		$meta_tags[] = sprintf( '<link rel="apple-touch-icon-precomposed" href="%s" />', esc_url( $icon_180 ) );
	}
```

Aside, the latest beta of Safari is still not respecting the icons in the Web App Manifest: https://medium.com/dev-channel/progressive-web-app-progress-in-ios-12-2-beta-1-build-16e5181f-a18cd05ca361

> `<meta name="apple-touch-fullscreen" content="YES" />`

As for `apple-touch-fullscreen`, I think this should only be done if the web app manifest has `display` set to `fullscreen` and perhaps `standalone`.

> `<link rel="apple-touch-startup-image" href="<?php echo esc_url_raw( $icon_url ); ?>">`

Would this need to be different than the other icon?

> `<meta name="apple-mobile-web-app-title" content="<?php echo esc_attr( get_bloginfo( 'name' ) ); ?>">`
> <meta name='application-name' content='<?php echo esc_attr( get_bloginfo( 'name' ) ); ?>'>

Probably could simply be the same value as the `name` for the web app manifest, yes?

> `<meta name="apple-mobile-web-app-capable" content="yes" />`
[...]

Issue #124, opened 2019-01-24
Update PHP requirements to 5.6 and install wp-dev-lib via Composer


felixarntz:
Since WordPress will most likely have raised its minimum PHP requirement to 5.6 by the time we are gonna propose this for merge, let's do that now. We can still leave the code 5.2-compatible for the time being, at least as long as improving it wouldn't introduce significant benefits.

With that change, let's also install `wp-dev-lib` via Composer, in accordance with how the AMP plugin does it.

See https://github.com/xwp/pwa-wp/pull/121#issuecomment-457313936 and https://github.com/ampproject/amp-wp/pull/1131

PR #129, opened 2019-01-29
Add extra <meta> in the head for iOS

aristath:
See #118

### DIFF
--- a/wp-includes/class-wp-web-app-manifest.php
+++ b/wp-includes/class-wp-web-app-manifest.php
@@ -63,18 +63,26 @@ class WP_Web_App_Manifest {
 		?>
 		<link rel="manifest" href="<?php echo esc_url( rest_url( self::REST_NAMESPACE . self::REST_ROUTE ) ); ?>">
 		<meta name="theme-color" content="<?php echo esc_attr( $manifest['theme_color'] ); ?>">
-		<meta name="apple-mobile-web-app-capable" content="yes" />
-		<meta name="mobile-web-app-capable" content="yes" />
-		<meta name="apple-touch-fullscreen" content="YES" />
+		<meta name="apple-mobile-web-app-capable" content="yes">
+		<meta name="mobile-web-app-capable" content="yes">
+		<meta name="apple-touch-fullscreen" content="YES">
 		<?php if ( ! empty( $manifest['icons'] ) ) : ?>
-			<?php if ( ! empty( $manifest['icons'][1] ) && ! empty( $manifest['icons'][1]['src'] ) ) : ?>
-				<link rel="apple-touch-startup-image" href="<?php echo esc_url_raw( $manifest['icons'][1]['src'] ); ?>">
-			<?php elseif ( ! empty( $manifest['icons'][0] ) && ! empty( $manifest['icons'][0]['src'] ) ) : ?>
-				<link rel="apple-touch-startup-image" href="<?php echo esc_url_raw( $manifest['icons'][0]['src'] ); ?>">
+			<?php
+			$icons = $manifest['icons'];
+			usort(
+				$icons,
+				function( $a, $b ) {
+					return intval( strtok( $a['sizes'], 'x' ) ) - intval( strtok( $b['sizes'], 'x' ) );
+				}
+			);
+			$icon = array_shift( $icons );
+			?>
+			<?php if ( ! empty( $icon ) ) : ?>
+				<link rel="apple-touch-startup-image" href="<?php echo esc_url( $icon['src'] ); ?>">
 			<?php endif; ?>
 		<?php endif; ?>
 		<meta name="apple-mobile-web-app-title" content="<?php echo esc_attr( $manifest['short_name'] ); ?>">
-		<meta name='application-name' content='<?php echo esc_attr( $manifest['short_name'] ); ?>'>
+		<meta name="application-name" content="<?php echo esc_attr( $manifest['short_name'] ); ?>">
 		<?php
 	}
 
@@ -201,12 +209,12 @@ class WP_Web_App_Manifest {
 	 *
 	 * Mainly copied from Jetpack_PWA_Manifest::build_icon_object() and Jetpack_PWA_Helpers::site_icon_url().
 	 *
-	 * @return array|null $icon_object An array of icons, or null if there's no site icon.
+	 * @return array $icon_object An array of icons, which may be empty.
 	 */
 	public function get_icons() {
 		$site_icon_id = get_option( 'site_icon' );
 		if ( ! $site_icon_id || ! function_exists( 'get_site_icon_url' ) ) {
-			return null;
+			return array();
 		}
 
 		$icons     = array();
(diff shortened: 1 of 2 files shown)


wp-includes/class-wp-web-app-manifest.php
```diff
@@ -59,9 +59,22 @@ public function init() {
 	 * Mainly copied from Jetpack_PWA_Manifest::render_manifest_link().
 	 */
 	public function manifest_link_and_meta() {
+		$manifest = $this->get_manifest();
 		?>
 		<link rel="manifest" href="<?php echo esc_url( rest_url( self::REST_NAMESPACE . self::REST_ROUTE ) ); ?>">
-		<meta name="theme-color" content="<?php echo esc_attr( $this->get_theme_color() ); ?>">
+		<meta name="theme-color" content="<?php echo esc_attr( $manifest['theme_color'] ); ?>">
+		<meta name="apple-mobile-web-app-capable" content="yes" />
```
westonruter: We can remove the trailing `/` here and below. No need to pretend we're XHTML anymore 😄

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -59,9 +59,22 @@ public function init() {
...
+		<meta name="mobile-web-app-capable" content="yes" />
+		<meta name="apple-touch-fullscreen" content="YES" />
+		<?php if ( ! empty( $manifest['icons'] ) ) : ?>
+			<?php if ( ! empty( $manifest['icons'][1] ) && ! empty( $manifest['icons'][1]['src'] ) ) : ?>
+				<link rel="apple-touch-startup-image" href="<?php echo esc_url_raw( $manifest['icons'][1]['src'] ); ?>">
+			<?php elseif ( ! empty( $manifest['icons'][0] ) && ! empty( $manifest['icons'][0]['src'] ) ) : ?>
+				<link rel="apple-touch-startup-image" href="<?php echo esc_url_raw( $manifest['icons'][0]['src'] ); ?>">
```
westonruter: Is there a need for duplication here? To me it seems what you want to do is sort all icons by size and then get the first one, yes?

For example, first given this change:

```diff
diff --git a/wp-includes/class-wp-web-app-manifest.php b/wp-includes/class-wp-web-app-manifest.php
index b7c9552..21f77fd 100644
--- a/wp-includes/class-wp-web-app-manifest.php
+++ b/wp-includes/class-wp-web-app-manifest.php
@@ -181,12 +181,12 @@ class WP_Web_App_Manifest {
 	 *
 	 * Mainly copied from Jetpack_PWA_Manifest::build_icon_object() and Jetpack_PWA_Helpers::site_icon_url().
 	 *
-	 * @return array|null $icon_object An array of icons, or null if there's no site icon.
+	 * @return array $icon_object An array of icons, which may be empty.
 	 */
 	public function get_icons() {
 		$site_icon_id = get_option( 'site_icon' );
 		if ( ! $site_icon_id || ! function_exists( 'get_site_icon_url' ) ) {
-			return null;
+			return array();
 		}
 
 		$icons     = array();
```

Then change the above code to:


```php
$manifest = $this->get_manifest();
$icons = $manifest['icons'];
usort(
	$icons,
	function( $a, $b ) {
		return intval( strtok( $a['sizes'], 'x' ) ) - intval( strtok( $b['sizes'], 'x' ) );
	}
);
$icon = array_shift( $icons );
if ( ! empty( $icon ) ) ?>
	...Now print the link...
<?php endif;
```

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -59,9 +59,22 @@ public function init() {
...
+			<?php endif; ?>
+		<?php endif; ?>
+		<meta name="apple-mobile-web-app-title" content="<?php echo esc_attr( $manifest['short_name'] ); ?>">
+		<meta name='application-name' content='<?php echo esc_attr( $manifest['short_name'] ); ?>'>
```
westonruter: Let's use double-quoted attribute values here.

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -59,9 +59,22 @@ public function init() {
...
 		?>
 		<link rel="manifest" href="<?php echo esc_url( rest_url( self::REST_NAMESPACE . self::REST_ROUTE ) ); ?>">
-		<meta name="theme-color" content="<?php echo esc_attr( $this->get_theme_color() ); ?>">
+		<meta name="theme-color" content="<?php echo esc_attr( $manifest['theme_color'] ); ?>">
+		<meta name="apple-mobile-web-app-capable" content="yes" />
+		<meta name="mobile-web-app-capable" content="yes" />
+		<meta name="apple-touch-fullscreen" content="YES" />
+		<?php if ( ! empty( $manifest['icons'] ) ) : ?>
+			<?php if ( ! empty( $manifest['icons'][1] ) && ! empty( $manifest['icons'][1]['src'] ) ) : ?>
+				<link rel="apple-touch-startup-image" href="<?php echo esc_url_raw( $manifest['icons'][1]['src'] ); ?>">
```
westonruter: This should use `esc_url()` not `esc_url_raw()`

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -59,9 +59,30 @@ public function init() {
...
 	public function manifest_link_and_meta() {
+		$manifest = $this->get_manifest();
 		?>
 		<link rel="manifest" href="<?php echo esc_url( rest_url( self::REST_NAMESPACE . self::REST_ROUTE ) ); ?>">
-		<meta name="theme-color" content="<?php echo esc_attr( $this->get_theme_color() ); ?>">
+		<meta name="theme-color" content="<?php echo esc_attr( $manifest['theme_color'] ); ?>">
+		<meta name="apple-mobile-web-app-capable" content="yes">
+		<meta name="mobile-web-app-capable" content="yes">
+		<meta name="apple-touch-fullscreen" content="YES">
+		<?php if ( ! empty( $manifest['icons'] ) ) : ?>
```
westonruter: This can condition can be removed now that `$manifest['icons']` always is an array.

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -59,9 +59,30 @@ public function init() {
...
+			<?php
+			$icons = $manifest['icons'];
+			usort(
+				$icons,
+				function( $a, $b ) {
```
westonruter: The use of a closure is causing a syntax error in PHP 5.2. Refactoring it into a class method would fix the problem. I gave the closure as more of an example code not not necessarily for it to be used verbatim. Nevertheless, we are also talking about updating the minimum required PHP version to 5.6 (#124), at which point this would be unnecessary.

westonruter:
There's a PHPCS error:

> class-wp-web-app-manifest.php:227:7: error - Whitespace found at end of line (Squiz.WhiteSpace.SuperfluousWhitespace.EndLine)

And a failing unit test:

> 1) Test_WP_Web_App_Manifest::test_get_icons
> Array () does not match expected type "NULL".
> /tmp/wordpress/src/wp-content/plugins/pwa-wp/tests/test-class-wp-web-app-manifest.php:203

aristath:
That last commit should fix the errors

Issue #131, opened 2019-02-05
Array to string conversion


aristath:
PHP error caused by the theme_support added in #118 
If a theme adds something like `add_theme_support( 'theme-color', '#333333' );` then `get_theme_support( 'theme-color' )` returns an array and that throws an error in `esc_attr()` because it's expecting a string.

PR #132, opened 2019-02-05
Fix array to string conversion

aristath:
This one fixes https://github.com/xwp/pwa-wp/issues/131
Adds a simple check to properly get the theme-color.


wp-includes/class-wp-web-app-manifest.php
```diff
@@ -93,6 +93,9 @@ public function get_theme_color() {
 		if ( current_theme_supports( 'theme-color' ) ) {
 			$theme_color = get_theme_support( 'theme-color' );
 			if ( $theme_color ) {
+				if ( is_array( $theme_color ) && ! empty( $theme_color[0] ) ) {
```
westonruter: I think `! empty( $theme_color[0] )` should be removed here, and the line below can just do:
```php
return array_shift( $theme_color );
```
This will ensure that an array is never returned from this method.

aristath: You're right. Pushed another commit :+1:

PR #138, opened 2019-02-11
Update readme.md

WraithKenny:
Fix readme formating error for php script with opening `<?php`

The PHP formatter errors on the PRE version, making the rest of the readme unreadable. This simple fix solves it.


westonruter:
Thanks. The `readme.md` is actually generated from the `readme.txt`. The lack of quote marks on the `pre[lang]` attribute was causing the problem. Fixed in 65bc42e.

PR #139, opened 2019-02-14
Update contributing guidelines

swissspidy:
Needed in preparation for a move to a Google-owned GitHub org.


westonruter:
I don't think this is needed as I don't believe we'd be moving this repo to a different organization. It's a feature plugin started at XWP which we'll be contributing to for proposal for core.

Issue #119, opened 2019-01-16
Unable to register callback on `wp_front_service_worker` action


su-narthur:
I'm trying to register a service worker in a plugin I'm developing. I've installed `pwa-wp`, and then added this to my [plugin-name].php file, and it isn't doing anything.

```php
function registerServiceWorker()
{
	var_dump('hello');die;
}

\add_action(
	"wp_front_service_worker",
	"registerServiceWorker"
);
```

What do I need to do to be able to use this action so I can go on to register my service worker script?

Thanks!

westonruter:
The action only is triggered when the service worker is actually being served. So you'll not see `registerServiceWorker()` invoked on regular requests. Only the request to the endpoint that serves back the service worker JS will result in it getting called.

su-narthur:
Thanks! In general, what are the steps I need to work through to get a service worker running?

westonruter:
@su-narthur you can use examples [from the readme](https://github.com/xwp/pwa-wp#service-workers), but otherwise it depends on heavily what you want to use the service worker for. You'd be better off referring to Workbox for general use cases: https://developers.google.com/web/tools/workbox/guides/common-recipes

westonruter:
The service worker installation script should be printed at the `wp_print_scripts` action as can be seen here: 

https://github.com/xwp/pwa-wp/blob/c3b06fe6f7cfc78538c2b256448a41e4a050602d/wp-includes/default-filters.php#L10-L13

The logic for `wp_print_service_workers()` is as follows:

https://github.com/xwp/pwa-wp/blob/733563f46996ea2d5aaf52a576971f92c3748164/wp-includes/service-workers.php#L112-L191

It could be that the `$on_front_domain` variable is `false` if you are on `localhost` the `home_url()` is not the same. That's where you can try debugging.

su-narthur:
@westonruter Yeah, that seems like it may be the problem.  `$_SERVER['HTTP_HOST']` evaluates to `localhost:8000`, whereas `wp_parse_url( home_url(), PHP_URL_HOST )` evaluates to just `localhost` without the port, so `$on_front_domain` ends up as `false`. This doesn't appear to be an issue with my settings, becuase I have both the WordPress Address and the Site Address settings set to `http://localhost:8000`.

Would you be interested in a PR that accounted for ports on localhost?

westonruter:
@su-narthur yes, a PR would be great!

su-narthur:
@westonruter Great!

[...]

PR #143, opened 2019-03-26
Don't force HTTPS for front worker script on localhost

narthur:
Without this change, https is specified when pulling in the service worker script even on localhost without ssl. By removing the explicit schema, the generated url is https with ssl and http on localhost, allowing the front service worker to be successfully registered.


westonruter:
See also https://github.com/xwp/pwa-wp/issues/119#issuecomment-469081079

PR #147, opened 2019-04-02
Prefix cache names for PHP-registered caching strategies

### LINKED ISSUE #146: Automatically prefix cacheName for the current site ID
Consider the following code:

```php
$service_workers->caching_routes()->register(
	'^' . preg_quote( set_url_scheme( content_url( '/' ), 'https' ), '/' ) . '[^\?]+?\.(?:png|gif|jpg|jpeg|svg|webp)(\?.*)?$',
	array(
		'cacheName' => 'images',
		'plugins'   => array(
			'cacheableResponse' => array(
				'statuses' => array( 0, 200 ),
			),
			'expiration'        => array(
				'maxEntries'    => 60,
				'maxAgeSeconds' => MONTH_IN_SECONDS,
			),
		),
	)
);
```

On a subdirectory multisite configuration, the `images` cache here will be shared across all sites. This could result in sites polluting each other's caches. For example, the caches look like this:

image

Note that we are prefixing the `runtime` and `precache` caches via the `prefix` configuration in Workbox:

https://github.com/xwp/pwa-wp/blob/34d171e69ea50d60d65ef2140a921e2b79fc8a1a/wp-includes/components/class-wp-service-worker-configuration-component.php#L76-L80

But this doesn't apply to other caches.

So the question is: should we automatically prepend this same prefix to any `cacheName` supplied when registering a new caching route?

westonruter:
Prevent collisions in caches for subdirectory installs.

Fixes #146.

Before:

image

After:

image


wp-includes/components/class-wp-service-worker-configuration-component.php
```diff
@@ -72,19 +72,24 @@ public function get_script() {
...
+			'prefix'   => $prefix,
 			// Also precache name by scope (front vs admin) so that different assets can be precached in each respective application.
 			'precache' => sprintf( 'precache-%s', WP_Service_Workers::SCOPE_FRONT === $current_scope ? 'front' : 'admin' ),
 			'suffix'   => 'v1',
 		);
 
 		$script .= sprintf( "workbox.core.setCacheNameDetails( %s );\n", wp_service_worker_json_encode( $cache_name_details ) );
 
+		// Export the prefix as a constant since it is not accessible via workbox.core.cacheNames, unfortunately (yet).
+		$script .= sprintf( 'const WP_SERVICE_WORKER_CACHE_PREFIX = %s;', wp_service_worker_json_encode( $prefix ) );
```
westonruter: See https://github.com/GoogleChrome/workbox/issues/1998 for an enhancement to Workbox to allow accessing the cache prefix.

westonruter: And see fix in https://github.com/GoogleChrome/workbox/pull/2001 which we can take advantage of once the next version of Workbox is released.

felixarntz: Just double-checking, can the cache name prefix contain special characters such as slashes? Typically the subdirectory in a multisite is just a single folder, but that can be different. _If_ there is a restriction in that regard, we should sanitize the prefix to replace such characters (for example with dashes or underscores).

westonruter: Slashes haven't caused a problem yet. The runtime and precache caches have worked fine with the `/` in the name.

westonruter: If we merge https://github.com/xwp/pwa-wp/pull/150 first then we can avoid adding this `const`.

westonruter: As Workbox 4.2.0 has been released with the `prefix` getter.

westonruter:
Depends on merging https://github.com/xwp/pwa-wp/pull/150 which will allow this PR to improve how it gets the prefix.

PR #149, opened 2019-04-02
Use service_worker theme support flag instead of constant/filter for integrations

### LINKED ISSUE #148: Enable integrations via theme support
Following https://github.com/xwp/pwa-wp/commit/34a2ed43a21e812e6dd21767174233bedddd439d, integrations that are built on the service worker core foundation are opt-in, hidden behind a `WP_SERVICE_WORKER_INTEGRATIONS_ENABLED` constant. This is to keep the foundation separate from these integrations.

Following the conversation around service worker theme support for the AMP plugin (see https://github.com/ampproject/amp-wp/pull/1261), we should use theme support for the integrations in the PWA plugin as well:
* Allow theme developers to do the following:
```php
add_theme_support( 'service_worker', array(
    'wp-site-icon'     => true,
    'wp-custom-header' => true,
    // More service worker integrations.
) );
```
* The `WP_SERVICE_WORKER_INTEGRATIONS_ENABLED` constant should be removed, and thus we should go back to loading these files unconditionally. Alternatively, we only load the three base files handling integration and defer loading the specific integration classes to when they're needed (see below).
* Each of the integrations in https://github.com/xwp/pwa-wp/commit/34a2ed43a21e812e6dd21767174233bedddd439d#diff-b8397f4353e4396828c88a9f9fc51ae3R24 should only be instantiated if service worker theme support for the integration is added. Per the suggestion in the other bullet point, we might also load the integration here on request.

cc @westonruter

westonruter:
Fixes #148 .

### DIFF
--- a/integrations/functions.php
+++ b/integrations/functions.php
@@ -10,42 +10,71 @@
  */
 
 /**
- * Registers all default service workers.
+ * Registers service worker integrations.
+ *
+ * These integrations are separate from the service worker core and need to be opted in via theme support.
+ * To enable all integrations, use the following:
+ *
+ *     add_theme_support( 'service_worker', true );
+ *
+ * Alternatively, you can also pass an array of integration slugs as keys and a boolean indicating that
+ * integration's status as values, for example:
+ *
+ *     add_theme_support(
+ *         'service_worker',
+ *         array(
+ *             'wp-custom-logo'   => true,
+ *             'wp-custom-header' => true,
+ *         )
+ *     );
  *
  * @since 0.2
  *
  * @param WP_Service_Worker_Scripts $scripts Instance to register service worker behavior with.
  */
 function pwa_register_service_worker_integrations( $scripts ) {
+	// Bail if not supported by theme.
+	$theme_support = get_theme_support( 'service_worker' );
+	if ( ! $theme_support ) {
+		return;
+	}
+
+	// Set script default parameters.
 	$scripts->base_url        = site_url();
 	$scripts->content_url     = defined( 'WP_CONTENT_URL' ) ? WP_CONTENT_URL : '';
 	$scripts->default_version = get_bloginfo( 'version' );
 
+	// Load integration base.
+	require_once PWA_PLUGIN_DIR . '/integrations/interface-wp-service-worker-integration.php';
+	require_once PWA_PLUGIN_DIR . '/integrations/class-wp-service-worker-base-integration.php';
+
 	$integrations = array(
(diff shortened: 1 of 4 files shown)


integrations/functions.php
```diff
@@ -16,7 +16,7 @@
  *
  * @param WP_Service_Worker_Scripts $scripts Instance to register service worker behavior with.
  */
-function wp_default_service_workers( $scripts ) {
+function pwa_register_service_worker_integrations( $scripts ) {
```
felixarntz: Nitpicking, but we can use a typehint for the class here. :)

felixarntz:
@westonruter I added the granular enabling of integrations by supporting an array to be passed to `service_worker` theme support. Related to that, I ensured that the integration classes are only loaded on demand.

I removed the filter for the integrations for now, as it is still easily possible to provide integrations without it, and having that filter might cause confusion in combination with the theme support method.

Let me know what you think.

readme.txt
```diff
@@ -152,22 +152,15 @@ See [labeled GitHub issues](https://github.com/xwp/pwa-wp/issues?q=label%3Aservi
...
-Plugin Name: Enable Service Worker Integrations
-Plugin URI:  https://github.com/xwp/pwa-wp
-Description: Enables the experimental service worker integrations in the PWA plugin.
-Version:     1.0.0
-Author:      XWP, Google, and contributors
-Author URI:  https://github.com/xwp/pwa-wp
-*/
-
-add_filter( 'wp_service_worker_integrations_enabled', '__return_true' );
+add_theme_support( 'service_worker', true );
```
westonruter: @felixarntz should this be updated to include the array syntax?

integrations/functions.php
```diff
@@ -10,42 +10,71 @@
...
+ *     add_theme_support( 'service_worker', true );
+ *
+ * Alternatively, you can also pass an array of integration slugs as keys and a boolean indicating that
+ * integration's status as values, for example:
+ *
+ *     add_theme_support(
+ *         'service_worker',
+ *         array(
+ *             'wp-custom-logo'   => true,
+ *             'wp-custom-header' => true,
```
westonruter: Since they all start with `wp-` should this part be omitted?

readme.txt
felixarntz: Added an extra example to the readme.

integrations/functions.php
felixarntz: I think they should remain as it clarifies these are "WordPress default features" (as in the PWA being a feature plugin). Other plugins could add their own and integrate with this, then use another prefix.

PR #150, opened 2019-04-05
Improve build process and update to require PHP 5.6

### LINKED ISSUE #124: Update PHP requirements to 5.6 and install wp-dev-lib via Composer

westonruter:
* Remove workbox files from being version controlled; install Workbox.js as part of build step.
* Eliminate `composer install` from `npm install`.
* Add admin notice for when running from source.
* Bump required PHP to 5.6.
* Upgrade Workbox to 4.2.0.
* Install wp-dev-lib via Composer instead of NPM.

Fixes #124.


westonruter:
Build is failing because, of course, the plugin is short-circuiting since the Workbox.js files are no longer in the repo! So Travis needs to do `npm run build`.

pwa.php
```diff
@@ -17,7 +17,7 @@
  * License:           GPL-2.0+
  * License URI:       http://www.gnu.org/licenses/gpl-2.0.txt
  * GitHub Plugin URI: https://github.com/xwp/pwa-wp
- * Requires PHP:      5.2
+ * Requires PHP:      5.6
  * Requires WP:       4.9
```
felixarntz: If we require PHP 5.6 because of WordPress core doing so, we should also require WordPress 5.2 here.

composer.json
```diff
@@ -5,11 +5,12 @@
...
   "license": "GPL-2.0",
   "require": {
-    "php": "^5.2.0 || ^7.0"
+    "php": "^5.6.0 || ^7.0"
   },
   "require-dev": {
     "dealerdirect/phpcodesniffer-composer-installer": "^0.4.4",
     "phpcompatibility/phpcompatibility-wp": "^1",
-    "wp-coding-standards/wpcs": "^1"
+    "wp-coding-standards/wpcs": "^1",
```
felixarntz: Probably better in a separate PR, but we should update these to `^2`.

westonruter:
> Why do we update the minimum required PHP version? I'd argue if it's only because WordPress does so we should:

@felixarntz It's so that we can start using PHP 5.6 features. However, plugins can still support WordPress versions older than 5.2 (which isn't released yet) and still require 5.6. But I think we could at least bump the requirement to 5.1, as we're not targeting support for older versions. But I don't feel strongly. Please go ahead and amend the PR as you see fit.

felixarntz:
@westonruter Makes sense. I bumped minimum WordPress to 5.1 for now, let's bump it to 5.2 once that is out.

PR #165, opened 2019-06-03
Prevent optimizer plugins from erroneously stripping HTML comments from JS

### LINKED ISSUE #155: Errors with service worker js
I'm having a hard time getting service workers to be registered and work. Seems like in the chrome console log there is an a js error in the generated service work

```
?wp_service_worker=1:129 Uncaught SyntaxError: missing ) after argument list
```

and
 
```
(index):1 Uncaught (in promise) TypeError: Failed to register a ServiceWorker: ServiceWorker script evaluation failed
```

and when using Lighthouse extension in chrome, I get the following errors
- Current page does not respond with a 200 when offline
- start_url does not respond with a 200 when offline. Unable to fetch start URL via service worker.
- Does not register a service worker that controls page and start_url

on my local dev environment, I get non of these issues and its a very similar install. the one thing that does happen in my local dev environment is if the plugin is installed and I add the theme support, the page continually refreshes as the service works are working and updating non-stop.

I'm not really using many of the functions as I'm experimenting.

```php
add_theme_support( 'service_worker', true );

/**
 * WEB APP Manifest Filters
 *
 * @param array $manifest
 * @return array $manifest
 */
add_filter('web_app_manifest', 'wundertheme_web_app_manifest_filter', 1, 1);
function wundertheme_web_app_manifest_filter($manifest)
{
    $manifest['theme_color'] = '#D9422B';
    $manifest['short_name']  = 'Mo Hamad';
    $manifest['orientation'] = "any";
    $manifest['display']     = "standalone";
    return $manifest;
}

/**
 * Adds an HSTS header to the response.
 *
 * @param array $headers The headers to filter.
 * @return array $headers The filtered headers.
 */
add_filter( 'wp_headers', function( $headers ) {
    $headers['Strict-Transport-Security'] = 'max-age=3600'; // Or another max-age.
    return $headers;
});

wp_register_service_worker_caching_route(
    '/.*\.(?:png|gif|jpg|jpeg|svg|webp|css|js)(\?.*)?$',
        array(
            'cacheName' => 'assets',
            'plugins'   => array(
                'expiration'        => array(
                    'maxEntries'    => 60,
                    'maxAgeSeconds' => 60 * 60 * 24,
            ),
        ),
    )
);
```

if anyone is so inclined, you can check out the lighthouse results for the site I'm working on
https://www.mohamed-hamad.com/

I'm also testing with https://www.pwabuilder.com, and their tool picks up everything but the service worker.

is there a tutorial on how to get the basics in, or an article that goes through the options and how to set it up?

westonruter:
Plugins which try to optimize HTML output by stripping HTML comments can erroneously strip them from JS logic in the service worker. This should prevent that from happening.

Fixes #155.

### DIFF
--- a/wp-includes/js/service-worker-navigation-routing.js
+++ b/wp-includes/js/service-worker-navigation-routing.js
@@ -65,9 +65,9 @@ ERROR_OFFLINE_BODY_FRAGMENT_URL, STREAM_HEADER_FRAGMENT_QUERY_VAR, NAVIGATION_BL
 							headers: errorResponse.headers
 						};
 
-						let body = text.replace( /<!--WP_SERVICE_WORKER_ERROR_MESSAGE-->/, errorMessages.error );
+						let body = text.replace( /[<]!--WP_SERVICE_WORKER_ERROR_MESSAGE-->/, errorMessages.error );
 						body = body.replace(
-							/(<!--WP_SERVICE_WORKER_ERROR_TEMPLATE_BEGIN-->)((?:.|\n)+?)(<!--WP_SERVICE_WORKER_ERROR_TEMPLATE_END-->)/,
+							/([<]!--WP_SERVICE_WORKER_ERROR_TEMPLATE_BEGIN-->)((?:.|\n)+?)([<]!--WP_SERVICE_WORKER_ERROR_TEMPLATE_END-->)/,
 							( details ) => {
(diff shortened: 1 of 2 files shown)


westonruter:
Fix confirmed in https://github.com/xwp/pwa-wp/issues/155#issuecomment-498466681

PR #173, opened 2019-06-11
Replace WP_DEBUG with more appropriate flags

### LINKED ISSUE #170: Workbox debug messages clutter console
### Problem

Workbox's `debug` mode is currently tied to `WP_DEBUG`:

https://github.com/xwp/pwa-wp/blob/a4bcf446cda9f6e308da356eb665668bb9c5831f/wp-includes/components/class-wp-service-worker-configuration-component.php#L69-L70

The messages from Workbox are so noisy that they clutter the log to the point where it's very distracting, and almost unusable for me. That's still the case with the 700+ `console.debug` messages filtered off, so that only `console.log` messages are being displayed.

I think the devex here should conform to the same principle that the UX would: under normal circumstances, nothing should be output to the console. For developers, having `WP_DEBUG` enabled is a normal circumstance. I should only see Workbox logging when I explicitly enable it, in order to debug something.

### Potential Solution

Ideally, I don't think this should be tied to `WP_DEBUG` at all, but should have a separate constant like `WORKBOX_DEBUG`.

At the very least, though, there should be a filter around `$options` so that developers can disable that for themselves.

I haven't thought a lot about the proper way to solve this, though, so I'm not strongly tied to either of those, they were just the first ideas that came to mind.

westonruter:
Fixes #170.

* Use `SCRIPT_DEBUG` to indicate that non-production sources of Workbox should be used.
* Look at `PWA_VERSION` instead of `WP_DEBUG` to determine if it is a dev version of the plugin.

By default the Workbox dev logs are disabled. To enable them, define this constant in the `wp-config.php`:

```php
define( 'WP_SERVICE_WORKER_DEBUG_LOG', true );
```

This only has any effect when `SCRIPT_DEBUG` is also enabled, as it causes the Workbox dev scripts to be used. When constant is present, the service worker will have the `__WB_DISABLE_DEV_LOGS` variable set which causes Workbox to not output the noisy logging messages. For more on that, https://github.com/GoogleChrome/workbox/issues/1619#issuecomment-557105799.

### DIFF
--- a/wp-includes/components/class-wp-service-worker-configuration-component.php
+++ b/wp-includes/components/class-wp-service-worker-configuration-component.php
@@ -68,8 +68,8 @@ class WP_Service_Worker_Configuration_Component implements WP_Service_Worker_Com
 			);
 		} else {
 			// Inline the workbox-sw.js to avoid an additional HTTP request.
-			$script .= file_get_contents( PWA_PLUGIN_DIR . '/' . $workbox_dir . 'workbox-sw.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
-			$script .= preg_replace( '://# sourceMappingURL=.+?\.map:', '', $script );
+			$wbjs    = file_get_contents( PWA_PLUGIN_DIR . '/' . $workbox_dir . 'workbox-sw.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+			$script .= preg_replace( '://# sourceMappingURL=.+?\.map\s*$:s', '', $wbjs );
 		}
 
 		$options = array(
--- a/wp-includes/components/class-wp-service-worker-configuration-component.php
+++ b/wp-includes/components/class-wp-service-worker-configuration-component.php
@@ -54,16 +54,22 @@ class WP_Service_Worker_Configuration_Component implements WP_Service_Worker_Com
 		$current_scope = wp_service_workers()->get_current_scope();
 		$workbox_dir   = 'wp-includes/js/workbox/';
 
+		$script = '';
 		if ( SCRIPT_DEBUG ) {
+			$enable_debug_log = defined( 'WP_SERVICE_WORKER_DEBUG_LOG' ) && WP_SERVICE_WORKER_DEBUG_LOG;
+			if ( ! $enable_debug_log ) {
+				$script .= "self.__WB_DISABLE_DEV_LOGS = true;\n";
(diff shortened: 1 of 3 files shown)


wp-includes/components/class-wp-service-worker-configuration-component.php
```diff
@@ -54,20 +54,26 @@ public function get_script() {
...
+			$script .= sprintf(
 				"importScripts( %s );\n",
 				wp_service_worker_json_encode( PWA_PLUGIN_URL . $workbox_dir . 'workbox-sw.js' )
 			);
 		} else {
 			// Inline the workbox-sw.js to avoid an additional HTTP request.
-			$script = file_get_contents( PWA_PLUGIN_DIR . '/' . $workbox_dir . 'workbox-sw.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
-			$script = preg_replace( '://# sourceMappingURL=.+?\.map:', '', $script );
+			$script .= file_get_contents( PWA_PLUGIN_DIR . '/' . $workbox_dir . 'workbox-sw.js' ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
+			$script .= preg_replace( '://# sourceMappingURL=.+?\.map:', '', $script );
```
felixarntz: This probably shouldn't be appended - likely an oversight :)

wp-includes/components/class-wp-service-worker-configuration-component.php
```diff
@@ -54,20 +54,26 @@ public function get_script() {
 		$current_scope = wp_service_workers()->get_current_scope();
 		$workbox_dir   = 'wp-includes/js/workbox/';
 
-		if ( WP_DEBUG ) {
+		$script = '';
+		if ( SCRIPT_DEBUG ) {
+			$enable_debug_log = defined( 'WP_SERVICE_WORKER_DEBUG_LOG' ) && WP_SERVICE_WORKER_DEBUG_LOG;
+			if ( ! $enable_debug_log ) {
+				$script .= "self.__WB_DISABLE_DEV_LOGS = true;\n";
+			}
```
felixarntz: 👍

wp-includes/components/class-wp-service-worker-configuration-component.php
westonruter: Good catch. Fixed in 22605b45123c3dcbb43b610b50a1b0a7e572ec32.

iandunn:
That approach sounds good to me, thanks!

PR #179, opened 2019-06-27
Fix trailing slash

nuvoPoint:
Fixes the issue where a site has a trailing slash in the home_url.

### DIFF
--- a/wp-includes/class-wp-web-app-manifest.php
+++ b/wp-includes/class-wp-web-app-manifest.php
@@ -119,7 +119,7 @@ class WP_Web_App_Manifest {
 	public function get_manifest() {
 		$manifest = array(
 			'name'      => wp_kses_decode_entities( get_bloginfo( 'name' ) ),
-			'start_url' => get_home_url(),
+			'start_url' => user_trailingslashit( get_home_url() ),
 			'display'   => 'minimal-ui',
 			'dir'       => is_rtl() ? 'rtl' : 'ltr',
 		);
--- a/wp-includes/class-wp-web-app-manifest.php
+++ b/wp-includes/class-wp-web-app-manifest.php
@@ -119,7 +119,7 @@ class WP_Web_App_Manifest {
 	public function get_manifest() {
 		$manifest = array(
 			'name'      => wp_kses_decode_entities( get_bloginfo( 'name' ) ),
-			'start_url' => user_trailingslashit( get_home_url() ),
+			'start_url' => home_url( '/' ),
 			'display'   => 'minimal-ui',
 			'dir'       => is_rtl() ? 'rtl' : 'ltr',
 		);
--- a/tests/test-class-wp-web-app-manifest.php
+++ b/tests/test-class-wp-web-app-manifest.php
@@ -143,7 +143,7 @@ class Test_WP_Web_App_Manifest extends WP_UnitTestCase {
 			'short_name'       => $short_name_matches[0],
 			'lang'             => get_bloginfo( 'language' ),
 			'dir'              => is_rtl() ? 'rtl' : 'ltr',
-			'start_url'        => get_home_url(),
+			'start_url'        => home_url( '/' ),
 			'theme_color'      => WP_Web_App_Manifest::FALLBACK_THEME_COLOR,
 			'icons'            => $this->instance->get_icons(),
 		);


wp-includes/class-wp-web-app-manifest.php
```diff
@@ -119,7 +119,7 @@ public function get_theme_color() {
 	public function get_manifest() {
 		$manifest = array(
 			'name'      => wp_kses_decode_entities( get_bloginfo( 'name' ) ),
-			'start_url' => get_home_url(),
+			'start_url' => user_trailingslashit( get_home_url() ),
```
westonruter: Shouldn't this be `trailingslashit()`?

nuvoPoint: I have tested it with and without a trailing slash in the home url and as I understand user_trailingslashit documentation it will check if the permalink settings has a trailing slash.

"Conditionally adds a trailing slash if the permalink structure has a trailing slash, strips the trailing slash if not."

https://codex.wordpress.org/Function_Reference/user_trailingslashit

westonruter:
What is the underlying issue being fixed?

nuvoPoint:
get_home_url() returns the home url without a trailing slash and if the site has a trailing slash it will result in the service worker having a wrong scope.

westonruter:
Are you working with a subdirectory install? What are the specific URLs you are working with?

nuvoPoint:
Yes Sir, it's a multi site installation with subdirectory setup, where each language/country has urls like https://www.site.com/en/ and https://www.site.com/da/

westonruter:
The intended behavior is for each site on a multisite install to get its own service worker installed. You're saying that if you go to the English site then the scope is wrong. Is it `/en`? But then why is your change for the `start_url` and not the scope?

nuvoPoint:
If I do a lighthouse audit I can see the following issue:

> Does not register a service worker that controls page and start_url
> This page is controlled by a service worker, however the start_url ("https://localhost/en") is not in the service worker's scope ("https://localhost/en/").

There is no issues with the fix and each site will have its own service worker.

westonruter:
Ok, so the issue is that the start URL needs to be trailing slashed to match the scope, right?

In that case, `trailingslashit` should be used to unconditionally force the slash.

nuvoPoint:
Correct, but if any other site has a setup without a trailing slash in the home_url then there might be a issue for them eg. https://www.site.com.

What user_trailingslashit ensures is that if the permalinks config is setup to have a trailing slash it keeps it otherwise it will remove it.

westonruter:
But if the `start_url` is `/en` but the scope is `/en/` then we need to make sure that the start URL is in the scope.

nuvoPoint:
I have tried to look into the code for where the scope it getting set and it seems it in wp_get_service_worker_url and the home_url( '/' ) call.

So we could force set a trailing slash for the start_url, but wouldn't it be better to use user_trailingslashit( get_home_url() ) in both places and let the users permalinks config handle that decision?

nuvoPoint:
@westonruter Is there anything you need me to do?

westonruter:
@nuvoPoint Please share specifically how you have configured your site so I can reproduce your setup. You have multisite subdirectory install, but you have turned off trailing slashes?

westonruter:
> Please share specifically how you have configured your site so I can reproduce your setup. You have multisite subdirectory install, but you have turned off trailing slashes?

@nuvoPoint ☝️

nuvoPoint:
Sorry, I did not see your last comment.

On my localhost it's setup on NginX with nothing more than:

And in the wp-config:

```
server {
    listen 80 default_server;
    listen 443 ssl default_server;
    server_name _ ;
    root "C:/laragon/www/";
    
    index index.html index.htm index.php;
 
    # Access Restrictions
    allow       127.0.0.1;
    deny        all;
    # allow 192.168.0.0/24;
 
    include "C:/laragon/etc/nginx/alias/*.conf";

    location / {
        #try_files $uri $uri/ =404;
		try_files $uri $uri/ /index.php?$args;
		autoindex on;
    }
    
    location ~ \.php$ {
        include snippets/fastcgi-php.conf;
        fastcgi_pass php_upstream;		
        #fastcgi_pass unix:/run/php/php7.0-fpm.sock;
    }
    
    # Enable SSL
    ssl_certificate "C:/laragon/etc/ssl/laragon.crt";
    ssl_certificate_key "C:/laragon/etc/ssl/laragon.key";
    ssl_session_timeout 5m;
    ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
    ssl_ciphers ALL:!ADH:!EXPORT56:RC4+RSA:+HIGH:+MEDIUM:+LOW:+SSLv3:+EXP;
    ssl_prefer_server_ciphers on;
	
	
    charset utf-8;
	
    location = /favicon.ico { access_log off; log_not_found off; }
    location = /robots.txt  { access_log off; log_not_found off; }
    location ~ /\.ht {
        deny all;
    }
	
	# Adds support for WP Multisites, where the setup is as subfolders.
	# Must be last, as the rewrites break if statements if placed before.
	if (!-e $request_filename) {
	  # Adds a trailing slash to /wp-admin
	  rewrite /wp-admin$ $scheme://$host$uri/ permanent;
	  # Rewrites so that /our-site/wp-{admin/content/includes} works.
	  rewrite ^/[_0-9a-zA-Z-]+(/wp-.*) $1 last;
	  # Same as the above, but for php files.
	  rewrite ^/[_0-9a-zA-Z-]+(/.*\.php)$ $1 last;
	}
}
```

And in wp-config:
```
define('MULTISITE', true);
define('SUBDOMAIN_INSTALL', false);
define('DOMAIN_CURRENT_SITE', 'localhost');
define('PATH_CURRENT_SITE', '/');
define('SITE_ID_CURRENT_SITE', 1);
define('BLOG_ID_CURRENT_SITE', 1);

$_SERVER['HTTPS'] = 'on';
if ( ! defined( 'WP_CLI' ) ) {
	define( 'COOKIE_DOMAIN', $_SERVER['HTTP_HOST'] );
}
define( 'FORCE_SSL_ADMIN', true );
```

Our live servers have a bit different setup on NginX, but with the same results.

Please let me know if your need anything else?

wp-includes/class-wp-web-app-manifest.php
westonruter: What about instead just doing. 

```php
'start_url' => home_url( '/' ),
```

This is what core does to get the URL to the front page: https://github.com/WordPress/wordpress-develop/blob/a8a4c09f33356475f4a3f3a9f5bdacd0c25bcb42/src/wp-includes/link-template.php#L332-L333

westonruter:
@nuvoPoint See my comment about suggesting `home_url( '/' )` to follow the pattern of how core does it. This will match the scope which is done via `home_url( '/', 'relative' )`. So both are trailing-slashed by default. It seems the front page in WordPress always will get trailingslashed, unless extraordinary efforts are done to remove the slash.

There is an open issue in the service worker spec regarding `scope` resolution for URLs that don't end in slashes: https://github.com/w3c/ServiceWorker/issues/1272

See this related issue I just opened: https://github.com/xwp/pwa-wp/issues/203

nuvoPoint:
Forcing the start_url to have a slash with home_url( '/' ) will work in our setup and fully acceptable for me. But it will might not work for people who has made "extraordinary efforts" to remove the slash, which user_trailingslashit( get_home_url() ) should take care of.

westonruter:
I think this is the best option until https://github.com/w3c/ServiceWorker/issues/1272 is resolved.

nuvoPoint:
That's fine for us - thanks! 
Do you need us to do anything or are you just updating it?

westonruter:
@nuvoPoint I just updated it. Can you confirm the change works now for you?

nuvoPoint:
I can confirm that it works!

PR #207, opened 2019-08-02
Ensure $current_screen global is set before applying admin_title filter

### LINKED ISSUE #206: admin_title filter called without ensuring current_screen is set
In /wp-admin/error.php, the `admin_title` filter is used without ensuring that the current_screen is set: https://github.com/xwp/pwa-wp/blob/7cf111f43dff028d6331009eabdc25848e6cd44f/wp-admin/error.php#L45

admin-header.php does this with https://github.com/WordPress/WordPress/blob/master/wp-admin/admin-header.php#L29:L32

before using the filter in https://github.com/WordPress/WordPress/blob/master/wp-admin/admin-header.php#L68.

This is causing the fatal error reported in https://github.com/woocommerce/woocommerce-admin/issues/2732 .

westonruter:
Fixes #206.

### DIFF
--- a/wp-admin/error.php
+++ b/wp-admin/error.php
@@ -41,6 +41,12 @@ $admin_title = get_bloginfo( 'name' );
 /* translators: Admin screen title. 1: Admin screen name, 2: Network or site name */
 $admin_title = sprintf( __( '%1$s &lsaquo; %2$s &#8212; WordPress', 'default' ), $title_prefix, $admin_title );
 
+// Ensure that $current_screen is set for admin_title to apply.
+global $current_screen;
+if ( empty( $current_screen ) ) {
+	set_current_screen();
+}
+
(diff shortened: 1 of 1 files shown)


nuvoPoint:
We have implemented this fix and are keeping an eye on the log files for the next couple of hours to see if this fixes #206

nuvoPoint:
We haven't see the fatal error yet in our logs, but now we are getting the following:

`PHP Notice: Undefined index: hook_suffix in /var/www/html/wp-admin/includes/class-wp-screen.php on line 213`

The only two things changed in our setup is this fix and #179

westonruter:
@nuvoPoint Thanks for that. I've also set `$hook_suffix` to address this in ad7adab.

nuvoPoint:
It seems to work fine now with $hook_suffix - thanks!

PR #242, opened 2020-02-04
get_validated_file_path should accept false

szepeviktor:
Discovered by @phpstan on Level 4

false may come from https://developer.wordpress.org/reference/functions/get_header_image/

### DIFF
--- a/wp-includes/class-wp-service-worker-scripts.php
+++ b/wp-includes/class-wp-service-worker-scripts.php
@@ -181,7 +181,7 @@ class WP_Service_Worker_Scripts extends WP_Scripts implements WP_Service_Worker_
 	/**
 	 * Get validated path to file.
 	 *
-	 * @param string|false $url Relative path or false.
+	 * @param string $url Relative path or false.
 	 * @return string|WP_Error
 	 */
 	public function get_validated_file_path( $url ) {
--- a/wp-includes/class-wp-service-worker-scripts.php
+++ b/wp-includes/class-wp-service-worker-scripts.php
@@ -181,7 +181,7 @@ class WP_Service_Worker_Scripts extends WP_Scripts implements WP_Service_Worker_
 	/**
 	 * Get validated path to file.
 	 *
-	 * @param string $url Relative path or false.
+	 * @param string $url Relative path.
 	 * @return string|WP_Error
 	 */
 	public function get_validated_file_path( $url ) {
--- a/wp-includes/class-wp-service-worker-scripts.php
+++ b/wp-includes/class-wp-service-worker-scripts.php
@@ -181,7 +181,7 @@ class WP_Service_Worker_Scripts extends WP_Scripts implements WP_Service_Worker_
 	/**
 	 * Get validated path to file.
 	 *
-	 * @param string $url Relative path.
+	 * @param string|false $url Relative path or false.
 	 * @return string|WP_Error
 	 */
 	public function get_validated_file_path( $url ) {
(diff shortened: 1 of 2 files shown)


westonruter:
Actually, a boolean cannot ever be passed in, as its only reference is:

https://github.com/GoogleChromeLabs/pwa-wp/blob/f6993dfb34e93eafe119a7a51770d1c501909a27/wp-includes/class-wp-service-worker-scripts.php#L160-L161

If we do allow booleans, then the logic should probably be hardened here:

https://github.com/GoogleChromeLabs/pwa-wp/blob/f6993dfb34e93eafe119a7a51770d1c501909a27/wp-includes/class-wp-service-worker-scripts.php#L187-L197

Perhaps it should return a `WP_Error` if a `bool` is passed in?

szepeviktor:
The mentioned `get_header_image` is about this line:

https://github.com/GoogleChromeLabs/pwa-wp/blob/351eeeca0ef67e92c8db3e2f1cbf609572faeb1f/integrations/class-wp-service-worker-custom-header-integration.php#L45-L47

Core docs say it could return a boolean.
Am I wrong?

westonruter:
@szepeviktor I think I got issues confused. The summary of this issue is for “get_validated_file_path accepts false”. Is that intended? Or how is this related to `get_header_image`?

szepeviktor:
Currently `$scripts->get_validated_file_path( get_header_image() );` false may be fed into `get_validated_file_path` but it does not accept it.

This PR makes `get_validated_file_path` accept a boolean.

szepeviktor:
@westonruter Excuse me for the naming confusion.

westonruter:
@szepeviktor OK, I think a different approach then is needed to just be explicit that a string is only ever passed into this function. For example:

```diff
diff --git a/integrations/class-wp-service-worker-custom-header-integration.php b/integrations/class-wp-service-worker-custom-header-integration.php
index 6f25556..a4d1de6 100644
--- a/integrations/class-wp-service-worker-custom-header-integration.php
+++ b/integrations/class-wp-service-worker-custom-header-integration.php
@@ -54,16 +54,15 @@ class WP_Service_Worker_Custom_Header_Integration extends WP_Service_Worker_Base
 			return;
 		}
 
-		$header     = get_custom_header();
-		$attachment = ! empty( $header->attachment_id ) ? get_post( get_custom_header()->attachment_id ) : null;
-
+		$header       = get_custom_header();
+		$attachment   = ! empty( $header->attachment_id ) ? get_post( get_custom_header()->attachment_id ) : null;
+		$header_image = get_header_image();
 		if ( $attachment ) {
 			foreach ( $this->get_attachment_image_urls( $attachment->ID, array( $header->width, $header->height ) ) as $image_url ) {
 				$scripts->precaching_routes()->register( $image_url, $attachment->post_modified );
 			}
-		} elseif ( get_header_image() ) {
-			$url      = get_header_image();
-			$file     = $scripts->get_validated_file_path( get_header_image() );
+		} elseif ( is_string( $header_image ) ) {
+			$file     = $scripts->get_validated_file_path( $header_image );
 			$revision = null;
 			if ( is_string( $file ) ) {
 				$revision = md5( file_get_contents( $file ) ); // phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
```

szepeviktor:
Done.

westonruter:
Please try merging the latest from `master` to address the build failure.

wp-includes/class-wp-service-worker-scripts.php
```diff
@@ -181,7 +181,7 @@ public function do_item( $handle, $group = false ) {
 	/**
 	 * Get validated path to file.
 	 *
-	 * @param string $url Relative path.
+	 * @param string $url Relative path or false.
```
westonruter: This change can be reverted now that the `is_string()` check is being used to verify that `$header_image` is _not_ false, no?

```suggestion
	 * @param string $url Relative path.
```

szepeviktor: Done :)

PR #279, opened 2020-04-26
Prevent service worker from varying by authentication state (and prevent auto-reloading)

### LINKED ISSUE #166: Website keeps reloading after installing PWA
After installing PWA my [website ](https://oglasevanjenaspletu.si/)keeps on reloading in some cases. 

I'm using OneSignal but is currently deactivated. 

I also just installed Autoptimize & autoptimize criticalCSS and converted all images to WebP. But I think this isn't an issue.

### LINKED ISSUE #202: Firefox Continually Refreshing Pages
I've noticed that Firefox is continually refreshing on various pages of my site (after the "Allow Notifications" prompt is shown, but it refreshes before the user can really act on that prompt). I'm thinking it's tied to the
```js
let refreshedPage = false;
navigator.serviceWorker.addEventListener( 'controllerchange', () => {
	if ( ! refreshedPage ) {
		refreshedPage = true;
		window.location.reload();
	}
} );
```
snippet added to the page by PWA.

Oddly, the homepage is pretty much the same as the other pages having the refresh issue (also having this same code snippet), but it's happily not refreshing again & again with the "Allow Notifications" prompt to be acted on.

Any idea of what might be the fix for this?

### LINKED ISSUE #205: Discontinue skip waiting by default (and automatically reloading) when service worker updates
Currently when an updated service worker is installed, the current page will reload automatically. This can be disorienting and it can even cause data to be lost, if the user starts entering data while the updated service worker is being installed.

In other words, we should discontinue skipWaiting by default. We could potentially show the service worker update button in the space where the admin bar would be displayed normally, if the admin bar is not already on the page. So we can reuse the admin bar area and class names and styles to show notices related to PWA, including offline status, new version of page available, refresh app, etc.

For more see https://github.com/xwp/pwa-wp/issues/166#issuecomment-499166929.

### LINKED ISSUE #223: Admin panel not loading pages correctly when plugin activated
Hi there,

Thank you for this plugin. Having the ability to register service workers within Wordpress is sorely needed and I'm very thankful that this plugin helps us with that.

I've recently noticed that some of our admin pages do not load when this plugin is activated. Instead, the browser is served the current page.

I'm not totally sure what's happening. The URL sometimes changes to the new, expected URL. Other times the browser refreshes as if the new page is about to load but the existing page is returned instead with the original URL.

All of this is a little vague but I'm hoping that someone might have seen similar behaviour before and knows what the issue might be.

If I check the "Bypass for network" checkbox in Chrome Dev tools, everything goes back to working correctly. Similarly if I disable the PWA plugin, everything works as expected again.

Our use case for the PWA plugin is to load a custom service worker. At the moment, the plugin that creates that service worker is also disabled, so shouldn't be having any impact on this issue.

Thank you for any help and I hope I've got the right place for support. If this needs to be a StackOverflow question please let me know.

Again, thank you for all the work going into this plugin.

Wordpress Version: 5.2.2
PWA Plugin Version: 0.3.0

Here's a gif demonstrating what happens when I'm on the Dashboard and I click the Plugins menu item. The page refreshes but the Dashboard is displayed again instead of the Plugins page.

AScreen Recording 2019-09-30 at 13 54 31

**Update**

If I hit the plugins url directly, I'm redirected back to `wp-admin/index.php`.

**Update 2**

It only seems to be doing this with the PWA plugin activated and Chrome Dev tools open. Closing the Dev Tools, or disabling the PWA plugin resolves the issue.

### LINKED ISSUE #249: When “Update on reload” is checked, navigation blocked
When Chrome DevTools is open and “Update on reload” is checked in the Application tab:

> image

Clicking around the site will result in the initial page just reloading. The cause is this code:

https://github.com/GoogleChromeLabs/pwa-wp/blob/9bd27f3ecf65416742ff3f5aa9d5ed6e739a1608/wp-includes/service-workers.php#L207

In reality, we should no longer be automatically reloading when the service worker updates. It just causes problems. See #205.

westonruter:
This PR does a few things to fix some long-standing annoyances.

1. When an updated service worker is activated, the window will not automatically reload.
2. As soon as a service worker is installed, clients are now claimed by default; this makes sense to do since WordPress pages are not (normally) single-page-applications. In other words, a page accessed without a service worker is the same as a page served with a service worker (it's just the latter may be served from cache). If `clients.claim()` is not desired, it can be disabled via `add_filter( 'wp_service_worker_clients_claim', '__return_false' )`. This seems the best default for at least the 80% use case.
3. Eliminate the “update site” link which was (rarely) shown. It was only shown when skip-waiting was disabled and if the admin bar was showing. Now skip-waiting is enabled and no reloading is being done, there is no need for this additional button. If a site wants to disable skip-waiting and add such a button, they can still do so. Again, it's removed since it's not deemed part in the 80% use case.
4. Prevent the error templates for offline and 500 from being served with logged-in user state. Site owners should expect that that user-specific information should not be displayed on this page. This prevents the need for the service worker to update once logging-in or logging-out. 
5. In the same way, the service workers are also served with the user authentication being cleared to avoid serving user-specific information in the service worker. This also guards against the service worker varying by authentication state. This is somewhat similar to how the REST API behaves. Nevertheless, this may end up being an undesirable and we'll need to watch this for use cases. It could be useful to serve some auth key (e.g. REST API nonce) in the service worker, but this would end up being a pain as well once the key expires.

Fixes #205.
Fixes #166.
Fixes #202.
Fixes #249.
Fixes #223.

### DIFF
--- a/wp-includes/class-wp-service-workers.php
+++ b/wp-includes/class-wp-service-workers.php
@@ -91,6 +91,10 @@ class WP_Service_Workers implements WP_Service_Worker_Registry_Aware {
 	 * @see wp_service_worker_loaded()
 	 */
 	public function serve_request() {
+		// Clear the currently-authenticated user to ensure that the service worker doesn't vary between users.
+		// @todo Would it be better to use the determine_current_user filter for this?
+		wp_set_current_user( 0 );
+
 		// See wp_debug_mode() for how this is also done for REST API responses.
 		@ini_set( 'display_errors', 0 ); // phpcs:ignore Generic.PHP.NoSilencedErrors.Discouraged, WordPress.PHP.NoSilencedErrors.Discouraged, WordPress.PHP.DiscouragedPHPFunctions.runtime_configuration_ini_set, WordPress.PHP.IniSet.display_errors_Blacklisted
 
--- a/wp-includes/class-wp-service-workers.php
+++ b/wp-includes/class-wp-service-workers.php
@@ -91,8 +91,14 @@ class WP_Service_Workers implements WP_Service_Worker_Registry_Aware {
 	 * @see wp_service_worker_loaded()
 	 */
 	public function serve_request() {
-		// Clear the currently-authenticated user to ensure that the service worker doesn't vary between users.
-		// @todo Would it be better to use the determine_current_user filter for this?
+		/*
+		 * Clear the currently-authenticated user to ensure that the service worker doesn't vary between users.
+		 * Note that clearing the authenticated user in this way is in keeping with REST API requests wherein the
+		 * WP_REST_Server::serve_request() method calls WP_REST_Server::check_authentication() which in turn applies
+		 * the rest_authentication_errors filter which runs rest_cookie_check_errors() which is then responsible for
+		 * calling wp_set_current_user( 0 ) if it was previously-determined a user was logged-in with the required
+		 * nonce cookie set when wp_validate_auth_cookie() triggers one of the auth_cookie_* actions.
+		 */
 		wp_set_current_user( 0 );
 
 		// See wp_debug_mode() for how this is also done for REST API responses.
--- a/wp-includes/general-template.php
+++ b/wp-includes/general-template.php
@@ -101,13 +101,16 @@ function wp_add_error_template_no_robots() {
 }
 
 /**
- * Hide the admin bar if serving the offline template.
+ * Ensure current user is unset when serving an error page (either offline or server error).
  *
- * @since 0.2
+ * This is important so that what the service worker initially precaches for the user when first accessing the site
+ * will persist even after they have authenticated.
+ *
+ * @since 0.4.1
  */
-function wp_hide_admin_bar_offline() {
-	if ( is_offline() ) {
-		show_admin_bar( false );
+function wp_unset_current_user_error_template() {
+	if ( is_offline() || is_500() ) {
+		wp_set_current_user( 0 );
 	}
 }
 
--- a/wp-includes/general-template.php
+++ b/wp-includes/general-template.php
@@ -108,7 +108,7 @@ function wp_add_error_template_no_robots() {
  *
  * @since 0.5
  */
-function wp_unset_current_user_error_template() {
+function wp_unauthenticate_error_template_requests() {
 	if ( is_offline() || is_500() ) {
 		wp_set_current_user( 0 );
 	}
--- a/wp-includes/general-template.php
+++ b/wp-includes/general-template.php
@@ -106,7 +106,7 @@ function wp_add_error_template_no_robots() {
  * This is important so that what the service worker initially precaches for the user when first accessing the site
  * will persist even after they have authenticated.
  *
- * @since 0.4.1
+ * @since 0.5
  */
 function wp_unset_current_user_error_template() {
 	if ( is_offline() || is_500() ) {
--- a/.phpcs.xml.dist
+++ b/.phpcs.xml.dist
@@ -11,6 +11,7 @@
 		https://github.com/squizlabs/PHP_CodeSniffer/wiki/Advanced-Usage#ignoring-files-and-folders -->
 	<exclude-pattern>./node_modules/</exclude-pattern>
 	<exclude-pattern>./vendor/</exclude-pattern>
+	<exclude-pattern>./build/</exclude-pattern>
 
 	<!-- How to scan -->
 	<arg value="sp"/> <!-- Show sniff and progress -->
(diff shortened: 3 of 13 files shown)


westonruter:
Build for testing: [pwa.zip](https://github.com/GoogleChromeLabs/pwa-wp/files/4536625/pwa.zip) (0.5.0-alpha-88c1314-20200426T224736Z)

wp-includes/general-template.php
```diff
@@ -101,13 +101,16 @@ function wp_add_error_template_no_robots() {
...
+ * @since 0.5
  */
-function wp_hide_admin_bar_offline() {
-	if ( is_offline() ) {
-		show_admin_bar( false );
+function wp_unset_current_user_error_template() {
+	if ( is_offline() || is_500() ) {
+		wp_set_current_user( 0 );
 	}
 }
```
felixarntz, line 115: Can we maybe bake this somehow as a method into `WP_Service_Worker_Navigation_Routing_Component` instead? It's a bit hard to digest what it relates to from here and why this is being done, and that class mainly deals with these two templates. It also contains the other change (remove user-specific affix for revision) that makes this here necessary.

wp-includes/class-wp-service-workers.php
```diff
@@ -91,6 +91,10 @@ public function get_current_scope() {
 	 * @see wp_service_worker_loaded()
 	 */
 	public function serve_request() {
+		// Clear the currently-authenticated user to ensure that the service worker doesn't vary between users.
+		// @todo Would it be better to use the determine_current_user filter for this?
+		wp_set_current_user( 0 );
```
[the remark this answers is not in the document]
westonruter: Yeah, that's a good point. I was trying to model this after `wp_service_worker_loaded()` after `rest_api_loaded()` and `WP_Service_Workers::serve_request()` after `WP_REST_Server::serve_request()`. When doing a REST request, the authentication is set outside of the `serve_request` method. Namely, it is done via `wp_validate_auth_cookie()` (in a round about way).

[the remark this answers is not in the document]
westonruter: Actually, `WP_REST_Server::serve_request()` calls `WP_REST_Server::check_authentication()` which applies the `rest_authentication_errors` filter which calls `rest_cookie_check_errors()`.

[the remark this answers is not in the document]
westonruter: @felixarntz I stepped through what core is doing and I came to the conclusion that calling `wp_set_current_user()` is most in-line with what core is doing presently with the REST API. I added an explanation in 182b4e25e1bff8acb2a69df28f20d871a07da06f.

An added benefit of not calling `die()` in this method after having cleared the current user is that we'd be able to test to confirm that the authenticated user is indeed cleared, which I've been able to just add a test for in 98c7e12.

wp-includes/general-template.php
westonruter: Humm. Thinking about this. The `WP_Service_Worker_Navigation_Routing_Component` class is about referencing the URL to the offline/500 page; it isn't involved in actually serving it. 

WordPress would serve these templates via `wp-includes/template-loader.php`, though at the moment the PWA plugin implements that via the `template_include` filter. What we need to do here is ensure that when that `template_include` logic is run to include the offline/500 template, the user is unset. The earliest point we can check `is_offline()` or `is_500()` is at the `parse_query` action, and that is when here `wp_unset_current_user_error_template()` is called.

So I think this is pretty good where it is here actually. 

I've added some tests in f0e5984.

felixarntz: Hmm, I see what you're saying related to serving the request. On the other hand, unsetting the user is necessary because of the logic in `WP_Service_Worker_Navigation_Routing_Component`. If it didn't provide the 500/offline templates to the service-worker for precaching, `wp_unauthenticate_error_template_requests` wouldn't be necessary.

westonruter: It wouldn't be necessary if `WP_Service_Worker_Navigation_Routing_Component` were able to construct service worker JS in a way that would prevent cookies from being sent in the request to precache the offline/500 page.

felixarntz: Makes sense. Hence my other general comment - maybe that is not technically possible, but if it was, I think it'd be a nicer solution.

westonruter:
That may open up other problems though. Namely, what if a site is behind some authentication layer that gates users by some cookie-based auth? If so, then the SW would fail to ever install properly.

westonruter:
So at the moment, `wp_unauthenticate_error_template_requests` hooking into `parse_query` to clear the authenticated user is happening at the same time as when `wp_service_worker_loaded` runs to also clear the user. In the former, `is_500()` and `is_offline()` conditions are used to clear the user (and serve the required template) and in the latter the `wp_service_worker` query var is used to clear the user (and serve the service worker JS response). This seems pretty good to me.

felixarntz:
Thanks for clarifying, makes sense to me now.

westonruter:
Thanks for the review and for challenging me!

PR #289, opened 2020-06-05
Add pretty permalinks for service worker

### LINKED ISSUE #287: Handling service worker query vars
Hey! Dan here from [Shifter](https://www.getshifter.io). ICYMI and for context, Shifter is a static site generator _for_ WordPress. We're long time users of the this plugin and it's the PWA plugin we recommend to our users.

We're running into an issue with the way the server-worker is handled now and our hosting platform is unable to read the query var required to load the environment.

https://github.com/GoogleChromeLabs/pwa-wp/blob/460ca974aa05c58bc6e00e4df84f9197e7862f4e/wp-includes/class-wp-service-workers.php#L23-L44

To use `?wp_service_worker=1` we need to run that PHP file. Since Shifter is all about static WordPress that PHP file will never run.

To add, we are also unable to determine the MIME type when hosting in our static site environment with that URL: `/?wp_service_worker=1`

Is there any way to remedy this with a filter or an option? It's not common for us to create MU plugins to apply common filters to our sites but I'm not sure where to go on this one.

I can follow-up with an example URL if that is helpful. Thanks!

swissspidy:
Fixes #287.

To-do:

- [x] Testing

### DIFF
--- a/pwa.php
+++ b/pwa.php
@@ -254,6 +254,28 @@ require_once PWA_PLUGIN_DIR . '/wp-includes/class-wp-query.php';
 require_once PWA_PLUGIN_DIR . '/wp-admin/admin.php';
 
 /**
+ * Plugin activation hook.
+ */
+function _pwa_activate_plugin() {
+	pwa_add_rewrite_rules();
+	flush_rewrite_rules( false );
+}
+
+register_activation_hook( PWA_PLUGIN_FILE, '_pwa_activate_plugin' );
+
+/**
+ * Plugin deactivation hook.
+ */
+function _pwa_deactivate_plugin() {
+	/* @var WP_Rewrite $wp_rewrite */
+	global $wp_rewrite;
+	unset( $wp_rewrite->extra_rules_top['^wp-service-worker\.js$'] );
+	flush_rewrite_rules( false );
+}
+
+register_deactivation_hook( PWA_PLUGIN_FILE, '_pwa_deactivate_plugin' );
+
+/**
  * Load service worker integrations.
  *
  * @since 0.2.0
--- a/pwa.php
+++ b/pwa.php
@@ -269,7 +269,7 @@ register_activation_hook( PWA_PLUGIN_FILE, '_pwa_activate_plugin' );
 function _pwa_deactivate_plugin() {
 	/* @var WP_Rewrite $wp_rewrite */
 	global $wp_rewrite;
-	unset( $wp_rewrite->extra_rules_top['^wp-service-worker\.js$'] );
+	unset( $wp_rewrite->extra_rules_top['^wp\.serviceworker$'] );
 	flush_rewrite_rules( false );
 }
 
--- a/wp-includes/class-wp.php
+++ b/wp-includes/class-wp.php
@@ -8,8 +8,18 @@
 
 /**
  * Adds rewrite rules to enable pretty permalinks for the service worker script.
+ * @global WP_Rewrite $wp_rewrite
  */
 function pwa_add_rewrite_rules() {
+	/* @var WP_Rewrite $wp_rewrite */
+	if ( ! isset( $wp_rewrite->extra_rules_top['^wp-service-worker\.js$'] ) ) {
+		add_action(
+			'admin_init', 
+			function () {
+				flush_rewrite_rules( false );
+			} 
+		);
+	}
 	add_rewrite_tag( '%' . WP_Service_Workers::QUERY_VAR . '%', '([^?]+)' );
 	add_rewrite_rule( '^wp-service-worker\.js$', 'index.php?' . WP_Service_Workers::QUERY_VAR . '=' . WP_Service_Workers::SCOPE_FRONT, 'top' );
 }
--- a/wp-includes/class-wp.php
+++ b/wp-includes/class-wp.php
@@ -8,20 +8,26 @@
 
 /**
  * Adds rewrite rules to enable pretty permalinks for the service worker script.
+ *
  * @global WP_Rewrite $wp_rewrite
  */
 function pwa_add_rewrite_rules() {
+	global $wp_rewrite;
+	$rewrite_rule_regex = '^wp-service-worker\.js$';
+
 	/* @var WP_Rewrite $wp_rewrite */
-	if ( ! isset( $wp_rewrite->extra_rules_top['^wp-service-worker\.js$'] ) ) {
+	if ( ! isset( $wp_rewrite->extra_rules_top[ $rewrite_rule_regex ] ) ) {
+		// Note: This logic will not be required as part of core merge since rewrite rules are flushed upon DB upgrade (as long as the DB version is bumped).
 		add_action(
-			'admin_init', 
+			'admin_init',
 			function () {
 				flush_rewrite_rules( false );
-			} 
+			}
 		);
 	}
+	add_rewrite_rule( $rewrite_rule_regex, 'index.php?' . WP_Service_Workers::QUERY_VAR . '=' . WP_Service_Workers::SCOPE_FRONT, 'top' );
+
 	add_rewrite_tag( '%' . WP_Service_Workers::QUERY_VAR . '%', '([^?]+)' );
-	add_rewrite_rule( '^wp-service-worker\.js$', 'index.php?' . WP_Service_Workers::QUERY_VAR . '=' . WP_Service_Workers::SCOPE_FRONT, 'top' );
 }
 
 add_action( 'init', 'pwa_add_rewrite_rules' );
--- a/wp-includes/class-wp.php
+++ b/wp-includes/class-wp.php
@@ -7,6 +7,16 @@
  */
 
 /**
+ * Adds rewrite rules to enable pretty permalinks for the service worker script.
+ */
+function pwa_add_rewrite_rules() {
+	add_rewrite_tag( '%' . WP_Service_Workers::QUERY_VAR . '%', '([^?]+)' );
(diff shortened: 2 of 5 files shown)


pwa.php
```diff
@@ -253,6 +253,28 @@ function _pwa_check_disabled_navigation_preload() {
...
 
+/**
+ * Plugin activation hook.
+ */
+function _pwa_activate_plugin() {
+	pwa_add_rewrite_rules();
+	flush_rewrite_rules( false );
+}
+
+register_activation_hook( PWA_PLUGIN_FILE, '_pwa_activate_plugin' );
```
westonruter, line 264: The flushing is only happening upon activation. What about on network-activted multisite?

westonruter: I've just inquired the same on wp-sitemaps: https://github.com/GoogleChromeLabs/wp-sitemaps/issues/205

swissspidy: Haven‘t really tested Multisite, will take a look.

swissspidy: https://jeremyfelt.com/2015/07/17/flushing-rewrite-rules-in-wordpress-multisite-for-fun-and-profit/ 🤯

swissspidy: cc @felixarntz for Multisite expertise.

felixarntz: Iterating through all sites like in the above post works for small enough networks, but quickly runs into scaling issues. I think the only scalable way for a plugin to deal with setup like this in multisite is to not use an activation hook, but do it conditionally on something like `admin_init`:
* A common approach is to add an option for the plugin's version (e.g. we have `googlesitekit_db_version` in Site Kit).
* If version in DB does not exist or is lower than the current plugin version, run according installation / migration logic.
* This ensures it's run for every site as necessary - caveat is only if somebody logs in. Of course it could also be run on `init`, but I'm generally wary of doing that because it would also potentially run during frontend loads and without an authenticated user.

For installing rewrite rules like here a simple boolean flag would work, but I'd recommend something version-based because it's always likely to need more logic in the future which then will become migrations from older versions.

westonruter: Currently in the XML Sitemaps plugin it is [flushing](https://github.com/GoogleChromeLabs/wp-sitemaps/blob/584efba4eb63e918bc691c92bc573bba8945f07b/inc/class-wp-sitemaps.php#L166-L170) when the `WP_SITEMAPS_REWRITE_VERSION` changes [at `wp_loaded`](https://github.com/GoogleChromeLabs/wp-sitemaps/blob/584efba4eb63e918bc691c92bc573bba8945f07b/inc/class-wp-sitemaps.php#L68). That should rather have run at `admin_init`, no? In any case, that won't be needed if `flush_rewrite_rules9) ` is added to `upgrade_550()` as you noted in https://github.com/GoogleChromeLabs/wp-sitemaps/issues/205#issuecomment-640115945. The same could be done in the future if and when PWA is merged into core.

In the meantime, however, we should amend this PR with the `admin_init` hook that conditionally flushes rewrite rules if need be. As opposed to storing the version in an option like the XML Sitemap plugin is doing, how about something like this:

```php
add_action( 'admin_init', function() {
	/* @var WP_Rewrite $wp_rewrite */
	global $wp_rewrite;
	if ( ! isset( $wp_rewrite->extra_rules_top['^wp-service-worker\.js$'] ) ) {
		flush_rewrite_rules( false );
	}
} );
```

westonruter:
Since this PR will result in an apparent JS file being served, let's add a comment to the response to make it clear that the file is dynamically-generated. So in the line after:

https://github.com/GoogleChromeLabs/pwa-wp/blob/330c76bd37bf9ef3655361ea50c3acdb8e9fce59/wp-includes/class-wp-service-workers.php#L146

Add something like:

```php
echo '/* ';
/* translators: %s is the WordPress action hook */
printf( __( 'Note: This JS file is dynamically-generated with PHP. To manipulate the contents of this file, use the "%s" action in WordPress.', 'pwa' ), is_admin() ? 'wp_admin_service_worker' : 'wp_front_service_worker' ) );
echo " /*\n\n";
```

swissspidy:
Sounds good.

Does this also need a second rewrite rule for the admin scope service worker? 🤔

westonruter:
No, the admin service worker is served via `admin-ajax.php`, so it wouldn't be subjected to the same issue here.

pwa.php
swissspidy: I don't think this will work since `$wp_rewrite->extra_rules_top['^wp-service-worker\.js$']` will always be set by `pwa_add_rewrite_rules` on `init`.

westonruter: Good point. In that case, perhaps `pwa_add_rewrite_rules()` should check to see if it actually added anything to the array, and if so, set a flag that can be accessed somewhere?

wp-includes/class-wp.php
```diff
@@ -6,6 +6,16 @@
  * @since 0.2
  */
 
+/**
+ * Adds rewrite rules to enable pretty permalinks for the service worker script.
+ */
+function pwa_add_rewrite_rules() {
```
westonruter: What about this:

```suggestion
 * Adds rewrite rules to enable pretty permalinks for the service worker script.
 * @global WP_Rewrite $wp_rewrite
 */
function pwa_add_rewrite_rules() {
	/* @var WP_Rewrite $wp_rewrite */
	if ( ! isset( $wp_rewrite->extra_rules_top['^wp-service-worker\.js$'] ) ) {
		add_action(
			'admin_init', 
			function () {
				flush_rewrite_rules( false );
			} 
		);
	}
```

swissspidy: That could work. Let's give it a try.

pwa.php
jeremyfelt: :wave: y'all!

FWIW, since that post was written I've had success with the method posted by Thorsten on this ticket: https://core.trac.wordpress.org/ticket/18450

I've used it on `init` when the rewrite rules were very important, but it could be used on `admin_init` as well.

westonruter: Thanks for the input! Yes, I'd say this qualifies as being very important since if the PWA plugin is network-activated, all the network sites would try to install the service worker. But if the rewrite rule is not present, then they'd fail to install.

Nevertheless, we're currently blocked on using rewrite rule, at least using `.js` file extension since various hosts will short-circuit such requests from going to WordPress to instead try to go straight to the filesystem: https://github.com/GoogleChromeLabs/pwa-wp/issues/287#issuecomment-642946205

wp-includes/class-wp.php
```diff
@@ -6,6 +6,32 @@
...
  */
 
+/**
+ * Adds rewrite rules to enable pretty permalinks for the service worker script.
+ *
+ * @global WP_Rewrite $wp_rewrite
+ */
+function pwa_add_rewrite_rules() {
+	global $wp_rewrite;
+	$rewrite_rule_regex = '^wp-service-worker\.js$';
```
westonruter:
```suggestion
	$rewrite_rule_regex = '^wp\.serviceworker$';
```

pwa.php
```diff
@@ -253,6 +253,28 @@ function _pwa_check_disabled_navigation_preload() {
...
+
+/**
+ * Plugin deactivation hook.
+ */
+function _pwa_deactivate_plugin() {
+	/* @var WP_Rewrite $wp_rewrite */
+	global $wp_rewrite;
+	unset( $wp_rewrite->extra_rules_top['^wp-service-worker\.js$'] );
```
westonruter:
```suggestion
	unset( $wp_rewrite->extra_rules_top['^wp\.serviceworker$'] );
```

westonruter (commented):
To avoid the problem of `wp-service-worker.js` bypassing PHP and being routed straight to the file system by web servers, how about we just use `wp.serviceworker`.

See https://github.com/GoogleChromeLabs/pwa-wp/issues/287#issuecomment-683195256.

wp-includes/service-workers.php
```diff
@@ -86,12 +86,19 @@ function wp_register_service_worker_caching_route( $route, $args = array() ) {
...
+	global $wp_rewrite;
+
 	if ( WP_Service_Workers::SCOPE_FRONT !== $scope && WP_Service_Workers::SCOPE_ADMIN !== $scope ) {
 		_doing_it_wrong( __FUNCTION__, esc_html__( 'Scope must be either WP_Service_Workers::SCOPE_FRONT or WP_Service_Workers::SCOPE_ADMIN.', 'pwa' ), '?' );
 		$scope = WP_Service_Workers::SCOPE_FRONT;
 	}
 
 	if ( WP_Service_Workers::SCOPE_FRONT === $scope ) {
+		if ( $wp_rewrite->using_permalinks() ) {
+			return home_url( '/wp-service-worker.js' );
```
swissspidy:
```suggestion
			return home_url( '/wp.serviceworker' );
```

pwa.php
westonruter: Now that we have `/wp.serviceworker` as the route, the only thing left here is making sure that the route will be detected across a network before rewrite rules have been flushed. 

Here's an idea: en lieu of waiting for someone to log-in to each site to cause rewrite rules to be flushed, we can implement a fallback route of sorts. 

I think I've got a good solution for that in 043beb8c29671cf3a32d40cf9d7bc5ce606a4884.

westonruter (approved):

westonruter:
This is working well according to https://github.com/GoogleChromeLabs/pwa-wp/issues/287#issuecomment-689308894. Any additional needed changes anyone sees prior to merging?

hideokamoto:
@westonruter 
Hi.
Has the PR published in wp.org?

I check wp.org today, but the version is still 0.5 ...

swissspidy:
@hideokamoto No, as you can see at https://github.com/GoogleChromeLabs/pwa-wp/milestone/6 version 0.6 is not yet complete.

hideokamoto:
@swissspidy 
Roger that.

Thanks !

PR #305, opened 2020-07-07
Add manifest-URL helper

ethitter:
Provide easier access to manifest URL without requiring as much knowledge of the plugin's architecture.


westonruter (approved):
Good idea.

PR #308, opened 2020-07-11
Allow HTTPS detection and UI to be disabled

ethitter:
In cases where HTTPS redirection is handled by the web host, or is otherwise outside of the control of WordPress, the UI and detection can be confusing or superfluous.

### DIFF
--- a/wp-includes/class-wp-https-detection.php
+++ b/wp-includes/class-wp-https-detection.php
@@ -39,6 +39,12 @@ class WP_HTTPS_Detection {
 	const INSECURE_CONTENT_OPTION_NAME = 'insecure_content';
 
 	/**
+	 * Filter tag to disable HTTPS detection and UI, such as when redirection is
+	 * handled outside of WordPress.
+	 */
+	const FILTER_DISABLE = 'wp_https_detection_ui_disabled';
+
+	/**
 	 * The tag names for insecure content.
 	 *
 	 * @see https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content
@@ -57,6 +63,10 @@ class WP_HTTPS_Detection {
 	 * Initializes the object.
 	 */
 	public function init() {
+		if ( apply_filters( self::FILTER_DISABLE, false ) ) {
+			return;
+		}
+
 		add_action( 'init', array( $this, 'schedule_cron' ) );
 		add_action( self::CRON_HOOK, array( $this, 'update_https_support_options' ) );
 		add_filter( 'cron_request', array( $this, 'conditionally_prevent_sslverify' ), PHP_INT_MAX );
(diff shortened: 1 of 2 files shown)


wp-includes/class-wp-https-ui.php
```diff
@@ -79,6 +79,10 @@ public function __construct( $wp_https_detection ) {
 	 * Initializes the object.
 	 */
 	public function init() {
+		if ( apply_filters( WP_HTTPS_Detection::FILTER_DISABLE, false ) ) {
+			return;
+		}
+
```
westonruter: Is this necessary here? The `WP_HTTPS_Detection::init()` method is what instantiates this class, so the filter there should suffice.

wp-includes/class-wp-https-detection.php
```diff
@@ -57,6 +63,10 @@ class WP_HTTPS_Detection {
 	 * Initializes the object.
 	 */
 	public function init() {
+		if ( apply_filters( self::FILTER_DISABLE, false ) ) {
```
westonruter: Core uses string literals in calls to `apply_filters()` and `do_action()` to facilitate generating documentation. And IDEs like PhpStorm parse strings in code indexing which allow for autocompletion among other things.

```suggestion
		/**
		 * Filters whether the HTTPS Detection and UI should be disabled.
		 *
		 * @param bool $disabled Disabled.
		 */
		if ( apply_filters( 'wp_https_detection_ui_disabled', false ) ) {
```

wp-includes/class-wp-https-detection.php
```diff
@@ -38,6 +38,12 @@ class WP_HTTPS_Detection {
 	 */
 	const INSECURE_CONTENT_OPTION_NAME = 'insecure_content';
 
+	/**
+	 * Filter tag to disable HTTPS detection and UI, such as when redirection is
+	 * handled outside of WordPress.
+	 */
+	const FILTER_DISABLE = 'wp_https_detection_ui_disabled';
+
```
westonruter:
```suggestion
```

westonruter (changes requested):

wp-includes/class-wp-https-ui.php
ethitter: Not strictly, I was being cautious/paranoid. If the `WP_HTTPS_Detection` class is disabled but something instantiated the `WP_HTTPS_UI` class, it would add UI that doesn't do anything (since the redirect relies on data from `WP_HTTPS_Detection`) and I wanted to avoid any potential confusion.

wp-includes/class-wp-https-detection.php
ethitter: Fair point. I hate repeating strings, but it makes sense for Core.

westonruter (approved):

PR #314, opened 2020-07-22
Update wording of the serverOffline message.

KZeni:
The serverOffline message was being shown even though the site was online & working perfectly fine while it was an issue with the client's connection (while they weren't necessarily considered offline.) As such, it should mention both possibilities so people don't instantly blame the website when that might not be the cause for the issue.

### DIFF
--- a/wp-includes/template.php
+++ b/wp-includes/template.php
@@ -144,7 +144,7 @@ function wp_service_worker_get_error_messages() {
 		'wp_service_worker_error_messages',
 		array(
 			'clientOffline' => __( 'It seems you are offline. Please check your internet connection and try again.', 'pwa' ),
-			'serverOffline' => __( 'The server appears to be down. Please try again later.', 'pwa' ),
+			'serverOffline' => __( 'The server appears to be down, or your connection isn\'t working as expected. Please try again later.', 'pwa' ),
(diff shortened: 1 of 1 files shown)


westonruter:
Thanks for.the PR. We tried to detect whether the user was offline in https://github.com/GoogleChromeLabs/pwa-wp/pull/190 and serve a `clientOffline` message in that case. Nevertheless, `navigator.onLine` is limited in it's effectiveness, such as to handle Lie-Fi scenarios. Is this what you experienced?

KZeni:
I can't say for certain. I was forwarded a screenshot from a client where it was giving this message even though the site was verifiably still online & working properly for others.

westonruter:
Were they able to access other websites successfully? Just wondering if we can also improve the Lie-Fi detection. 

Can you inquire if they were trying to access the site and the request took a long time until finally it failed? One thing I want to explore is make use of service worker timeouts to short-circuit requests in these cases.

At the very least, it's true that the current `onLine` detection is not robust so your change makes sense.

westonruter (approved):

KZeni:
It appears they were having issues with other sites, based on their screenshot. I'll let you know if I hear any further details back.

Thanks for merging 👍

PR #317, opened 2020-08-13
Correct version handling for precached assets

ethitter:
Investigating cache misses for precached assets revealed that the plugin wasn't building URLs in the same way that the `WP_Scripts` and `WP_Styles` classes do. 

The revision handling also contributed to cache misses. Per https://developers.google.com/web/tools/workbox/modules/workbox-precaching#explanation_of_the_precache_list, when the URL includes a version, the revision should be set to `null`.

### DIFF
--- a/integrations/class-wp-service-worker-scripts-integration.php
+++ b/integrations/class-wp-service-worker-scripts-integration.php
@@ -75,16 +75,21 @@ class WP_Service_Worker_Scripts_Integration extends WP_Service_Worker_Base_Integ
 			/** This filter is documented in wp-includes/class.wp-scripts.php */
 			$url = apply_filters( 'script_loader_src', $url, $handle );
 
+			$revision = null;
+			$version  = '';
+
 			if ( null === $dependency->ver ) {
 				$revision = wp_scripts()->default_version;
 			} else {
-				$url = add_query_arg(
-					'ver',
-					$dependency->ver ?: wp_scripts()->default_version,
-					$url
-				);
+				$version = $dependency->ver ? $dependency->ver : wp_scripts()->default_version;
+			}
 
-				$revision = null;
+			if ( isset( wp_scripts()->args[ $handle ] ) ) {
+				$version = $version ? $version . '&amp;' . wp_scripts()->args[ $handle ] : wp_scripts()->args[ $handle ];
+			}
+
+			if ( ! empty( $version ) ) {
+				$url = add_query_arg( 'ver', $version, $url );
 			}
 
 			// @todo Issue a warning when it is not a local file?
--- a/integrations/class-wp-service-worker-scripts-integration.php
+++ b/integrations/class-wp-service-worker-scripts-integration.php
@@ -72,10 +72,20 @@ class WP_Service_Worker_Scripts_Integration extends WP_Service_Worker_Base_Integ
 				$url = wp_scripts()->base_url . $url;
 			}
 
-			$revision = false === $dependency->ver ? get_bloginfo( 'version' ) : $dependency->ver;
-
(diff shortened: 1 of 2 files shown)


westonruter:
Good catch. 

In looking at `WP_Scripts::do_item()` I can see [something](https://github.com/WordPress/wordpress-develop/blob/dddddc47f4f89b9136865bf34506fcb99cfc2302/src/wp-includes/class.wp-scripts.php#L283-L285) else that may be missing ([same](https://github.com/WordPress/wordpress-develop/blob/dddddc47f4f89b9136865bf34506fcb99cfc2302/src/wp-includes/class.wp-styles.php#L164-L166) in `WP_Styles::do_item()`):

```php
if ( isset( $this->args[ $handle ] ) ) {
	$ver = $ver ? $ver . '&amp;' . $this->args[ $handle ] : $this->args[ $handle ];
}
```

I've never actually seen this code executed before. I'm not sure it's even used in core. Nevertheless, I can see that Jetpack is also accounting for this code in how it [reconstructs script URLs](https://github.com/Automattic/jetpack/blob/be0708aa2d550b4578c53e85c2386bad3f6cd327/modules/infinite-scroll/infinity.php#L1115-L1122).

ethitter:
Amusingly, [I added that](https://plugins.trac.wordpress.org/changeset/610612) to Infinite Scroll. I don't think I've ever seen it used, but if the handle is passed as `foo?bar`, then `bar` is added to the query string. I'll update the PR to account for that.

westonruter (approved):
I'll merge this next week.

integrations/class-wp-service-worker-styles-integration.php
```diff
@@ -72,11 +72,26 @@ public function register( WP_Service_Worker_Scripts $scripts ) {
...
+			$version  = '';
+
+			if ( null === $dependency->ver ) {
+				$revision = wp_styles()->default_version;
+			} else {
+				$version = $dependency->ver ? $dependency->ver : wp_styles()->default_version;
+			}
+
+			if ( isset( wp_styles()->args[ $handle ] ) ) {
+				$version = $version ? $version . '&amp;' . wp_styles()->args[ $handle ] : wp_styles()->args[ $handle ];
```
westonruter, line 88: Ditto for above:

```suggestion
				$version = $version ? $version . '&' . wp_styles()->args[ $handle ] : wp_styles()->args[ $handle ];
```

integrations/class-wp-service-worker-scripts-integration.php
```diff
@@ -72,11 +72,26 @@ public function register( WP_Service_Worker_Scripts $scripts ) {
...
+			$version  = '';
+
+			if ( null === $dependency->ver ) {
+				$revision = wp_scripts()->default_version;
+			} else {
+				$version = $dependency->ver ? $dependency->ver : wp_scripts()->default_version;
+			}
+
+			if ( isset( wp_scripts()->args[ $handle ] ) ) {
+				$version = $version ? $version . '&amp;' . wp_scripts()->args[ $handle ] : wp_scripts()->args[ $handle ];
```
westonruter, line 88: I just realized that in the context of the service worker, the `&amp;` here should be `&` since the resulting URL will not be output to HTML for the browser to decode `&amp;` back to `&`:

```suggestion
				$version = $version ? $version . '&' . wp_scripts()->args[ $handle ] : wp_scripts()->args[ $handle ];
```

westonruter (commented):
@ethitter One thing I realized after merging…

westonruter:
@ethitter Please review https://github.com/GoogleChromeLabs/pwa-wp/pull/325

PR #325, opened 2020-08-28
Avoid constructing asset versions with HTML entities in context of SW

westonruter:
Follow up on #317.

I realized that in the context of the service worker, the `&amp;` here should be changed to just `&` since the resulting URL will not be output to HTML for the browser to decode `&amp;` back to `&`.


ethitter (approved):
👍 

Good catch!

PR #326, opened 2020-08-29
Use application/javascript as content-type instead of text/javascript

### LINKED ISSUE #292: Use application/javascript instead of obsolete text/javascript for Content-Type
Apparently `text/javascript` is obsolete and should now be `application/javascript`. So this should be changed:

https://github.com/GoogleChromeLabs/pwa-wp/blob/943de6d0eba49e2814bdc66564e324c7b4c233ac/wp-includes/class-wp-service-workers.php#L116

westonruter:
Fixes #292


westonruter:
Apparently I'm wrong. See https://github.com/GoogleChromeLabs/pwa-wp/issues/292#issuecomment-683200030

PR #330, opened 2020-09-07
Add X-Robots-Tag to service worker response for good measure

### LINKED ISSUE #300: PWA 'pages' should set meta robots indexing directives
Requests which return service worker 'pages' or content - like `example.com/?wp_service_worker=1`, `example.com/?wp_error_template=offline` and `example.com/?wp_error_template=500` (and any other/similar responses) - should indicate that they shouldn't be indexed by search engines or social media.

This can be achieved by returning an `x-robots-tag` HTTP header with such responses, with a value of `noindex, follow`.

westonruter:
Fixes #300


jonoalderson (approved):
Awesome!

PR #338, opened 2020-10-10
Add default caching strategies for navigations and page assets

### LINKED ISSUE #176: Add default args for navigation caching strategy
Currently if you enable a navigation caching strategy, like Network-First:

```php
add_filter(
	'wp_service_worker_navigation_caching_strategy',
	function() {
		return WP_Service_Worker_Caching_Routes::STRATEGY_NETWORK_FIRST;
	}
);
```

You also have to configure it with arguments like this:

```php
add_filter(
	'wp_service_worker_navigation_caching_strategy_args',
	function( $args ) {
		$args['cacheName']                           = 'pages';
		$args['plugins']['expiration']['maxEntries'] = 50;
		return $args;
	}
);
```

However, in the several sites I've done this one, the args are always the same. The `cacheName` should never really need to change from site to site.

The `maxEntries` may need to be variable, but again, it seems like there should be a good default for this. For example, the fonts caching integration defaults to 30 entries:

https://github.com/xwp/pwa-wp/blob/eaed8fb7062ae510ce7309fe477902fbcc374810/integrations/class-wp-service-worker-fonts-integration.php#L43-L46

Should there also perhaps be a `maxAgeSeconds` for navigation requests caching?

### LINKED ISSUE #264: Assets required by the offline page are not cached (add network-first default caching strategy for resources)
When the offline page is served and the browser cache is disabled (or the server is configured to not cache assets) the result is a broken page:

> image

Granted, when sites are configured properly with the right `Cache-Control` headers for assets, this would not be the result. Nevertheless, it is not wise to rely on servers being properly configured.

To rectify this, all core/theme/plugin resources should have a default network-first caching strategy. This would include specifically the scripts, styles, and fonts being served from the site. It would also need to include the custom logo, header image(s), and background image. It should not include other uploaded images. 

The approach taken can be to cache by default the scripts and styles as [shown](https://gist.github.com/westonruter/1a63d052beb579842461f6ad837715fb#file-basic-site-caching-php-L54-L65) in the Basic Site Caching plugin:

```php
// Cache theme assets with runtime network-first caching strategy. This includes both the parent theme and child theme.
add_action(
	'wp_front_service_worker',
	function ( \WP_Service_Worker_Scripts $scripts ) {
		$theme_directory_uri_patterns = [
			preg_quote( trailingslashit( get_template_directory_uri() ), '/' ),
		];
		if ( get_template() !== get_stylesheet() ) {
			$theme_directory_uri_patterns[] = preg_quote( trailingslashit( get_stylesheet_directory_uri() ), '/' );
		}

		$scripts->caching_routes()->register(
			'^(' . implode( '|', $theme_directory_uri_patterns ) . ').*',
			array(
				'strategy'  => \WP_Service_Worker_Caching_Routes::STRATEGY_NETWORK_FIRST,
				'cacheName' => 'theme-assets',
				'plugins'   => array(
					'expiration' => array(
						'maxEntries' => 25, // Limit the cached entries to the number of files loaded over network, e.g. JS, CSS, and PNG.
					),
				),
			)
		);
	}
);
```

It would also then need to loop over an array of specific theme images and add them to the runtime cache directly, as [exemplified](https://github.com/ampproject/amp-wp/blob/1094ea03bd5dc92889405a47a8c41de1a88908de/assets/js/amp-service-worker-runtime-precaching.js#L4-L8) in the AMP plugin:

```js
	self.addEventListener( 'install', ( event ) => {
		event.waitUntil(
			caches.open( wp.serviceWorker.core.cacheNames.runtime ).then( ( cache ) => cache.addAll( URLS ) ),
		);
	} );
```

Or rather, each URL can be registered as a pattern for runtime caching that images which are never used won't be cached.

Note that this will make most of the “integrations” obsolete, as they rely on precaching to do the same. But precaching has drawbacks because resources are downloaded which are never used.

### LINKED ISSUE #265: Default to network-first caching strategy
The [Basic Site Caching](https://gist.github.com/westonruter/1a63d052beb579842461f6ad837715fb) plugin contains what have turned out to be the most common baseline caching strategy for sites. In particular, a network-first caching strategy ensures that the freshest content is served when connected to the network with fallbacks for to the cache when the network fails

This is the required code:

```php
<?php
// Enable network-first caching strategy for navigation requests (i.e. clicking around the site).
add_filter(
	'wp_service_worker_navigation_caching_strategy',
	function () {
		return \WP_Service_Worker_Caching_Routes::STRATEGY_NETWORK_FIRST;
	}
);

// Hold on to a certain number of navigated pages in the cache.
add_filter(
	'wp_service_worker_navigation_caching_strategy_args',
	function ( $args ) {
		$args['cacheName'] = 'pages';

		$args['plugins']['expiration']['maxEntries'] = 20;

		return $args;
	}
);
```

A network-first caching strategy is then also needed for the scripts and styles that a site depends on (coming from core, themes, and plugins). See #264.

A cache-first strategy would also make sense for uploaded files:

```php
<?php
// Add caching for uploaded images.
add_action(
	'wp_front_service_worker',
	function ( \WP_Service_Worker_Scripts $scripts ) {
		$upload_dir = wp_get_upload_dir();
		$scripts->caching_routes()->register(
			'^(' . preg_quote( $upload_dir['baseurl'], '/' ) . ').*\.(png|gif|jpg|jpeg|svg|webp)(\?.*)?$',
			array(
				'strategy'  => \WP_Service_Worker_Caching_Routes::STRATEGY_CACHE_FIRST,
				'cacheName' => 'uploads',
				'plugins'   => array(
					'expiration' => array(
						'maxAgeSeconds' => MONTH_IN_SECONDS,
					),
				),
			)
		);
	}
);
```

Note that this does not currently account for uploaded files that are hosted on an external CDN, so that should be accounted for as well.

I think this should be the default behavior of the plugin, of course allowing a site to override the default behavior with the filters. Then the PWA plugin would provide for more than just an offline page out of the box. This will in part make #176 obsolete.

westonruter:
This merges the [Basic Site Caching](https://gist.github.com/westonruter/1a63d052beb579842461f6ad837715fb) extension plugin into PWA, as well as further iterating on it.

Fixes #265.
Fixes #176.
Fixes #264.

## Navigation Caching

Previously navigation requests used [`NetworkOnly` strategy](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#network_only) by default, which meant that pages accessed while navigating would not be cached at all, so they would not be available for offline browsing. Instead, the offline page would have been served instead, which is not super helpful. This PR changes the default to the [`NetworkFirst` caching strategy](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#network_first_network_falling_back_to_cache) for navigation requests, which will allow for previously-visited pages to continue to be accessed while offline, which is they key benefit of PWA.

The strategy is also now configured by default as follows:

* `cache_name`: 'navigations'. This is the cache used for storing navigated pages. Note that the PWA plugin will automatically prefix the cache name to be specific to the WP site, especially for the sake of subdirectory installs. For example: `wp-/-navigations`.
* `network_timeout_seconds`: 2 seconds. When accessing a page, it will first attempt to fetch the most recent version from origin. If that takes longer than 2 seconds, then the previously-cached page will be served  This hopes that already-cached page assets will result in a page taking <500ms to render, thus leading to a good LCP measurement.
* `expiration.max_entries`: 10. The 10 most recently-visited pages are cached.

Previously, when wanting to configure the navigation caching strategy, you'd need to supply two filters:

```php
// 👎 Obsolete example.
add_filter(
	'wp_service_worker_navigation_caching_strategy',
	function () {
		return \WP_Service_Worker_Caching_Routes::STRATEGY_NETWORK_FIRST;
	}
);

// 👎 Obsolete example.
add_filter(
	'wp_service_worker_navigation_caching_strategy_args',
	function ( $args ) {
		$args['cacheName'] = 'navigations';
		$args['networkTimeoutSeconds'] = 2;
		$args['plugins']['expiration']['maxEntries'] = 20;
		return $args;
	}
);
```

To avoid needing to use two separate filters, this PR introduces a new `wp_service_worker_navigation_caching` filter which is passed an array that can be modified to specify _both_ the strategy and its args. With that, the `wp_service_worker_navigation_caching_strategy` and `wp_service_worker_navigation_caching_strategy_args` filters will be deprecated.  You can disable the caching of navigation responses by filtering to return an empty array like so: `add_filter( 'wp_service_worker_navigation_caching', '__return_empty_array' );`.

Note also how in the above (obsolete) examples the strategy args are in `camelCase` instead of `snake_case`, which isn't the PHP-WordPressy way manipulating arrays, even though these are what Workbox.js itself is expecting. So you now provide the keys as `snake_case` instead of `camelCase`, though the old `camelCase` names will still work as well (but you should start to only use `snake_case` going forward). The keys are converted to `camelCase` before passing into Workbox.js.

There is also this `plugins` array key which is essentially how Workbox.js is passed the configuration information to these [Workbox plugins](https://developers.google.com/web/tools/workbox/modules/workbox-strategies#using_plugins), but in the WordPress context this is confusing and unnecessary. So the array is flattened so that the plugin names can be supplied as top-level keys, so `$args['expiration']` instead of `$args['plugins']['expiration']`.

To summarize these changes:

1. The `wp_service_worker_navigation_caching` filter is used to make any changes to the `strategy` and its configuration.
2. The filters `wp_service_worker_navigation_caching_strategy` and `wp_service_worker_navigation_caching_strategy_args` should no longer be used. Eventually usage will cause deprecation notices.
3. The `snake_case` convention is used for all array keys in the configuration.
4. The configurations for Workbox plugin are provided as top-level array keys rather than being nested under `plugins`. (Similar to [`RuntimeCachingEntry`](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.RuntimeCachingEntry).)

### Example

You can now configure the navigation caching strategy with much less code. For example, here's how to configure the service worker to:

1. Use a `StaleWhileRevalidate` strategy for serving pages.
2. Only hold onto those pages for 1 minute.
3. [Only cache responses](https://developers.google.com/web/tools/workbox/modules/workbox-cacheable-response) which are OK and for non-authenticated users.
4. [Broadcast updates](https://developers.google.com/web/tools/workbox/modules/workbox-broadcast-update) when pages are updated (so you can show a notification).

You can do so with the following plugin code:

```php
// First, send a stale-while-revalidate Cache-Control header with pages for users that are not authenticated.
add_filter(
	'wp_headers',
	function ( $headers ) {
		if ( ! is_user_logged_in() && ! isset( $headers['Cache-Control'] ) ) {
			$headers['Cache-Control'] = 'max-age=1, stale-while-revalidate=59';
		}
		return $headers;
	}
);

// Now, configure the service worker to cache navigations to these pages.
add_filter(
	'wp_service_worker_navigation_caching',
	function ( $config ) {
		// Use the stale-while-revalidate strategy in the service worker.
		// Formerly specified via the wp_service_worker_navigation_caching_strategy filter.
		$config['strategy'] = WP_Service_Worker_Caching_Routes::STRATEGY_STALE_WHILE_REVALIDATE;

		// Formerly needing to have a parent 'plugins' key.
		$config['expiration'] = [
			// Only hold on to cached pages for 1 minute.
			// Formerly needing to be maxAgeSeconds.
			'max_age_seconds' => 60,
		];

		$config['cacheable_response'] = [
			// Only cache responses that are 200 OK.
			'statuses' => [ 200 ],
			// Only cache responses for unauthenticated users per wp_headers filter above.
			'headers'  => [
				'Cache-Control' => 'max-age=1, stale-while-revalidate=59', // See <https://web.dev/stale-while-revalidate/>.
			],
		];

		// Notify the page via postMessage when a stale page has been updated in the background.
		$config['broadcast_update'] = [];

		return $config;
	}
);
```

## Site Asset Caching

In addition to enabling the `NetworkFirst` caching strategy by default for navigation requests, this PR also enables the `NetworkFirst` caching strategy for theme, plugin, and core (`wp-includes`) assets. This strategy is used as opposed to `StaleWhileRevaliate` for the sake of development and since serving stale assets can leave the page in bad state. In any case, assets should already have far-future expiration in browser cache anyway. 

The expiration `max_entries` for each are informed by the 75th percentile on HTTP Archive, see https://github.com/GoogleChromeLabs/pwa-wp/issues/265#issuecomment-706612536:

  * For themes, maximum asset entries are capped at 34. This includes ~622KB.
  * For plugins, maximum asset entries are capped at 44. This includes ~449KB.
  * For core (`wp-includes`), maximum asset entries capped at 14. This includes ~178KB.

Three new filters are introduced for customizing the caching strategy configuration for these three categories:

* `wp_service_worker_theme_asset_caching`
* `wp_service_worker_plugin_asset_caching`
* `wp_service_worker_core_asset_caching`

These filters are passed a configuration array that looks the same as the array passed to `wp_service_worker_navigation_caching` for navigation requests, with one addition: there is a `route` array key which includes a regular expression pattern to match the asset URL routes to be included in caching.

## Uploaded Image Caching

Lastly, this PR adds a `StaleWhileRevalidate` caching strategy for uploaded images. This includes the files located in the site's `uploads` directory which have the file extensions returned by `wp_get_ext_types()['image']`. Cache expiration is currently set to 1 month, with limiting the number of cached images to 100 entries.

As with configuring caching strategies for navigations, core assets, theme assets, and plugin assets, the uploaded images can be configured with the `wp_service_worker_uploaded_image_caching` filter.

## Todo

- [x] Add testing.
- [x] Make sure existing camel case filtering is not overridden with defaults doe navigation requests.
- [x] Switch to the stale-while-revalidate strategy for uploaded files.
- [x] Add max_entries to uploaded files as a safeguard.


wp-includes/components/class-wp-service-worker-plugin-asset-caching-component.php
```diff
@@ -0,0 +1,60 @@
...
+	 *
+	 * @param WP_Service_Worker_Scripts $scripts Instance to register service worker behavior with.
+	 */
+	public function serve( WP_Service_Worker_Scripts $scripts ) {
+		if ( is_admin() ) {
+			return;
+		}
+
+		$scripts->caching_routes()->register(
+			'^' . preg_quote( trailingslashit( plugins_url() ), '/' ) . '.*',
```
westonruter: This should probably be limited to explicitly match images, CSS, JS, and fonts.

wp-includes/components/class-wp-service-worker-theme-asset-caching-component.php
```diff
@@ -0,0 +1,67 @@
...
+
+		$theme_directory_uri_patterns = array(
+			preg_quote( trailingslashit( get_template_directory_uri() ), '/' ),
+		);
+		if ( get_template() !== get_stylesheet() ) {
+			$theme_directory_uri_patterns[] = preg_quote( trailingslashit( get_stylesheet_directory_uri() ), '/' );
+		}
+
+		$scripts->caching_routes()->register(
+			'^(' . implode( '|', $theme_directory_uri_patterns ) . ').*',
```
westonruter: This should probably be limited to explicitly match images, CSS, JS, and fonts.

wp-includes/class-wp-service-workers.php
```diff
@@ -57,10 +57,13 @@ class WP_Service_Workers implements WP_Service_Worker_Registry_Aware {
...
 		$components = array(
-			'configuration'      => new WP_Service_Worker_Configuration_Component(),
-			'navigation_routing' => new WP_Service_Worker_Navigation_Routing_Component(),
-			'precaching_routes'  => new WP_Service_Worker_Precaching_Routes_Component(),
-			'caching_routes'     => new WP_Service_Worker_Caching_Routes_Component(),
+			'configuration'            => new WP_Service_Worker_Configuration_Component(),
+			'navigation_routing'       => new WP_Service_Worker_Navigation_Routing_Component(),
+			'theme_asset_caching'      => new WP_Service_Worker_Theme_Asset_Caching_Component(),
+			'plugin_asset_caching'     => new WP_Service_Worker_Plugin_Asset_Caching_Component(),
+			'uploaded_images__caching' => new WP_Service_Worker_Uploaded_Image_Caching_Component(),
```
westonruter: I'm not sure these should be “components”. They could rather be just global functions that are added to `wp_front_service_worker` in `default-filters.php`. This will allow them to easily be removed or replaced.

westonruter:
```suggestion
			'theme_asset_caching'      => new WP_Service_Worker_Theme_Asset_Caching_Component(),
			'plugin_asset_caching'     => new WP_Service_Worker_Plugin_Asset_Caching_Component(),
			'uploaded_images_caching' => new WP_Service_Worker_Uploaded_Image_Caching_Component(),
```

wp-includes/components/class-wp-service-worker-navigation-routing-component.php
```diff
@@ -97,17 +97,44 @@ public function serve( WP_Service_Worker_Scripts $scripts ) {
...
+			 * @since 0.6 Added $caching_strategy param and initial array has default values provided.
 			 * @see WP_Service_Worker_Caching_Routes::register()
 			 *
-			 * @param array $caching_strategy_args Caching strategy args.
+			 * @param array $caching_strategy_args {
+			 *     Caching strategy args.
+			 *
+			 *     @type string $cacheName             Cache name to store navigation responses.
+			 *     @type int    $networkTimeoutSeconds Network timeout seconds when NetworkFirst strategy is used.
+			 *     @type array  $plugins               Configuration for plugins, in particular expiration.
```
westonruter: While Workbox has this `plugins` level of nesting, I think it would make sense to promote the children up a level as it would be less confusing when extending in a WordPress context. So instead of there being `plugins.expiration` there could be just `expiration`. Naturally we'd need to move into a `plugins` array when passing over to Workbox, but we already have a set of such known plugins when we `prepare_strategy_args_for_js_export`:

https://github.com/GoogleChromeLabs/pwa-wp/blob/ec984523a90d94afe49ebf65f0ec419b026322d4/wp-includes/components/class-wp-service-worker-caching-routes.php#L173-L179

So that would be easy to do.

wp-includes/class-wp-service-workers.php
```diff
@@ -57,10 +57,13 @@ class WP_Service_Workers implements WP_Service_Worker_Registry_Aware {
...
 	public function __construct() {
 		$components = array(
-			'configuration'      => new WP_Service_Worker_Configuration_Component(),
-			'navigation_routing' => new WP_Service_Worker_Navigation_Routing_Component(),
-			'precaching_routes'  => new WP_Service_Worker_Precaching_Routes_Component(),
-			'caching_routes'     => new WP_Service_Worker_Caching_Routes_Component(),
+			'configuration'            => new WP_Service_Worker_Configuration_Component(),
+			'navigation_routing'       => new WP_Service_Worker_Navigation_Routing_Component(),
+			'theme_asset_caching'      => new WP_Service_Worker_Theme_Asset_Caching_Component(),
+			'plugin_asset_caching'     => new WP_Service_Worker_Plugin_Asset_Caching_Component(),
```
westonruter: Just realized that this is missing a caching strategy for assets from `wp-includes`, for example `jquery`, `wp-embed`, and `wp-emoji-release`.

wp-includes/components/class-wp-service-worker-plugin-asset-caching-component.php
westonruter: Perhaps not. What else would it be?

wp-includes/components/class-wp-service-worker-core-asset-caching-component.php
```diff
@@ -0,0 +1,66 @@
...
+	 *
+	 * @param WP_Service_Worker_Scripts $scripts Instance to register service worker behavior with.
+	 */
+	public function serve( WP_Service_Worker_Scripts $scripts ) {
+		if ( is_admin() ) {
+			return;
+		}
+
+		$scripts->caching_routes()->register(
+			'^' . preg_quote( trailingslashit( includes_url() ), '/' ) . '.*',
```
westonruter: This should also include assets under `s.w.org`. See 2cea537d65ee16fc581d0ff9e8995433ca73bebe.

westonruter: Otherwise, all the emojis will show as broken images when offline.

westonruter: We'll need to request that responses from `s.w.org` be served with CORS headers, however.

wp-includes/components/class-wp-service-worker-navigation-routing-component.php
westonruter: Done in efb519d3f47aa405e90c87d3842ddeae4f576b08.

wp-includes/components/class-wp-service-worker-core-asset-caching-component.php
westonruter: Requested here: https://meta.trac.wordpress.org/ticket/5477

Nevertheless, the emoji responses from `s.w.org` is are expiring in 17 years (on Thu, 31 Dec 2037 23:55:55 GMT). So we should actually be able to rely on the normal browser cache for such assets and not have to explicitly cache them in the service worker.

westonruter: Also, it would be problematic to include emoji in the same cache bucket as other core assets, since it is relatively an unbounded number of assets. The assets in core are a fixed number, but the number of emoji are vast, hence why they are distributed via a CDN. So I don't think they should be included in the end.

wp-includes/components/class-wp-service-worker-navigation-routing-component.php
```diff
@@ -97,19 +97,88 @@ public function serve( WP_Service_Worker_Scripts $scripts ) {
...
+			 * @param array $caching_strategy_args {
+			 *     Caching strategy args.
+			 *
+			 *     @type string $cache_name              Cache name to store navigation responses.
+			 *     @type int    $network_timeout_seconds Network timeout seconds when NetworkFirst strategy is used.
+			 *     @type array  $plugins                 Configuration for plugins, in particular expiration.
+			 * }
+			 * @param string $caching_strategy Caching strategy being used.
+			 */
+			$caching_strategy_args = apply_filters( 'wp_service_worker_navigation_caching_strategy_args', $caching_strategy_args, $caching_strategy );
```
westonruter: The `wp_service_worker_navigation_caching_strategy` and `wp_service_worker_navigation_caching_strategy_args` filters can be deprecated in favor of the one `wp_service_worker_navigation_caching` filter.

wp-includes/components/class-wp-service-worker-uploaded-image-caching-component.php
```diff
@@ -0,0 +1,86 @@
...
+
+		/*
+		 * Note that the path alone is used because CDN plugins may load from another domain. For example, given an
+		 * uploaded image located at:
+		 *   https://example.com/wp-content/uploads/2020/04/foo.png
+		 * Jetpack can change rewrite the URL to be:
+		 *   https://i2.wp.com/example.com/wp-content/uploads/2020/04/foo.png?fit=900%2C832&ssl=1
+		 * Therefore, the following will include any URL ending in an image file extension which also is also
+		 * preceded by '/wp-content/uploads/'.
+		 */
```
westonruter: This is not going to work. Specifically, when a CacheFirst strategy is used, if the third-party domain is not responding with CORS headers then Workbox will refuse to cache the opaque responses by default: https://developers.google.com/web/tools/workbox/guides/handle-third-party-requests

We should probably just assume that an image CDN will already have the assets cached with a far-future expires header anyway and not worry about explicitly caching the responses by the service worker beyond the browser cache.

westonruter: Additionally, it would be good to determine if WP sites generally have correct `Cache-Control` response headers being served for uploaded files. If so, and browsers are caching them for a long time, then it would not be necessary to add caching for uploaded images

wp-includes/components/class-wp-service-worker-navigation-routing-component.php
```diff
@@ -331,6 +415,7 @@ public function get_navigation_route_denylist_patterns() {
 
 			// Exclude service worker requests (to ease debugging).
 			$denylist_patterns[] = '.*\?(.*&)?(' . join( '|', array( WP_Service_Workers::QUERY_VAR ) ) . ')=';
+			$denylist_patterns[] = '.*/wp\.serviceworker(\?.*)?$';
```
westonruter, line 421: This was missed in #289.

wp-includes/service-workers.php
```diff
@@ -86,7 +86,6 @@ function wp_register_service_worker_caching_route( $route, $args = array() ) {
  * @return string Service Worker URL.
  */
 function wp_get_service_worker_url( $scope = WP_Service_Workers::SCOPE_FRONT ) {
-	/* @var WP_Rewrite $wp_rewrite */
```
swissspidy, line 89: Unrelated change?

westonruter: I was getting phpcs complaints about it, and since it wasn't needed, I just decided to remove it.

wp-includes/components/class-wp-service-worker-caching-routes.php
```diff
@@ -66,22 +80,64 @@ class WP_Service_Worker_Caching_Routes implements WP_Service_Worker_Registry {
...
+	 *                                          This is deprecated in favor of defining the plugins in the top-level.
+	 *                                          See <https://developers.google.com/web/tools/workbox/guides/using-plugins#workbox_plugins>.
+	 * @return bool Whether the registration was successful.
 	 * }
 	 */
 	public function register( $route, $args = array() ) {
-		if ( ! is_array( $args ) ) {
+		if ( is_string( $args ) ) {
 			$args = array( 'strategy' => $args );
 		}
```
felixarntz: With this change, maybe we should ensure elsewhere that `$args` is in fact an array. Potentially add a `_doing_it_wrong` to the `normalize_configuration` method if `! is_array( $args )`?

wp-includes/components/class-wp-service-worker-caching-routes.php
```diff
@@ -66,22 +80,64 @@ class WP_Service_Worker_Caching_Routes implements WP_Service_Worker_Registry {
...
+				'0.2'
+			);
+			return false;
+		}
+
+		$errors = new WP_Error();
+		$args   = self::normalize_configuration( $args, $errors );
+		if ( ! empty( $errors->errors ) ) {
+			foreach ( $errors->errors as $error_messages ) {
+				_doing_it_wrong( __METHOD__, esc_html( current( $error_messages ) ), '0.6' );
```
felixarntz: Nit-picking, but why not go over all messages with another loop? I'm sure we'll never put multiple error messages for one error code, but we _could_, so maybe worth making this a bit more bulletproof. :)

wp-includes/components/class-wp-service-worker-caching-routes.php
```diff
@@ -66,22 +80,64 @@ class WP_Service_Worker_Caching_Routes implements WP_Service_Worker_Registry {
...
+		}
+
+		$args['route']  = $route;
+		$this->routes[] = $args;
+		return true;
+	}
+
+	/**
+	 * Normalize configuration.
+	 *
```
felixarntz, line 149:
```suggestion
	 *
	 * @since 0.6
	 *
```

wp-includes/components/class-wp-service-worker-caching-routes.php
```diff
@@ -137,66 +242,48 @@ public function get_all() {
...
-		if ( isset( $strategy_args['plugins'] ) ) {
-			$plugins = $strategy_args['plugins'];
-			unset( $strategy_args['plugins'] );
-		}
-
-		foreach ( $strategy_args as $strategy_arg_name => $strategy_arg_value ) {
-			if ( false !== strpos( $strategy_arg_name, '_' ) ) {
-				$strategy_arg_name = preg_replace_callback( '/_[a-z]/', array( __CLASS__, 'convert_snake_case_to_camel_case_callback' ), $strategy_arg_name );
+		foreach ( self::WORKBOX_CORE_PLUGINS as $plugin_name ) {
+			if ( array_key_exists( $plugin_name, $strategy_args ) ) {
```
felixarntz, line 275: Mixing plugin args with other args feels a bit error-prone to me - maybe we can keep the plugin-specific data in a `plugins` or `plugin_config` sub-key associative array like it was before? I think this will better prepare it for a potential future where we also allow customizability via other plugins than those we "recognize".

Something like:
```suggestion
			if ( ! empty( $strategy_args['plugins'] ) && array_key_exists( $plugin_name, $strategy_args['plugins'] ) ) {
```

As far as I can tell this will also keep backward-compatibility in the format which would be a nice side-effect?

wp-includes/components/class-wp-service-worker-caching-routes.php
```diff
@@ -66,22 +80,64 @@ class WP_Service_Worker_Caching_Routes implements WP_Service_Worker_Registry {
...
+	 *     @type array|null $broadcast_update   Broadcast update plugin configuration. See <https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-broadcast-update.BroadcastUpdatePlugin>.
+	 *     @type array|null $cacheable_response Cacheable response plugin configuration. See <https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-cacheable-response.CacheableResponsePlugin>.
+	 *     @type array|null $background_sync    Background sync plugin configuration. See <https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-background-sync.BackgroundSyncPlugin>.
+	 *     @type array|null $plugins            Deprecated. Array of plugins with configuration. The key of each plugin in the array must match the plugin's name.
+	 *                                          This is deprecated in favor of defining the plugins in the top-level.
+	 *                                          See <https://developers.google.com/web/tools/workbox/guides/using-plugins#workbox_plugins>.
+	 * @return bool Whether the registration was successful.
 	 * }
 	 */
 	public function register( $route, $args = array() ) {
```
felixarntz: I think the changes made here indicate that it might be more intuitive to make `route` part of `$args` here too and have that as the only parameter. The `strategy` argument is required anyway, so I think this array overall should be required.

We could change the method signature to something like `register( $args, $deprecated = array() )`, and then if `$args` is a string, assume that it is actually `$route` and then `$deprecated` would be `$args`.

wp-includes/components/class-wp-service-worker-core-asset-caching-component.php
```diff
@@ -0,0 +1,92 @@
...
+		 * }
+		 */
+		$config = apply_filters( 'wp_service_worker_core_asset_caching', $config );
+
+		if ( ! is_array( $config ) || ! isset( $config['route'], $config['strategy'] ) ) {
+			return;
+		}
+
+		$route = $config['route'];
+		unset( $config['route'] );
```
felixarntz, line 77: See my above suggestion, things like this wouldn't be needed then.

wp-includes/components/class-wp-service-worker-uploaded-image-caching-component.php
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @package PWA
+ */
+
+/**
+ * Use a cache-first caching strategy for uploaded images.
+ *
+ * @since 0.6
+ */
+class WP_Service_Worker_Uploaded_Image_Caching_Component implements WP_Service_Worker_Component {
```
felixarntz, line 13: I guess it makes sense to start with images here since they're by far the most common. What about e.g. audio though? I assume video would be too resource-heavy.

wp-includes/service-workers.php
felixarntz: Maybe worth adding a `@global WP_Rewrite $wp_rewrite WordPress rewrite rules object.` note to the doc-block instead, like common in core?

pwa.php
```diff
@@ -267,7 +271,6 @@ function _pwa_activate_plugin() {
  * Plugin deactivation hook.
  */
 function _pwa_deactivate_plugin() {
-	/* @var WP_Rewrite $wp_rewrite */
```
felixarntz, line 270: See above comment.

felixarntz (commented):
Overall looks solid to me, that's gonna be a massive enhancement! Just some feedback and suggestions.

wp-includes/components/class-wp-service-worker-uploaded-image-caching-component.php
westonruter: Yeah, that's right. And a reason against doing audio/video is that they add the additional complexity of browsers performing range requests to get back data. There are a few wrinkles for serving cached audio/video: https://developers.google.com/web/tools/workbox/guides/advanced-recipes#cached-av

This being the case, it seems best to leave audio/video caching to be left up to a site to opt-in to.

wp-includes/service-workers.php
westonruter: Of course! 👍

westonruter: Done in 8749128916c9807474f8bd6671ff654ff6dafc3f

pwa.php
westonruter: Done: Done in 8749128916c9807474f8bd6671ff654ff6dafc3f.

wp-includes/components/class-wp-service-worker-caching-routes.php
westonruter: How about da48a2294968e8970bd15e251fa218e67434d1b4?

wp-includes/components/class-wp-service-worker-caching-routes.php
westonruter: I actually went down that “route” in the course of this PR, but then I decided against it. The reason I decided to keep two arguments is that it keeps the method signature consistent with `WP_Service_Worker_Precaching_Routes::register()`. Both then take a URL/route and then args. In both cases, the args can be a string. For precaching, the `$args` can be a string as a shortcut to be the `revision`. For `WP_Service_Worker_Caching_Routes::register()`, the `$args` as a string is a shortcut to provide the `strategy`. 

It's kinda handy to be able to do:

```php
$scripts->caching_routes()->register( '/foo', WP_Service_Worker_Caching_Routes::STRATEGY_NETWORK_FIRST ); 
```

Rather than:

```php
$scripts->caching_routes()->register( '/foo', array(
    'strategy' => WP_Service_Worker_Caching_Routes::STRATEGY_NETWORK_FIRST,
) ); 
```

So that's the intention here.

wp-includes/components/class-wp-service-worker-caching-routes.php
westonruter: How about concatenating the messages together as in f1381df?

wp-includes/components/class-wp-service-worker-caching-routes.php
westonruter: I actually inquired about this with Jeff Posnick. I was also hesitant about whether flattening the config would cause conflicts. He said it would not be a problem, and in fact the flattened config is similar to what Workbox.js calls a [`RuntimeCachingEntry`](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.RuntimeCachingEntry) which you can see defines the plugin-specific configuration alongside strategy-specific configuration. It does have an additional `plugins` key:

image

But this isn't relevant in the PHP API since all of the plugins are among a fixed set:

https://github.com/GoogleChromeLabs/pwa-wp/blob/f1381dfe593dfc4df281bc82720bddb529db2df3/wp-includes/components/class-wp-service-worker-caching-routes.php#L61-L67

We limit the set because it's how we know how to construct the plugin object:

https://github.com/GoogleChromeLabs/pwa-wp/blob/f1381dfe593dfc4df281bc82720bddb529db2df3/wp-includes/components/class-wp-service-worker-caching-routes.php#L280-L285

I think it's advantageous to remove the `plugins` key from the configuration because it can be confusing for WP developers who have a different conception of what a plugin is. 

If someone wants to create a custom plugin, they can do so but they'd have to register it via JS instead of the PHP API.

wp-includes/components/class-wp-service-worker-core-asset-caching-component.php
westonruter: See my reply: https://github.com/GoogleChromeLabs/pwa-wp/pull/338/files#r523473404

wp-includes/components/class-wp-service-worker-caching-routes.php
felixarntz: Okay that makes sense then. Thanks for clarifying.

wp-includes/components/class-wp-service-worker-caching-routes.php
felixarntz: Hmm that's a fair point. However, then I'd suggest to give both of these methods three parameters: The URL/route and the strategy/revision are both required, so I think it's counterintuitive to have these in `$args`. IMO `$args` should always be an optional thing for more "advanced" customizations.

wp-includes/components/class-wp-service-worker-core-asset-caching-component.php
felixarntz: Left a reply, but either way, my previous comment on here would be irrelevant with my new suggestion there too.

felixarntz (commented):
@westonruter Just one follow-up comment about IMO more intuitive `register` method signatures in https://github.com/GoogleChromeLabs/pwa-wp/pull/338#discussion_r527330435.

wp-includes/components/class-wp-service-worker-caching-routes.php
westonruter: So you're saying for caching routes it should be:

```php
$scripts->caching_routes()->register( 
    '/foo', 
    WP_Service_Worker_Caching_Routes::STRATEGY_NETWORK_FIRST,
    array( /* Additional optional args. */ )
); 
```

And for precaching routes (which is already supported):

```php
$scripts->precaching_routes()->register( 
    '/foo', 
    $revision
); 
```

Oh, but actually `revision` is _not_ required in the [precache list](https://developers.google.com/web/tools/workbox/modules/workbox-precaching#explanation_of_the_precache_list). But it's the only argument that precaching currently supports. There are no other arguments. So this is fine too:

```php
$scripts->precaching_routes()->register( '/foo' ); 
```

This isn't the case for caching routes, where the strategy must be provided. So in that case, `register` for caching routes would have 3 arguments but `register` for precaching routes would have only 2?

felixarntz: Yes, the signatures you put as examples look good to me. Regarding `$revision` not being required, I think that's okay, in this case we can make it optional. Now, overall I think optional arguments make sense to go into an `$args` array, but if `$revision` is the only one, I don't think that would be needed here - so I suggest to make it an optional string parameter and not have `$args` for precaching routes at all.

If at some point we need additional optional arguments for precaching routes, we can transform the second parameter to an optional `$args` and keep backward-compatibility by still supporting a string (and then assuming that is the revision).

westonruter: Ah, problem: a method signature params of `( $route, $strategy, $args = array() )` will violate the implementation of `WP_Service_Worker_Registry::register` interface, which has `( $handle, $args = array() )`. Ditch the interface?

felixarntz: @westonruter Yeah, I think that makes sense. We should then remove `WP_Service_Worker_Registry_Aware` as well.

westonruter:
I've implemented that change in 5f1c00a but pulling that thread caused several other necessary changes. IMO the classes can further be consolidated and made more wordpressy now. It doesn't seem like the classes are helping so much as much as getting in the way at the moment. For example, the various `WP_Service_Worker_Component` classes could just as well be registered as functions on the `wp_front_service_worker` action. There doesn't seem to be a need for the component abstraction, especially for `WP_Service_Worker_Caching_Routes` and `WP_Service_Worker_Precaching_Routes`.

felixarntz:
@westonruter The changes you made look good to me. IMO keeping the component classes still makes sense over functions at least because they allow splitting up into multiple private methods rather than giant functions or "private" functions which aren't possible.

felixarntz (approved):
Awesome!

NicoHood:
I just came here from:
https://github.com/GoogleChromeLabs/pwa-wp/wiki/Service-Worker#page-navigation-caching

My issue is, that when I load the app (after a few days), the front page is still loading the cached version and not the most up to date information. I'd like to solve this.

I read that I can:
a) Disable caching
b) Set caching timeout to 5 seconds
c) Use `StaleWhileRevalidate` Strategy

But then I also see that time has passed and this PR was merged and I am curious if the wiki is still up to date. What would you recommend me to solve the issue?

westonruter:
@NicoHood Sorry for the delay. So the issue is that the `NetworkFirst` strategy is timing out at 2 seconds, and so then it serves the cached version?

First of all, you should really try to improve your server response times so that the responses are served in 600ms or less. Usually a good page caching plugin will get you what you need.

If this is not possible, increasing the TTL from the default of 2 seconds to 5 seconds should do the trick:

```php
add_filter( 'wp_service_worker_navigation_caching', static function ( $config ) {
	$config['network_timeout_seconds'] = 5;
	return $config;
} );
```

It is possible to use a `StaleWhileRevalidate` strategy as well with this code:

```php
add_filter( 'wp_service_worker_navigation_caching', static function ( $config ) {
	$config['strategy'] = 'StaleWhileRevalidate';
	return $config;
} );
```

However, this will _guarantee_ that stale responses will be served, which I don't think you want.

NicoHood:
Thanks!

I moved to another server which I hope will respond faster. Google Page Speed at least tells me that its better.

But what about mobile networks? Is 2 seconds the time the server will a) answer or b) transmit the whole webpage? Maybe 2 seconds is still a bit short for mobile networks, isn't it? Who set 2 seconds as a default? Is it just a best guess, or a specification default, etc?

westonruter:
The value of 2 seconds is informed by Core Web Vitals. To achieve a good LCP metric, the largest contentful paint needs to happen within 2.5 seconds of the page loading. So by having a 2-second TTL for loading a page from the network, the browser then has 500ms to render the page if it was retrieved from cache.

PR #359, opened 2020-12-10
Deprecate wp_service_worker_json_encode() in favor of wp_json_encode()

### LINKED ISSUE #341: Eliminate use of wp_service_worker_json_encode()
The `wp_service_worker_json_encode()` function was added to automatically strip slashes and pretty-print. This was for the sake of manually inspecting the service worker JS. However, this adds additional bytes to the response. And in practice, pretty-printing is easily done in DevTools anyway.

So I want to eliminate the use of `wp_service_worker_json_encode()`. We can just remove it since no other plugins are using this function, but to be safe we can deprecate it.

westonruter:
Fixes #341

### DIFF
--- a/wp-includes/deprecated.php
+++ b/wp-includes/deprecated.php
@@ -39,3 +39,17 @@ function wp_service_worker_update_node() {
 function wp_hide_admin_bar_offline() {
 	_deprecated_function( __FUNCTION__, '0.5' );
 }
+
+/**
+ * JSON-encodes with pretty printing.
+ *
+ * @since 0.2
+ * @deprecated 0.6 Now wp_json_encode() is used directly.
+ *
+ * @param mixed $data Data.
+ * @return string JSON.
+ */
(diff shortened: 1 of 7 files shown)


wp-includes/deprecated.php
```diff
@@ -33,9 +35,25 @@ function wp_service_worker_update_node() {
...
+ *
+ * @since 0.2
+ * @deprecated 0.6 Now wp_json_encode() is used directly.
+ * @codeCoverageIgnore
+ *
+ * @param mixed $data Data.
+ * @return string JSON.
+ */
+function wp_service_worker_json_encode( $data ) {
+	_deprecated_function( __FUNCTION__, '0.6' );
```
felixarntz: Maybe add the replacement pointer here?
```suggestion
	_deprecated_function( __FUNCTION__, '0.6', 'wp_json_encode()' );
```

felixarntz (approved):
LGTM, but one non-blocking suggestion.

PR #360, opened 2020-12-10
Make all classes final since not intended to be extended

### LINKED ISSUE #340: Make classes final
I just noticed that classes like `WP_Service_Worker_Caching_Routes` are not marked as `final`:

https://github.com/GoogleChromeLabs/pwa-wp/blob/230ed8ff8cd2f101b04b2d0afed6d83e00c5f4c1/wp-includes/components/class-wp-service-worker-caching-routes.php#L13

All classes that are not intended to be extended should be marked as `final`.

westonruter:
Fixes #340

### DIFF
--- a/integrations/class-wp-service-worker-admin-assets-integration.php
+++ b/integrations/class-wp-service-worker-admin-assets-integration.php
@@ -10,7 +10,7 @@
  *
  * @since 0.2
  */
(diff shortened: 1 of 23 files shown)


felixarntz (approved):
👍

PR #365, opened 2020-12-12
Add offline browsing option in Reading settings

### LINKED ISSUE #344: Provide means of opting-in to offline browsing
Default caching strategies for navigations, theme/plugin/core assets, and uploaded images have been introduced in #338. While the intention is that these are conservative and that they merely enable offline browsing, there is a danger that users may not want it. For that reason, there should probably be an opt-in setting in Reading settings for whether offline browsing is enabled. Something like this:

image

When that checkbox is not checked, then the behavior would be the same as if these filters being in place:

```php
add_filter( 'wp_service_worker_navigation_caching', '__return_empty_array' );
add_filter( 'wp_service_worker_theme_asset_caching', '__return_empty_array' );
add_filter( 'wp_service_worker_core_asset_caching', '__return_empty_array' );
add_filter( 'wp_service_worker_plugin_asset_caching', '__return_empty_array' );
add_filter( 'wp_service_worker_uploaded_image_caching', '__return_empty_array' );
```

westonruter:
Fixes #344.

This adds a new **Offline browsing** toggle to the Reading Settings screen:

> image

It is _disabled_ by default. When disabled, by default only the offline page and server error page templates will be served to the user (or any other custom rules that the user may have registered such as via the obsolete [Basic Site Caching](https://gist.github.com/westonruter/1a63d052beb579842461f6ad837715fb) plugin). When enabled, then the new default caching strategies are employed which were introduced in https://github.com/GoogleChromeLabs/pwa-wp/pull/338.

When someone updates to this version of a plugin, an admin pointer is displayed displayed to let them know about this feature:

> image

### DIFF
--- a/wp-admin/options-reading-offline-browsing.php
+++ b/wp-admin/options-reading-offline-browsing.php
@@ -1,0 +1,61 @@
+<?php
+/**
+ * Offline browsing field for the reading settings administration panel.
+ *
+ * As part of a core merge, the code in this file would go inside wp-admin/options-reading.php
+ *
+ * @package PWA
+ */
+
+namespace PWA_WP;
+
+/**
+ * Register offline browsing setting.
+ */
+function register_offline_browsing_setting() {
+	register_setting(
+		'reading',
+		'offline_browsing',
+		array(
+			'type'              => 'boolean',
+			'sanitize_callback' => 'rest_sanitize_boolean',
+		)
+	);
+}
+add_action( 'init', __NAMESPACE__ . '\register_offline_browsing_setting' );
+
+/**
+ * Register offline browsing setting field.
+ */
+function add_offline_browsing_setting_field() {
+	add_settings_field(
+		'offline_browsing',
+		__( 'Offline browsing', 'pwa' ),
+		__NAMESPACE__ . '\render_offline_browsing_setting_field',
+		'reading'
+	);
+}
+add_action( 'admin_init', __NAMESPACE__ . '\add_offline_browsing_setting_field' );
+
+/**
+ * Render offline browsing setting field.
+ */
+function render_offline_browsing_setting_field() {
+	?>
+	<fieldset>
+		<legend class="screen-reader-text"><span><?php esc_html_e( 'Offline browsing', 'pwa' ); ?> </span></legend>
+		<label for="offline_browsing">
+			<input name="offline_browsing" type="checkbox" id="offline_browsing" value="1" <?php checked( '1', get_option( 'offline_browsing' ) ); ?> />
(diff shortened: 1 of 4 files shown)


wp-admin/options-reading-offline-browsing.php
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Render offline browsing setting field.
+ */
+function render_offline_browsing_setting_field() {
+	?>
+	<fieldset>
+		<legend class="screen-reader-text"><span><?php esc_html_e( 'Offline browsing', 'pwa' ); ?> </span></legend>
+		<label for="offline_browsing">
+			<input name="offline_browsing" type="checkbox" id="offline_browsing" value="1" <?php checked( '1', get_option( 'offline_browsing' ) ); ?> />
+			<?php esc_html_e( 'Cache visited pages in the browser so visitors can re-access then when offline.', 'pwa' ); ?>
```
derekherman:
```suggestion
			<?php esc_html_e( 'Cache visited pages in the browser so visitors can re-access them when offline.', 'pwa' ); ?>
```

wp-admin/options-reading-offline-browsing.php
```diff
@@ -0,0 +1,131 @@
...
+		'content'  => $content,
+		'position' => array(
+			'align' => 'middle',
+			'edge'  => is_rtl() ? 'right' : 'left',
+		),
+	);
+
+	?>
+	<script type="text/javascript">
+		jQuery( function( $ ) {
```
derekherman, line 102: Is the jQuery dependency necessary or could this be written in vanilla JS?

westonruter: It could yes, but the [wp-pointer](https://github.com/WordPress/wordpress-develop/blob/master/src/js/_enqueues/lib/pointer.js) JS API is itself based on jQuery, so that's why I'm using it here.

derekherman:
@googlebot I consent.

wp-admin/options-reading-offline-browsing.php
```diff
@@ -0,0 +1,131 @@
...
+			if ( menuSettingsItem.hasClass( 'wp-menu-open' ) ) {
+				target = readingSettingsItem;
+			} else {
+				target = menuSettingsItem;
+			}
+
+			target.pointer( options ).pointer( 'open' );
+		} );
+	</script>
+	<?php
```
felixarntz, line 128: A shame we can't use this code from WP core and have to copy it here. But okay 🙃

felixarntz (approved):
LGTM!

PR #354, opened 2020-12-06
Eliminate Travis in favor of GitHub Actions for linting and testing

westonruter:
Props to @pierlon for the infrastructure work done originally in the [AMP plugin](https://github.com/ampproject/amp-wp), where GitHub actions were initially used to run tests instead of Travis in https://github.com/ampproject/amp-wp/pull/5201.

Fixes #197.

PR #400, opened 2021-01-25
Make README.md canonical, generating readme.txt at build time

westonruter:
Instead of maintaining `readme.txt` and synchronizing it to `readme.md`, we can just keep the markdown version alone and then generate the `readme.txt` at build time. Based on learnings from https://github.com/ampproject/amp-wp/pull/5807 and https://github.com/ampproject/amp-wp/pull/5815.

Also eliminates `xwp/wp-dev-lib` dependency, which is obsolete now that we're no longer using `generate-markdown-readme` nor the `check-diff.sh` (since `lint-staged` was added in #354).


pooja-muchandikar:
QA Passed ✅ 

readme.txt is getting generated at build time as expected. 
image

PR #414, opened 2021-02-01
Use wp_robots() to prevent indexing the error template on WP≥5.7

westonruter:
This prevents a `wp_no_robots()` deprecation warning from being raised in WP 5.7-alpha which was introduced in https://github.com/WordPress/wordpress-develop/pull/702.

### DIFF
--- a/wp-includes/default-filters.php
+++ b/wp-includes/default-filters.php
@@ -17,7 +17,12 @@ add_action( 'wp_ajax_wp_service_worker', 'wp_ajax_wp_service_worker' );
 add_action( 'wp_ajax_nopriv_wp_service_worker', 'wp_ajax_wp_service_worker' );
 add_action( 'parse_query', 'wp_unauthenticate_error_template_requests' );
 
-add_action( 'wp_head', 'wp_add_error_template_no_robots' );
-add_action( 'error_head', 'wp_add_error_template_no_robots' );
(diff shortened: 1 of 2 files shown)


wp-includes/default-filters.php
```diff
@@ -17,7 +17,12 @@
 add_action( 'wp_ajax_nopriv_wp_service_worker', 'wp_ajax_wp_service_worker' );
 add_action( 'parse_query', 'wp_unauthenticate_error_template_requests' );
 
-add_action( 'wp_head', 'wp_add_error_template_no_robots' );
-add_action( 'error_head', 'wp_add_error_template_no_robots' );
+if ( version_compare( strtok( get_bloginfo( 'version' ), '-' ), '5.7', '>=' ) ) {
```
felixarntz, line 20: Alternatively we could check for whether the `wp_robots` function exists to avoid the need for specific version lookup - but I guess the downside of that would be that someone else theoretically could declare a function of that name before. Just a thought, nothing blocking.

felixarntz (approved):
LGTM!

pooja-muchandikar:
QA Passed ✅ 

With WP 5.7 and PWA 0.6 version indexing was not part of error template
Screenshot 2022-04-19 at 2 59 36 PM

---

With WP 5.9.3 and PWA  v0.7.0-alpha-52d5096-20220419T054300Z `noindex` is getting appended on error template
Screenshot 2022-04-19 at 3 01 28 PM

PR #544, opened 2021-06-18
Don't flush rewrite_rules on every admin request

diedexx:
This PR fixes an issue where `flush_rewrite_rules` is called on every wp-admin pageload.

`$wp_rewrite->extra_rules_top` only includes rules that were registered during the current request. And because `add_rewrite_rule` is called after the logic that checks for the same rule on `extra_rules_top`, it will never exist and _always_ call `flush_rewrite_rules`. 

Instead, I suggest to do the check against `$wp_rewrite->wp_rewrite_rules()`, as that fetches the existing rules from the database.

### DIFF
--- a/wp-includes/class-wp.php
+++ b/wp-includes/class-wp.php
@@ -14,8 +14,9 @@
 function pwa_add_rewrite_rules() {
 	global $wp_rewrite;
 	$rewrite_rule_regex = '^wp\.serviceworker$';
+	$rules = $wp_rewrite->wp_rewrite_rules();
 
-	if ( ! isset( $wp_rewrite->extra_rules_top[ $rewrite_rule_regex ] ) ) {
+	if ( ! isset( $rules[ $rewrite_rule_regex ] ) ) {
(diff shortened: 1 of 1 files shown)


wp-includes/class-wp.php
```diff
@@ -14,8 +14,9 @@
 function pwa_add_rewrite_rules() {
 	global $wp_rewrite;
 	$rewrite_rule_regex = '^wp\.serviceworker$';
+	$rules = $wp_rewrite->wp_rewrite_rules();
```
westonruter:
```suggestion
	$rules              = $wp_rewrite->wp_rewrite_rules();
```

moorscode:
@googlebot I signed it!

westonruter:
@moorscode The CLA bot needs you to sign as well now.

moorscode:
@googlebot I fixed it.

diedexx:
@googlebot I consent.

westonruter (approved):
Thank you!

And sorry for the hassle with the CLA.

pooja-muchandikar:
QA Passed ✅ 

The changes are working as expected.

Issue #403, opened 2021-01-26
Deprecate integrations


westonruter:
As noted in the [Service Worker wiki page](https://github.com/GoogleChromeLabs/pwa-wp/wiki/Service-Worker) section on [Integrations](https://github.com/GoogleChromeLabs/pwa-wp/wiki/Service-Worker#integrations):

> For these reasons, precaching resources should be done very sparingly. (The only two resources which are always precached are the offline page and the error page.) Instead, runtime caching should be used with a network-first caching strategy so that the latest version of a resource will be used whenever it is changed and then fallback to a previously-cached resource.

The integrations were largely experimental and in practice they've not been used 20% (as they are only usable via code opt-in) to warrant being part of the plugin itself for core merge.

thelovekesh:
@westonruter Can you please share how we are planning to deprecate the integrations? Are we completely removing them in this version or just adding a deprecating notice for now.

If we are just going to add a message then this is just a matter of adding `_deprecated_function` in the `pwa.php`.
```php
/**
 * Load service worker integrations.
 *
 * @since 0.2.0
 * @deprecated 0.7 Support for Integrations will be removed in a future version.
 * @codeCoverageIgnore
 *
 * @param WP_Service_Worker_Scripts $scripts Instance to register service worker behavior with.
 */
function pwa_load_service_worker_integrations( WP_Service_Worker_Scripts $scripts ) {

	_deprecated_function( __FUNCTION__, '0.7', );

	if ( ! current_theme_supports( 'service_worker' ) ) {
		return;
	}

	/** WordPress Service Worker Integration Functions */
	require_once PWA_PLUGIN_DIR . '/integrations/functions.php';

	pwa_register_service_worker_integrations( $scripts );
}
```

If you have some different ideas with integrations deprecation then please let me know.

thelovekesh:
@westonruter Any update on this one?

westonruter:
Perhaps to start with we should add `@deprecated` tags to the classes/functions, and then when an integration is used it can emit a `console.warn()` saying something like:

```js
console.warn( "The X integration in the PWA plugin is deprecated and no longer being considered for WordPress core inclusion." );
```

That would probably be a good first step as a soft-deprecation.

pooja-muchandikar:
QA Passed ✅ 

To cross check this issue added a line of code in theme's function.php file `add_theme_support( 'service_worker', true );` once this was added console error gave all deprecated errors as expected.

[...]

PR #578, opened 2021-07-26
Utilize workbox-window for service worker registration

### LINKED ISSUE #204: Make use of Workbox window to register service worker and provide access to registration
Workbox now makes available a [`workbox-window` module](https://developers.google.com/web/tools/workbox/modules/workbox-window):

<blockquote>
The `workbox-window` package is a set of modules that are intended to run in the window context, which is to say, inside of your web pages. They're a complement to the other workbox packages that run in the service worker.

The key features/goals of `workbox-window` are:

* To simplify the process of service worker registration and updates by helping
developers identify the most critical moments in the service worker lifecycle, and making it easier to respond to those moments.
* To help prevent developers from making the most common mistakes.
* To enable easier communication between code running in the service worker and code running in the window.
</blockquote>

We should be using this as opposed to registering it directly:

https://github.com/xwp/pwa-wp/blob/eb4fcc12945acd7b7afbd10cfb0d91d0b3749c22/wp-includes/service-workers.php#L161-L163

The `Workbox` instance can then be stored in a `wp.serviceWorker` global. This will then allow plugins to make use of the [`getSW()`](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-window.Workbox#getSW) method, with a goal to provide a reliable way for themes and plugins to access the service worker controlling the page once it is registered and active.

See [reference docs](https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-window.Workbox).

### LINKED ISSUE #227: Backward Compatibility for IE11
Hi,

I am wondering if there is any particular reason why older browsers are not supported at the moment? 

The JavaScript parts of the plugins is written as ES6. This means the scripts cannot be loaded in  browser that does not support es 6 features like fat arrow.  

I know ServiceWorkers are not even supported in IE but an error free console is desirable(in my opinion).

It is possible to transpile the existing ES6 back to ES5 and keep the existing functionality with Babel. It could be integrated into the Grunt building process. Except in cases when JS is embedded in PHP code like [Here](https://github.com/mxsxs2/pwa-wp/blob/master/wp-includes/service-workers.php#L155). These would have to be written in ES5 manually.

### LINKED ISSUE #444: Eliminate custom skipWaiting message in favor of de facto Workbox message
We currently allow a client to post a `{action:'skipWaiting'}` message to cause the service worker to skip waiting (which is done by default in the SW now anyway):

https://github.com/GoogleChromeLabs/pwa-wp/blob/877ff1ee4d4a5e99c9f737383f58ee2537b5ea64/wp-includes/js/service-worker.js#L13-L22

However, Workbox indicates there is a _de facto_ message `{type:'SKIP_WAITING'}`. See https://github.com/GoogleChrome/workbox/issues/2394.

The code to handle this message is not part of a Workbox package but rather is part of the SW template: https://github.com/GoogleChrome/workbox/blob/ae29e1748bb976351db8e0c64a9ea9fc8ba2583f/packages/workbox-build/src/templates/sw-template.js#L34-L38

So we'll need to incorporate that instead of what we have now. 

This is part of the Workbox v6 upgrade and it coincides with opening up the use of `workbox.window` in #204, as the client could then do `workbox.messageSkipWaiting()`. See [v6 release notes](https://github.com/GoogleChrome/workbox/releases/tag/v6.0.0#:~:text=messageSkipWaiting).

westonruter:
* Fixes #204. Service worker registration is available at `wp.serviceWorkerWindow` at the `document`'s `DOMContentLoaded` event.
* Discontinue registering the service worker for the admin on the frontend when the user is logged-in. The service worker is installed when the user goes to the login screen already.
* Fixes #444. When `wp_service_worker_skip_waiting` is filtered `false` in PHP, service worker installation waiting can be skipped via `wp.serviceWorkerWindow.messageSkipWaiting()`. See [Workbox docs](https://developers.google.com/web/tools/workbox/modules/workbox-window#skip_waiting_helper).
* Deprecates `wp_disable_script_concatenation()` since admin service worker is not caching admin assets since `WP_Service_Worker_Admin_Assets_Integration` is not enabled by default (and it is deprecated per #403). As part of this the `wordpress_sw_installed` cookie is no longer set when the admin SW is installed.
* Use `module` script instead of `text/javascript` for SW installation on frontend. This incidentally prevents JS errors IE11 which doesn't understand JS modules. Fixes #227.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -32,7 +32,15 @@ module.exports = function (grunt) {
 			},
 			install_workbox: {
 				command:
-					'if [ -e wp-includes/js/workbox* ]; then rm -r wp-includes/js/workbox*; fi; npx workbox copyLibraries wp-includes/js/',
+					'if [ -e wp-includes/js/workbox* ]; then ' +
+					'rm -r wp-includes/js/workbox*; ' +
+					'fi; ' +
+					'npx workbox copyLibraries wp-includes/js/;' +
+					// Rename .mjs with .js since servers may not be configured to serve the former as text/javascript.
+					'for mjs in wp-includes/js/workbox*/*.mjs; do ' +
+					'sed "s/\\.mjs/.js/g" "$mjs" > "${mjs%mjs}js";' +
+					'rm $mjs;' +
+					'done;',
 			},
 			create_build_zip: {
 				command:
--- a/wp-includes/default-filters.php
+++ b/wp-includes/default-filters.php
@@ -24,5 +24,3 @@ if ( version_compare( strtok( get_bloginfo( 'version' ), '-' ), '5.7', '>=' ) ) 
 	add_action( 'wp_head', 'wp_add_error_template_no_robots' );
 	add_action( 'error_head', 'wp_add_error_template_no_robots' );
 }
-
-add_action( 'admin_init', 'wp_disable_script_concatenation' );
--- a/wp-includes/deprecated.php
+++ b/wp-includes/deprecated.php
@@ -57,3 +57,30 @@ function wp_service_worker_json_encode( $data ) {
 	_deprecated_function( __FUNCTION__, '0.6', 'wp_json_encode()' );
 	return wp_json_encode( $data, 128 | 64 /* JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES */ );
 }
+
+/**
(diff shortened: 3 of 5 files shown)


amedina (approved):
LGTM.

PR #581, opened 2021-07-27
Output link/meta tags depending on web app manifest display value

jefferyto:
This changes when the (`apple-`)`mobile-web-app-capable` meta tags (and `apple-touch-startup-image` link tag) are printed to match the display mode set in the web app manifest.

This also removes the `apple-touch-fullscreen` meta tag, as this tag has the same effect as the `apple-mobile-web-app-capable` meta tag but is not included in any Safari documentation.

### DIFF
--- a/wp-includes/class-wp-web-app-manifest.php
+++ b/wp-includes/class-wp-web-app-manifest.php
@@ -74,17 +74,27 @@ final class WP_Web_App_Manifest {
 		?>
 		<link rel="manifest" href="<?php echo esc_url( static::get_url() ); ?>">
 		<meta name="theme-color" content="<?php echo esc_attr( $manifest['theme_color'] ); ?>">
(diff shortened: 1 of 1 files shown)


jefferyto:
@googlebot I signed it!

westonruter:
> This also removes the `apple-touch-fullscreen` meta tag, as this tag has the same effect as the `apple-mobile-web-app-capable` meta tag but is not included in any Safari documentation.

Yes, I can confirm that at https://stackoverflow.com/a/10040966/93579

westonruter (approved):
Thanks!

pooja-muchandikar:
QA Passed ✅ 

On registering new plugin with following code was able to print meta tags based on the display value set in the code.

Screenshot 2022-04-20 at 4 48 59 PM

---

Screenshot 2022-04-20 at 4 46 40 PM

PR #595, opened 2021-08-11
Add apple_touch_startup_images filter

### LINKED ISSUE #586: Make apple-touch-startup-image link tags filterable
Not sure if there is interest in merging a PR for this, so I decided to open an issue first.

Apparently, the splash image for iOS needs to be sized exactly to the screen to be displayed. The [official documentation](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW6) doesn't mention this, but there are lots of articles/blog posts (like [this one](https://medium.com/appscope/adding-custom-ios-splash-screens-to-your-progressive-web-app-41a9b18bdca3) and [this one](https://www.simicart.com/blog/pwa-splash-screen/#For_iOS)) that do cover this constraint.

This leads to a [horrible mess of link tags](https://gist.github.com/EvanBacon/7fd4dc3be3d00096579bb0b134c56ec7) to match every possible Apple device and [entire](https://github.com/onderceylan/pwa-asset-generator) [projects](https://appsco.pe/developer/splash-screens) dedicated to generating these images.

Instead of managing this here, I suggest adding a filter (maybe `apple_touch_startup_image`) that plugins/themes can use to add all the splash screens they want.

jefferyto:
This adds an `apple_touch_startup_images` filter for plugins/themes to adjust the `rel="apple-touch-startup-image"` link tags outputted. This allows more images to be added, with an applicable `media` attribute for each image.

Fixes https://github.com/GoogleChromeLabs/pwa-wp/issues/586.

### DIFF
--- a/wp-includes/class-wp-web-app-manifest.php
+++ b/wp-includes/class-wp-web-app-manifest.php
@@ -83,15 +83,48 @@ final class WP_Web_App_Manifest {
 				?>
 				<meta name="apple-mobile-web-app-capable" content="yes">
 				<meta name="mobile-web-app-capable" content="yes">
+
 				<?php
 				$icons = isset( $manifest['icons'] ) ? $manifest['icons'] : array();
 				usort( $icons, array( $this, 'sort_icons_callback' ) );
 				$icon = array_shift( $icons );
-				?>
-				<?php if ( ! empty( $icon ) ) : ?>
-					<link rel="apple-touch-startup-image" href="<?php echo esc_url( $icon['src'] ); ?>">
-				<?php endif; ?>
-				<?php
+
+				$images = array();
+				if ( ! empty( $icon ) ) {
+					$images[] = array( 'href' => $icon['src'] );
+				}
+
+				/**
+				 * Filters splash screen images for Safari on iOS.
+				 *
+				 * @param array $images {
+				 *     Array of splash screen images and their attributes.
+				 *
+				 *     @type array ...$0 {
+				 *         Array of splash screen image attributes.
+				 *
+				 *         @type string $href  URL of splash screen image. Required.
+				 *         @type string $media Media query for when the splash screen image should be used.
+				 *     }
+				 * }
+				 */
+				$images = apply_filters( 'apple_touch_startup_images', $images );
+
+				foreach ( $images as $key => $image ) {
+					if ( ! is_array( $image ) ) {
+						continue;
+					}
+
+					if ( ! isset( $image['href'] ) || ! esc_url( $image['href'], array( 'http', 'https' ) ) ) {
+						continue;
+					}
+
+					printf( '<link rel="apple-touch-startup-image" href="%s"', esc_url( $image['href'] ) );
+					if ( isset( $image['media'] ) ) {
+						printf( ' media="%s"', esc_attr( $image['media'] ) );
+					}
+					echo ">\n";
+				}
 				break;
 		endswitch;
 		?>
--- a/tests/test-class-wp-web-app-manifest.php
+++ b/tests/test-class-wp-web-app-manifest.php
@@ -97,9 +97,43 @@ class Test_WP_Web_App_Manifest extends WP_UnitTestCase {
 	 * @covers WP_Web_App_Manifest::manifest_link_and_meta()
 	 */
 	public function test_manifest_link_and_meta() {
+		$this->mock_site_icon();
+
+		$added_images = array(
+			array(
+				'href'  => home_url( '/340.png' ),
+				'media' => '(device-width: 340px)',
+			),
+			array(
+				'href' => home_url( '/480.png' ),
+			),
+		);
+
+		add_filter(
+			'apple_touch_startup_images',
+			function ( $images ) use ( $added_images ) {
+				$this->assertInternalType( 'array', $images );
+				$this->assertCount( 1, $images );
+				$images   = array_merge( $images, $added_images );
+				$images[] = array( 'bad' => 'yes' );
+				return $images;
+			}
+		);
+
 		ob_start();
 		$this->instance->manifest_link_and_meta();
 		$output = ob_get_clean();
+
+		$this->assertSame( 3, substr_count( $output, '<link rel="apple-touch-startup-image"' ) );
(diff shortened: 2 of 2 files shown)


jefferyto:
The test failure for "Unit test: PHP 8.0, WP trunk" appears unrelated to this change.

westonruter:
> The test failure for "Unit test: PHP 8.0, WP trunk" appears unrelated to this change.

Yeah. That's something I need to fix along the lines of what @pierlon has done in https://github.com/ampproject/amp-wp/pull/6525.

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -83,15 +83,51 @@ public function manifest_link_and_meta() {
...
+				 * }
+				 */
+				$images = apply_filters( 'apple_touch_startup_image', $images );
+
+				foreach ( $images as $key => $image ) {
+					if ( ! is_array( $image ) ) {
+						continue;
+					}
+
+					$url = isset( $image['href'] ) ? esc_url( $image['href'], array( 'http', 'https' ) ) : '';
```
westonruter:
```suggestion
					$url = isset( $image['href'] ) ? esc_url_raw( $image['href'], array( 'http', 'https' ) ) : '';
```

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -83,15 +83,51 @@ public function manifest_link_and_meta() {
...
+					if ( ! $url ) {
+						continue;
+					}
+
+					$html = sprintf( 'rel="apple-touch-startup-image" href="%s"', $url );
```
westonruter: This seems safer to be more resilient against future changes which could accidentally make `$url` cause escaping issues.

```suggestion
					$html = sprintf( 'rel="apple-touch-startup-image" href="%s"', esc_attr( $url ) );
```

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -83,15 +83,51 @@ public function manifest_link_and_meta() {
...
+				 *
+				 *     @type array ...$0 {
+				 *         Array of splash screen image attributes.
+				 *
+				 *         @type string $href  URL of splash screen image. Required.
+				 *         @type string $media Media query for when the splash screen image should be used.
+				 *     }
+				 * }
+				 */
+				$images = apply_filters( 'apple_touch_startup_image', $images );
```
westonruter: Should this be plural?

```suggestion
				$images = apply_filters( 'apple_touch_startup_images', $images );
```

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -83,15 +83,51 @@ public function manifest_link_and_meta() {
...
+				 *
+				 * @param array $images {
+				 *     Array of splash screen images and their attributes.
+				 *
+				 *     @type array ...$0 {
+				 *         Array of splash screen image attributes.
+				 *
+				 *         @type string $href  URL of splash screen image. Required.
+				 *         @type string $media Media query for when the splash screen image should be used.
+				 *     }
```
westonruter, line 108: Nice. I don't recall seeing this phpdoc formatting before, but I see a few examples of `...$0` in core.

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -83,15 +83,51 @@ public function manifest_link_and_meta() {
...
+						continue;
+					}
+
+					$html = sprintf( 'rel="apple-touch-startup-image" href="%s"', $url );
+					if ( isset( $image['media'] ) ) {
+						$html .= sprintf( ' media="%s"', esc_attr( $image['media'] ) );
+					}
+					?>
+					<link <?php echo $html; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>>
+					<?php
```
westonruter: It would seem a bit better to do:

```suggestion
					printf( '<link rel="apple-touch-startup-image" href="%s"', esc_attr( $url ) );
					if ( isset( $image['media'] ) ) {
						printf( ' media="%s"', esc_attr( $image['media'] ) )
					}
					echo '>';
```

In this way the tag markup is all constructed together and there is no need to suppress any phpcs warnings.

westonruter (changes requested):

wp-includes/class-wp-web-app-manifest.php
jefferyto: I think you're suggesting `esc_url_raw()` here because you suggest `esc_attr( $url )` further down but I'm not sure this treatment is equal to `esc_url()`.

wp-includes/class-wp-web-app-manifest.php
jefferyto: I think whether this is more resilient is debatable (I would argue `esc_attr( $url )` looks confusing to most WordPress developers since all documentation says to use `esc_url()` for URLs), but I will rewrite this code to separate the URL used for validation and the URL used for display.

wp-includes/class-wp-web-app-manifest.php
jefferyto: I can get more pedantic here but I'm okay with making this plural.

wp-includes/class-wp-web-app-manifest.php
jefferyto: I based this code on [`wp_resource_hints()`](https://github.com/WordPress/wordpress-develop/blob/710f30eb966091408ebb89c276bec6a8f6f12a1d/src/wp-includes/general-template.php#L3384) (curious how that line doesn't trigger phpcs warnings) and I think separating HTML generation and output looks cleaner, but I will make this change.

wp-includes/class-wp-web-app-manifest.php
westonruter: Humm, yeah. I guess `esc_url()` is not the same as `esc_url_raw()`+`esc_attr()` after looking at the source. So nevermind.

westonruter (approved):

westonruter:
@googlebot I consent.

PR #691, opened 2022-02-08
Add short name field in General Settings when Site Title is longer than 12 characters

### LINKED ISSUE #210: Add Short Name as admin settings field alongside Site Title
The web app manifest removed the `short_name` because it was not reliable to automatically generate: https://github.com/xwp/pwa-wp/pull/201

In order to avoid a site having to add a `web_app_manifest` filter to supply the `short_name`, what if a new Short Name field is added after the Site Title:

image

This field could be hidden until the Site Title field has more than 12 characters in it.

westonruter:
Fixes #210.

This adds a new Short Name field to the General Settings screen:

image

This field is required if the Site Title is longer than 12 characters:

image

If if the Site Title is less than or equal to 12 characters long, then the `placeholder` value is set to match the Site Title (since it is used by default in the manifest in this case), and the field is not `required`:

image

If the Site Title is greater than 12 characters in length, a user must supply a Short Name to save the form (and the field has a `maxlength` of 12 supplied, which is enforced by the `sanitize_callback`):

image

If a site is supplying the Short Name via a plugin, like as follows (which was the only way to do so before now):

```php
add_filter( 'web_app_manifest', static function ( $manifest ) {
	$manifest['short_name'] = 'WP Dev';
	return $manifest;
} );
```

Then the Short Name field is disabled in the same way the Site URL and Home URL fields if they are defined with constants:

image

The Site Health test for the Short Name now includes a link to the General Settings screen to manage the field:

image

Other changes:

* Use `mb_strlen()` for `short_name` check.
* Use `trim()` on a short name before checking the length or adding to the manifest.
* Enable client-side form validation on the General Settings screen. (It was disabled for a bug in an 8-year old version of Firefox, per TRAC-22183.)

### DIFF
--- a/tests/test-class-wp-web-app-manifest.php
+++ b/tests/test-class-wp-web-app-manifest.php
@@ -162,6 +162,35 @@ class Test_WP_Web_App_Manifest extends TestCase {
 	}
 
 	/**
+	 * Data provider.
+	 *
+	 * @return array
+	 */
+	public function get_data_to_test_is_name_short() {
+		return array(
+			array( '0123456789ab', true ),
+			array( 'áâàéêèíîìóôò', true ),
+			array( ' áâàéêèíîìóôò ', true ),
+			array( ' 0123456789ab ', true ),
+			array( '0123456789abc', false ),
+			array( 'áâàéêèíîìóôò2', false ),
+		);
+	}
+
+	/**
+	 * Test is_name_short.
+	 *
+	 * @dataProvider get_data_to_test_is_name_short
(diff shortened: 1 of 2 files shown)


westonruter:
Still need to add tests, but otherwise this is ready for review.

- [x] Add tests.

westonruter:
- [x] Add link from Site Health test to Short Name field on General Settings screen.

thelovekesh (approved):
LGTM 👍

PR #697, opened 2022-02-14
Add page reloading in offline.php templates

### LINKED ISSUE #438: Automatically reload the offline page once the user is back online
See https://github.com/GoogleChrome/web.dev/pull/4717

thelovekesh:
Fixes #438  

Automatically reload the offline page once the user is back online.

Tasks Done:

- [x] Add JS to automatically reload the offline page once the user is back online

https://user-images.githubusercontent.com/54371619/152957784-e01b9259-ccde-46f3-8f85-4d0404e7b436.mp4

### DIFF
--- a/wp-includes/template.php
+++ b/wp-includes/template.php
@@ -186,7 +186,6 @@ function wp_service_worker_offline_page_reload() {
 		return;
 	}
 
-	ob_start();
 	?>
 	<script type="module">
 		/**
@@ -220,7 +219,6 @@ function wp_service_worker_offline_page_reload() {
 		checkNetworkAndReload();
 	</script>
 	<?php
-	return ob_get_clean();
 }
 
 add_action( 'wp_footer', 'wp_service_worker_offline_page_reload' );
--- a/wp-includes/template.php
+++ b/wp-includes/template.php
@@ -171,3 +171,47 @@ function wp_service_worker_error_details_template( $output = '' ) {
 function wp_service_worker_error_message_placeholder() {
 	echo '<p><!--WP_SERVICE_WORKER_ERROR_MESSAGE--></p>';
 }
+
+/**
+ * Reload the offline page and check if user comes online.
+ *
+ * @since 0.7
+ */
+function wp_service_worker_offline_page_reload() {
+	if ( ! is_offline() ) {
+		return;
+	}
+	?>
+	<script>
+		/**
+		 * Listen to changes in the network state, reload when online.
+		 * This handles the case when the device is completely offline.
+		 */
+		window.addEventListener('online', () => {
+			window.location.reload();
+		});
+
+		/**
+		 * Check if the server is responding and reload the page if it is.
+		 * This handles the case when the device is online, but the server is offline or misbehaving.
+		 */
+		async function checkNetworkAndReload() {
+			try {
+				const response = await fetch('.', {method: 'HEAD'});
+				// Verify we get a valid response from the server
+				if (response.status >= 200 && response.status < 500) {
+					window.location.reload();
+					return;
+				}
+			} catch {
+				console.log('User is offline');
+			}
+			window.setTimeout(checkNetworkAndReload, 2500);
+		}
+		checkNetworkAndReload();
+	</script>
+	<?php
+}
+
+add_action( 'wp_footer', 'wp_service_worker_offline_page_reload' );
+add_action( 'error_footer', 'wp_service_worker_offline_page_reload' );
--- a/wp-includes/template.php
+++ b/wp-includes/template.php
@@ -186,6 +186,7 @@ function wp_service_worker_offline_page_reload() {
 		return;
 	}
 
+	ob_start();
 	?>
 	<script type="module">
 		/**
@@ -219,6 +220,7 @@ function wp_service_worker_offline_page_reload() {
 		checkNetworkAndReload();
 	</script>
 	<?php
+	return ob_get_clean();
 }
 
 add_action( 'wp_footer', 'wp_service_worker_offline_page_reload' );
--- a/wp-includes/template.php
+++ b/wp-includes/template.php
@@ -200,7 +200,7 @@ function wp_service_worker_offline_page_reload() {
 		 */
 		async function checkNetworkAndReload() {
 			try {
-				const response = await fetch('.', {method: 'HEAD'});
+				const response = await fetch(location.href, {method: 'HEAD'});
 				// Verify we get a valid response from the server
 				if (response.status >= 200 && response.status < 500) {
 					window.location.reload();
--- a/wp-includes/template.php
+++ b/wp-includes/template.php
@@ -178,9 +178,14 @@ function wp_service_worker_error_message_placeholder() {
  * @since 0.7
  */
 function wp_service_worker_offline_page_reload() {
-	if ( ! is_offline() ) {
+	if ( isset( $_SERVER['REQUEST_METHOD'] ) && 'GET' !== $_SERVER['REQUEST_METHOD'] ) {
 		return;
 	}
+
+	if ( ! is_offline() && ! is_500() ) {
+		return;
+	}
+
 	?>
 	<script type="module">
 		/**
@@ -207,7 +212,7 @@ function wp_service_worker_offline_page_reload() {
 					return;
 				}
 			} catch {
-				console.log('User is offline');
+				// Unable to connect so do nothing.
 			}
 			window.setTimeout(checkNetworkAndReload, Math.pow(2, count++) * 2500);
 		}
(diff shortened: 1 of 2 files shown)


wp-includes/template.php
```diff
@@ -171,3 +171,47 @@ function wp_service_worker_error_details_template( $output = '' ) {
...
+			window.location.reload();
+		});
+
+		/**
+		 * Check if the server is responding and reload the page if it is.
+		 * This handles the case when the device is online, but the server is offline or misbehaving.
+		 */
+		async function checkNetworkAndReload() {
+			try {
+				const response = await fetch('.', {method: 'HEAD'});
```
westonruter: Any reason for `.` as opposed to `location.href`? The problem with `.` is it will drop the query parameters. 

```suggestion
				const response = await fetch(location.href, {method: 'HEAD'});
```

wp-includes/template.php
```diff
@@ -171,3 +171,47 @@ function wp_service_worker_error_details_template( $output = '' ) {
...
+				// Verify we get a valid response from the server
+				if (response.status >= 200 && response.status < 500) {
+					window.location.reload();
+					return;
+				}
+			} catch {
+				console.log('User is offline');
+			}
+			window.setTimeout(checkNetworkAndReload, 2500);
```
westonruter: Any reason specifically for 2,500ms? I suppose just because this is what was used in https://github.com/GoogleChrome/web.dev/pull/4717.

If the server is having trouble due to heavy traffic or some other reason, getting hit every 2.5 seconds will add additional load and may be interpreted by the host as a DOS attack. Therefore, I suggest the value here implement [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) for retrying. So there could be a counter in the script for the number of tries, and each time the logic runs and the counter is incremented, the next timeout delay can be multiplied by the counter. 

Count | Delay
------|------
1 | 2.5s
2 | 5s
3 | 10s
4 | 20s
5 | 40s

In other words, multiple the 2500 delay by `Math.pow( 2, count )` where the initial count is `0`.

wp-includes/template.php
```diff
@@ -171,3 +171,47 @@ function wp_service_worker_error_details_template( $output = '' ) {
...
+ * Reload the offline page and check if user comes online.
+ *
+ * @since 0.7
+ */
+function wp_service_worker_offline_page_reload() {
+	if ( ! is_offline() ) {
+		return;
+	}
+	?>
+	<script>
```
westonruter: Should this go through JS Coding Guidelines? https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/

Or should we rather apply Prettier formatting?

wp-includes/template.php
```diff
@@ -171,3 +171,47 @@ function wp_service_worker_error_details_template( $output = '' ) {
...
 	echo '<p><!--WP_SERVICE_WORKER_ERROR_MESSAGE--></p>';
 }
+
+/**
+ * Reload the offline page and check if user comes online.
+ *
+ * @since 0.7
+ */
+function wp_service_worker_offline_page_reload() {
+	if ( ! is_offline() ) {
```
westonruter: It would be useful to also do the reload in the case of server being down.

Also, something else that comes to mind: what happens if the error page is served in response to a POST request? If so, each reload will cause a re-POST message from the browser and that's probably not desired.

```suggestion
	if ( ! is_offline() || ! is_500() || 'GET' !== $_SERVER['REQUEST_METHOD'] ) {
```

wp-includes/template.php
```diff
@@ -171,3 +171,47 @@ function wp_service_worker_error_details_template( $output = '' ) {
...
+				if (response.status >= 200 && response.status < 500) {
+					window.location.reload();
+					return;
+				}
+			} catch {
+				console.log('User is offline');
+			}
+			window.setTimeout(checkNetworkAndReload, 2500);
+		}
+		checkNetworkAndReload();
```
westonruter, line 219: Either the entire script should be wrapped in an IIFE, or the `<script>` should have `type="module"` added to it. This will prevent the `checkNetworkAndReload` function from polluting the global scope.

westonruter (changes requested):

wp-includes/template.php
thelovekesh: > apply Prettier formatting

~~Sounds good to me.~~

I think JS code in PHP files that follows WPCS should follow WP JS standards.

wp-includes/template.php
thelovekesh: That's interesting. Above we are making a fetch request to base origin to check whether we are getting response OK or not. I wonder how making a fetch request to some origin can drop query params of another origin.

wp-includes/template.php
thelovekesh: I think we should also reset the counter after the count is quals to 5 or 6. There can be a possibility that the user or server comes back online and page reloading has to wait for another 1 or 2 minutes due to time delay. What do you think?

ankitrox: @thelovekesh We can cap the number of retries, so after `n` number of retries, there won't be more.

wp-includes/template.php
thelovekesh: Thanks for pointing this out. For sure we need to check the method of the request. I have implemented these changes.

wp-includes/template.php
westonruter: Note that core may be moving to Prettier: https://make.wordpress.org/core/2022/01/05/proposal-changes-to-javascript-coding-standards-for-full-prettier-compatibility/

westonruter:
Note that testing this is a bit tricky because the script is part of the precached template, so without a version bump it won't show up.

westonruter:
Let's just add a PHPUnit test to ensure that the cases are covered for this function outputting the `script`.

wp-includes/template.php
thelovekesh: @westonruter Please let me know your thoughts on it.

westonruter: I think the exponential backoff will be enough. After 15 tries the timeout goes up to ~1 day, so it will effectively be disabled once the count goes high enough.

thelovekesh:
> Note that testing this is a bit tricky because the script is part of the precached template, so without a version bump it won't show up.

@westonruter  Tests added 👍

wp-includes/template.php
```diff
@@ -171,3 +171,57 @@ function wp_service_worker_error_details_template( $output = '' ) {
...
+				}
+			} catch {
+				// Unable to connect so do nothing.
+			}
+			window.setTimeout(checkNetworkAndReload, Math.pow(2, count++) * 2500);
+		}
+		checkNetworkAndReload();
+	</script>
+	<?php
+	return ob_get_clean();
```
westonruter: Why adding the return here? If I'm not mistaken, the output buffering is causing the `script` to no longer be printed onto the page? The output buffering should rather be done in the tests.

thelovekesh: > output buffering is causing the script to no longer be printed

@westonruter Ahh. my bad. Might get confused during the testing. 
Fixed this now 👍

thelovekesh:
@westonruter Can you please confirm if this PR needs no more improvements?

westonruter (approved):
This is good, yes.

westonruter:
I just realized one thing that should be done to improve this. Namely, if you try going to `/?wp_error_template=offline` then the result is an infinite reload as it tries to go back online. Accessing this URL directly is key for being able to see how the offline template looks during development. So I think the logic to auto-reload should be disabled if this is true:

```js
( new URLSearchParams(location.search.substr(1)) ).has( 'wp_error_template' )
```

thelovekesh:
> then the result is an infinite reload 

Thanks for pointing it out and it's a great finding. I just checked it after your comment. We should add it else it will cause infinite reloads during development. 

@westonruter Should I push a commit for this?

westonruter:
@thelovekesh Yes please! Thanks.

wp-includes/template.php
westonruter: As part of #728, I just realized that this check will never do anything:

https://github.com/GoogleChromeLabs/pwa-wp/blob/124baabe30f70b718960fa18158f10a54943287e/wp-includes/template.php#L181-L184

This is because the offline template is always fetched via a `GET` request by the service worker when it is installing, so it will never not be `GET`. We need an alternative way of preventing the code from running on responses to non-GET requests.

Contrary to what I thought above, each reload does _not_ cause the browser re-POST message: https://github.com/GoogleChromeLabs/pwa-wp/pull/728#issuecomment-1099718487

westonruter: I believe I've addressed this via https://github.com/GoogleChromeLabs/pwa-wp/pull/763

westonruter: See https://github.com/GoogleChromeLabs/pwa-wp/pull/763

PR #698, opened 2022-02-14
Add maskable icon checkbox in customizer settings

### LINKED ISSUE #304: Facilitate indicating the site icon as being maskable
As reported on the [support forum](https://wordpress.org/support/topic/manifest-doesnt-have-a-maskable-icon-3/), when doing a Lighthouse PWA audit, the web app manifest generated by the plugin currently gets a warning:

> Manifest doesn't have a maskable icon

Read more about this on web.dev: [Manifest doesn't have a maskable icon](https://web.dev/maskable-icon-audit/).

A site owner can fix this problem by just adding the `purpose` property to all of the icons:

```php
add_filter( 'web_app_manifest', function( $manifest ) {
	$manifest['icons'] = array_map(
		function ( $icon ) {
			if ( ! isset( $icon['purpose'] ) ) {
				$icon['purpose'] = 'any maskable';
			}
			return $icon;
		},
		$manifest['icons']
	);
	return $manifest;
} );
```

The plugin could also automatically add this `maskable` property. But these both assume that the icon is actually maskable. Is this good to assume? What if we added a new checkbox after where the Site Icon is supplied in the Customizer to let the user indicate it is maskable:

image

Related: #246

ankitrox:
Closes #304


wp-admin/js/customizer.js
```diff
@@ -0,0 +1,70 @@
...
+		 * 4. Icon set as un-maskable.
+		 */
+		const iconUpdateListener = function () {
+			siteIcon = parseInt(this.iconId, 10);
+
+			// Check/uncheck maskable checkbox.
+			wp.customize('pwa_maskable_icon').set(this.checked);
+
+			if (!siteIcon) {
+				//alert( wp.customize.control( 'pwa_maskable_icon' ) );
```
westonruter:
```suggestion
```

wp-admin/js/customizer.js
```diff
@@ -0,0 +1,70 @@
+(function () {
```
westonruter: This IIFE isn't necessary.

wp-includes/class-wp-customize-manager.php
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @return void
+ * @since 0.7
+ */
+function pwa_customize_register_maskable_icon_setting( WP_Customize_Manager $wp_customize ) {
+	$wp_customize->add_setting(
+		'pwa_maskable_icon',
+		array(
+			'capability' => 'manage_options',
+			'type'       => 'option',
```
westonruter, line 27: Since it is a specific icon that is maskable, and the icon is an image uploaded to the media library, I wonder if it would be preferable to store this maskable flag as postmeta instead. That may be overkill.

wp-includes/class-wp-customize-manager.php
```diff
@@ -0,0 +1,62 @@
...
+add_action( 'customize_register', 'pwa_customize_register_maskable_icon_setting' );
+
+/**
+ * Enqueue scripts for maskable icon customizer setting.
+ *
+ * @return void
+ */
+function pwa_maskable_icon_scripts() {
+	wp_enqueue_script(
+		'pwa_customizer_script',
```
westonruter:
```suggestion
		'customize-controls-maskable-icon',
```

wp-includes/class-wp-customize-manager.php
```diff
@@ -0,0 +1,62 @@
...
+	$wp_customize->add_control(
+		'pwa_maskable_icon',
+		array(
+			'type'            => 'checkbox',
+			'section'         => 'title_tagline',
+			'label'           => __( 'Maskable icon', 'pwa' ),
+			'priority'        => 100,
+			'active_callback' => function() {
+				return (bool) get_option( 'site_icon' );
+			},
```
westonruter: I think this can be managed exclusively in JS.

wp-admin/js/customizer.js
```diff
@@ -0,0 +1,70 @@
...
+				return;
+			}
+
+			// At this point we are sure that icon is set, thus activate control.
+			wp.customize.control('pwa_maskable_icon').activate();
+
+			const iconPreview = document.querySelector('img.app-icon-preview');
+
+			if (iconPreview) {
+				document.querySelector('img.app-icon-preview').style.clipPath =
```
westonruter:
```suggestion
				iconPreview.style.clipPath =
```

wp-admin/js/customizer.js
```diff
@@ -0,0 +1,70 @@
...
+					this.checked && siteIcon ? 'inset(10% round 50%)' : '';
```
westonruter: The `siteIcon` check isn't needed because if it is not set then the `img` won't be in the document in the first place. Also, instead of `this.checked` you can just use `wp.customize('pwa_maskable_icon').get()`:

```suggestion
					wp.customize('pwa_maskable_icon').get() ? 'inset(10% round 50%)' : '';
```

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -235,6 +235,23 @@ public function get_manifest() {
...
 		}
 
+		/**
+		 * If user has not confirmed site icon as maskable, add icon purpose as `any maskable`.
+		 * By default, icons are maskable and lighthouse audits fails if icon is not really maskable.
+		 *
+		 * See <https://web.dev/maskable-icon-audit/>
+		 */
+		$manifest_icon_maskable = get_option( 'pwa_maskable_icon', false );
+		if ( ! $manifest_icon_maskable && isset( $manifest['icons'] ) ) {
```
westonruter: This logic seems inverted. It should only add `maskable` if the user has explicitly said that it _is_ maskable.

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -235,6 +235,23 @@ public function get_manifest() {
...
+			$manifest['icons'] = array_map(
+				function ( $icon ) {
+					$icon['purpose'] = 'any maskable';
+					return $icon;
+				},
+				$manifest['icons']
+			);
+		}
```
westonruter: This logic would be better located in the `get_icons` method.

tests/test-class-wp-web-app-manifest.php
```diff
@@ -64,6 +64,7 @@ class Test_WP_Web_App_Manifest extends TestCase {
 	public function setUp() {
 		parent::setUp();
 		$this->instance = new WP_Web_App_Manifest();
+		update_option( 'pwa_maskable_icon', true );
```
westonruter: After thinking about this some more, I don't think we should assume the icon is maskable. I had written this in the issue:

> But these both assume that the icon is actually maskable. Is this good to assume?

If we were to default maskable to true, then the option should rather be called something like `site_icon_unmaskable` since it would be empty by default, unless we explicitly set it as part of a database migration. But in any case, I think we should explicitly require the user to click the checkbox to say that it is maskable.

westonruter (changes requested):

wp-includes/class-wp-customize-manager.php
ankitrox: @westonruter `site_icon` is getting store in `*_options` table and maskable property is associated with `site_icon` itself. IMO, it makes sense to store `maskable` in `options` as there will be only one `site_icon` in customizer.

westonruter: But if someone picks a different image for the site icon, then it may not be maskable. So that's why I was thinking it could be useful to associate the “maskability“ with the image `attachment` post specifically. But this would add quite a bit of complexity and it isn't necessary as the user would clearly see that the icon is maskable or not upon selecting a new image.

ankitrox: @westonruter Before your code changes, when user was picking new image for site icon, I was resetting the maskable toggle value.

Here, we can consider that every newly uploaded image may not be maskable and if user want he can explicitly mark the maskable input. 

Only drawback of this approach is next time user selects previously uploaded icon, check box will not be marked/unmarked based on last input for that icon.

IMO, considering the tradeoff of simplicity vs experience, for now we can proceed with `options` and may change it in future if such requests are raised.

ankitrox:
@westonruter If all seems good, can we proceed with merging this one?

westonruter (approved):
Yes, sorry for the delay.

PR #713, opened 2022-02-28
Deprecate integrations

### LINKED ISSUE #403: Deprecate integrations

thelovekesh:
Fixes #403 

Tasks Done:

- [x] Add function to show user a warning about integrations deprecation.
- [x] Add `@deprecated` tags in functions and classes.

Screenshot:
image

### DIFF
--- a/integrations/class-wp-service-worker-admin-assets-integration.php
+++ b/integrations/class-wp-service-worker-admin-assets-integration.php
@@ -10,7 +10,7 @@
  *
  * @since 0.2
  *
- * @deprecated 0.7 Integrations will be removed in the future versions.
+ * @deprecated 0.7 Integrations will not be proposed for WordPress core merge.
  */
 final class WP_Service_Worker_Admin_Assets_Integration extends WP_Service_Worker_Base_Integration {
 
--- a/integrations/class-wp-service-worker-admin-assets-integration.php
+++ b/integrations/class-wp-service-worker-admin-assets-integration.php
@@ -46,6 +46,26 @@ final class WP_Service_Worker_Admin_Assets_Integration extends WP_Service_Worker
 			unset( $options['url'] );
 			$scripts->precaching_routes()->register( $url, $options );
 		}
+
+		// Add deprecation warning in user's console when service worker is installed.
+		$scripts->register(
+			__CLASS__ . '-deprecation',
+			array(
+				'src' => static function () {
+					return sprintf(
+						'console.warn( %s );',
+						wp_json_encode(
+							sprintf(
+								/* translators: %1$s: integration class name, %2$s: issue url */
+								__( 'The %1$s integration in the PWA plugin is no longer being considered WordPress core merge. See %2$s', 'pwa' ),
+								__CLASS__,
+								'https://github.com/GoogleChromeLabs/pwa-wp/issues/403'
+							)
+						)
+					);
+				},
+			)
+		);
 	}
 
 	/**
--- a/integrations/class-wp-service-worker-admin-assets-integration.php
+++ b/integrations/class-wp-service-worker-admin-assets-integration.php
@@ -9,7 +9,6 @@
  * Class representing the admin assets service worker integration.
  *
  * @since 0.2
- *
  * @deprecated 0.7 Integrations will not be proposed for WordPress core merge.
  */
 final class WP_Service_Worker_Admin_Assets_Integration extends WP_Service_Worker_Base_Integration {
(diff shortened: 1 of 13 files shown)


thelovekesh:
@westonruter I have added `@deprecated` tags to function/classes and a function to show warnings in the user console as you instructed. Please let me know if this PR needs more improvements.

integrations/class-wp-service-worker-admin-assets-integration.php
```diff
@@ -9,6 +9,8 @@
  * Class representing the admin assets service worker integration.
  *
  * @since 0.2
+ *
+ * @deprecated 0.7 Integrations will be removed in the future versions.
```
westonruter: Let's rename this and the following to: `Integrations will not be proposed for WordPress core merge.`

integrations/interface-wp-service-worker-integration.php
```diff
@@ -17,6 +19,8 @@ interface WP_Service_Worker_Integration {
 	 *
 	 * @since 0.2
 	 *
+	 * @deprecated 0.7 Integrations will be removed in the future versions.
```
westonruter: Are deprecation notices needed on each method since the class as a while has it?

pwa.php
```diff
@@ -290,6 +290,61 @@ function pwa_load_service_worker_integrations( WP_Service_Worker_Scripts $script
...
+			printf(
+				'console.warn( "%s" );',
+				esc_js( $message )
+			);
+		}
+		?>
+	</script>
+	<?php
+}
+add_action( 'wp_footer', 'pwa_deprecate_service_worker_integrations', 1000 );
```
westonruter: Instead of printing the warnings on the frontend, what about adding them to the integrations themselves to be logged from the service worker? A benefit of this is that they'd only get logged-out when the service worker is installed as opposed to being added with every page load. So something like this:

```diff
--- a/integrations/class-wp-service-worker-admin-assets-integration.php
+++ b/integrations/class-wp-service-worker-admin-assets-integration.php
@@ -44,6 +44,24 @@ final class WP_Service_Worker_Admin_Assets_Integration extends WP_Service_Worker
 			unset( $options['url'] );
 			$scripts->precaching_routes()->register( $url, $options );
 		}
+
+		$scripts->register(
+			__CLASS__ . '-deprecation',
+			array(
+				'src' => static function () {
+					return sprintf(
+						'console.warn( %s );',
+						wp_json_encode(
+							sprintf(
+								__( 'The %1$s integration in the PWA plugin is no longer being considered WordPress core merge. See %2$s', 'pwa' ),
+								__CLASS__,
+								'https://github.com/GoogleChromeLabs/pwa-wp/issues/403'
+							)
+						)
+					);
+				}
+			)
+		);
 	}
 
 	/**
```

westonruter (changes requested):

integrations/interface-wp-service-worker-integration.php
thelovekesh: Since deprecation notice has been added to the Interface itself we don't need them on methods. But one more thing can be helpful like ~~register~~() can gain more dev attention if he is using methods from deprecated interface. 

@westonruter Need your review on this one. Please let me know if we need to remove the notice from the methods.

pwa.php
thelovekesh: This seems more suitable to me. Thanks for this good suggestion.

integrations/interface-wp-service-worker-integration.php
thelovekesh: @westonruter Need your suggestion here.
 In case you missed.

thelovekesh: @westonruter If the above comment justifies the addition of a notice on methods then I think we are good to merge this PR. If not then please let me know so that I can push a commit to remove them.

westonruter (approved):

Issue #363, opened 2020-12-12
Disable offline commenting and offline browsing for authenticated users


westonruter:
There are two scenarios where offline commenting can currently fail due to user authentication:

When a site requires that “Users must be registered and logged in to comment”, when a user does try to submit a comment and it is queued via background sync, it could be that once the user goes back online their session will have expired. At that point, there comment will lost.

Otherwise, even when non-authenticated users can post comments, when an admin user is logged-in, the comment form includes a `_wp_unfiltered_html_comment` input that has a nonce:

![image](https://user-images.githubusercontent.com/134745/101968135-6b0b3680-3bd2-11eb-83a3-d05cdb23ad9e.png)

If an admin user submits a comment when offline and the comment is submitted after the session has expired, then if the user had included any privileged HTML in the comment, this will get stripped out by Kses filtering upon submission.

westonruter:
Offline commenting should probably be turned off entirely for authenticated users.

westonruter:
Offline browsing overall should probably be turned off for authenticated users, as currently this can cause issues with the admin bar being cached. See #252.

westonruter:
Another reason to turn off offline browsing for logged-in users: nonce links can fail.

rutviksavsani:
@westonruter  we will be taking this issues further on,
As per my discussion with @ankitrox  and @thelovekesh, we have analyzed the following approaches for working on this task.

Initial approaches :

1) Executing user authentication through offline commenting and navigation routing scripts.
**Issue Faced:**  
As we have those scripts in Service workers, we would not be able to access Cookies or DOM, which will prevent us to authenticate the users to allow offline commenting for the Non-authenticated users conditionally.

2) Through the Service workers PHP class
**Issue Faced:**  
In this case, to serve the same service workers to all the users, the User is set to none while registering the service workers and hence preventing us to authenticate the users to further conditionally load the offline commenting and navigation routing scripts in the Service worker.
ref: https://github.com/GoogleChromeLabs/pwa-wp/blob/develop/wp-includes/class-wp-service-workers.php#L125-L134

Suggested approach :

As mentioned above in the second approach. we need to conditionally load these scripts mentioned in https://github.com/GoogleChromeLabs/pwa-wp/blob/develop/wp-includes/components/class-wp-service-worker-navigation-routing-component.php#L298-L314
[...]

PR #717, opened 2022-03-01
Remove novalidate attribute from comment form to enable client-side validation

### LINKED ISSUE #364: Offline commenting can fail if the user does not provide valid fields
The comment form has a `novalidate` attribute to force server-side validation of the fields. This is done to bypass HTML5 client-side form validation, since it used to be that the validation was not reliable across browsers. Even so, `novalidate` remains. So at the moment someone could enter a comment but then leave the Name or Email empty, and submit the comment while offline. Once the user is back online, background sync will try to submit the comment but it will be rejected by WordPress due to not having a name or email. 

So the PWA plugin should remove the `novalidate` attribute from the comment form as I suggested in TRAC-47595.

rutviksavsani:
Fixes: #364 
This PR adds an action to remove the `novalidate` attribute which is added by WordPress core in the comments form. Removing that attribute will enable client-side validation too for comment form which may not allow invalid data to be submitted.

## Screenshot
image

### DIFF
--- a/pwa.php
+++ b/pwa.php
@@ -299,7 +299,7 @@ add_action( 'wp_default_service_workers', 'pwa_load_service_worker_integrations'
 function pwa_remove_novalidate_attribute_from_comments_form() {
 	?>
 	<script type="module">
-		var commentForm = document.getElementById( 'commentform' );
+		const commentForm = document.getElementById( 'commentform' );
 		if ( commentForm ) {
 			commentForm.removeAttribute( 'novalidate' );
 		}
--- a/pwa.php
+++ b/pwa.php
@@ -290,6 +290,24 @@ function pwa_load_service_worker_integrations( WP_Service_Worker_Scripts $script
 }
 add_action( 'wp_default_service_workers', 'pwa_load_service_worker_integrations', -1 );
 
+/**
+ * Remove `novalidate` attribute from the comment form tag.
+ *
+ * This is a fix for the `novalidate` attribute being added to the comment form by WordPress
+ * which disables HTML client-side validation for the form.
+ */
+function pwa_remove_novalidate_attribute_from_comments_form() {
+	?>
+	<script type="module">
+		var commentForm = document.getElementById( 'commentform' );
+		if ( commentForm ) {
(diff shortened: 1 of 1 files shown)


westonruter:
If offline commenting is removed as suggested in https://github.com/GoogleChromeLabs/pwa-wp/issues/363#issuecomment-1051385128, then this wouldn't be strictly necessary. I'm not totally settled on the direction of that, but I'm leaning towards removal of the feature.

pwa.php
```diff
@@ -290,6 +290,24 @@ function pwa_load_service_worker_integrations( WP_Service_Worker_Scripts $script
...
+/**
+ * Remove `novalidate` attribute from the comment form tag.
+ *
+ * This is a fix for the `novalidate` attribute being added to the comment form by WordPress
+ * which disables HTML client-side validation for the form.
+ */
+function pwa_remove_novalidate_attribute_from_comments_form() {
+	?>
+	<script type="module">
+		var commentForm = document.getElementById( 'commentform' );
```
westonruter:
```suggestion
		const commentForm = document.getElementById( 'commentform' );
```

rutviksavsani:
> If offline commenting is removed as suggested in [#363 (comment)](https://github.com/GoogleChromeLabs/pwa-wp/issues/363#issuecomment-1051385128), then this wouldn't be strictly necessary. I'm not totally settled on the direction of that, but I'm leaning towards removal of the feature.

Yes, I also agree with your point for the removal of offline commenting, but until it is completely removed we can add this feature and also remove it along with the offline commenting feature.
We will be deprecating the `offline commenting` feature at first, hence we can add this as a supporting feature until then.

westonruter:
> We will be deprecating the `offline commenting` feature at first, hence we can add this as a supporting feature until then.

That's the thing. I think we may just remove it entirely because (1) it's broken when a user tries to comment while logged-in and their nonce expires, and (2) I don't believe anyone is using it (at least I've never heard of someone talking about this feature).

thelovekesh:
@westonruter Since we are removing the offline-commenting feature in #728, should we close this PR?

westonruter:
Yes, thanks.

PR #719, opened 2022-03-02
Add Deprecation warning for offline commenting feature.

### LINKED ISSUE #363: Disable offline commenting and offline browsing for authenticated users

rutviksavsani:
Fixes: #363

As discussed on issue #363, We will be removing the Offline commenting feature from the next release, Hence I have added a console warning for the deprecation on service worker registration.

image

### DIFF
--- a/wp-includes/components/class-wp-service-worker-navigation-routing-component.php
+++ b/wp-includes/components/class-wp-service-worker-navigation-routing-component.php
@@ -312,6 +312,27 @@ final class WP_Service_Worker_Navigation_Routing_Component implements WP_Service
 			)
 		);
 
+		/*
+		 * Add offline commenting deprecation warning in user's console when the service worker is being installed.
+		 */
+		$scripts->register(
+			'wp-offline-commenting-deprecation-warning',
+			array(
+				'src' => static function () {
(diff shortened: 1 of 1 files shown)


westonruter:
I think we really should just remove the feature altogether.

rutviksavsani:
@westonruter If you could go through this query on the Issue and provide your inputs.
https://github.com/GoogleChromeLabs/pwa-wp/issues/363#issuecomment-1057993599

thelovekesh:
@rutviksavsani Closing this PR as we now removing the offline commenting feature completely in #728

PR #728, opened 2022-03-10
Remove offline commenting feature and add post request handler to display offline/error templates.

### LINKED ISSUE #363: Disable offline commenting and offline browsing for authenticated users

rutviksavsani:
Fixes: #363 

This PR 
- Removes the Offline commenting feature. 
- Introduces a request handler to display offline/error for POST requests.

## Before
image (1)
## After
image


thelovekesh:
When a user makes a POST request and submission fails due to offline/server-error:

https://user-images.githubusercontent.com/54371619/158224688-8b774753-2323-45f0-9ec7-9648f6a3ffe2.mp4

When a user makes a POST request and it opens on a new page:

https://user-images.githubusercontent.com/54371619/158224457-72b5c771-bcd0-4950-a5b9-9d4c36931e6e.mp4

thelovekesh:
@westonruter On thinking more about to show [go back](#) link, two use cases came up to my mind:

1 - When POST requests use the same tab to fulfill the request. In this case `history.back()` will work and go back to last occurrence that will be having bfcache.
2- When the user makes a POST request and it opens in a new tab(probably when `target=_blank`). In this case, there will be no history but `document.referrer` will be present.

I am considering both use cases as you can see in the upper comment. Please let me know your thought on it.

westonruter:
@thelovekesh You've made a great point about the go back link not working as expected when opening in a new window. In the case of form submitting into a new window, we may not be able to get `document.referrer` either if the form had a `rel="noreferrer"` as well. The other possibility would be check `history.length` as noted in https://stackoverflow.com/a/3588420/93579 but that doesn't look reliable. 

It may be better to just tell the user to go back manually.

thelovekesh:
> It may be better to just tell the user to go back manually.

This option looks cleaner to me.

thelovekesh:
@westonruter I have changes made changes to `go back` as you suggested in the above comment. This PR is ready for your review now.

westonruter:
Can you enable me to be able to push commits to your fork?

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -69,7 +69,11 @@ ERROR_OFFLINE_URL, ERROR_500_URL, NAVIGATION_DENYLIST_PATTERNS, ERROR_MESSAGES *
 
 							let body = text.replace(
 								'{{{WP_SERVICE_WORKER_ERROR_MESSAGE}}}',
-								errorMessages.error
+								'errorMessages.error'
```
westonruter: What is this line about? Why replace with `'errorMessages.error'`?

maitreyie-chavan:
> Can you enable me to be able to push commits to your fork?

@westonruter I've sent over the invite ✔️

wp-includes/js/service-worker-navigation-routing.js
thelovekesh: Fixed 👍🏼

wp-includes/js/service-worker-navigation-routing.js
```diff
@@ -71,6 +71,10 @@ ERROR_OFFLINE_URL, ERROR_500_URL, NAVIGATION_DENYLIST_PATTERNS, ERROR_MESSAGES *
 								'{{{WP_SERVICE_WORKER_ERROR_MESSAGE}}}',
 								errorMessages.error
 							);
+							body = body.replace(
+								'{{{WP_SERVICE_WORKER_SUBMISSION_FAILURE_MESSAGE}}}',
+								''
+							);
```
westonruter: A problem here is that this will leave the document with:

```html
<p><strong></strong></p>
```

westonruter:
When I try navigating to a page while offline:

> image

If I try submitting a comment while offline:

> image

When I forcibly try to introduce a 500 error:

> image

In this last case, I'm seeing something which may be a regression caused by #438:

https://user-images.githubusercontent.com/134745/163492942-9f50a441-cc84-4854-92fa-79a0a34d95d8.mov

This seems like a bug. If there is a 500 error, we definitely shouldn't be attempting to reload the page, and we shouldn't be doing a reload for a POST request either.

westonruter (approved):
But that can be addressed in another PR. This looks good and behaves as expected otherwise.

PR #763, opened 2022-04-14
Prevent auto-reloading on error pages in response to POST requests

westonruter:
Originally in https://github.com/GoogleChromeLabs/pwa-wp/pull/697#discussion_r850892778:

> As part of #728, I just realized that this check will never do anything:
> 
> https://github.com/GoogleChromeLabs/pwa-wp/blob/124baabe30f70b718960fa18158f10a54943287e/wp-includes/template.php#L181-L184
> 
> This is because the offline template is always fetched via a `GET` request by the service worker when it is installing, so it will never not be `GET`. We need an alternative way of preventing the code from running on responses to non-GET requests.
> 
> Contrary to what I thought above, each reload does _not_ cause the browser re-POST message: [#728 (comment)](https://github.com/GoogleChromeLabs/pwa-wp/pull/728#issuecomment-1099718487)

We need to prevent this from happening:

https://user-images.githubusercontent.com/134745/163492942-9f50a441-cc84-4854-92fa-79a0a34d95d8.mov

We need to export the request method from the service worker fetch handler to the JS in the `wp_service_worker_offline_page_reload()` so that it can decide whether or not to try reloading.

### DIFF
--- a/wp-includes/js/service-worker-offline-post-request-handling.js
+++ b/wp-includes/js/service-worker-offline-post-request-handling.js
@@ -4,6 +4,24 @@
 (() => {
 	const errorMessages = ERROR_MESSAGES;
 
+	/**
+	 * Inject navigation request properties.
+	 *
+	 * @param {string}   body
+	 * @param {Request}  request
+	 * @param {Response} response
+	 * @return {string} Modified body.
+	 */
+	const injectNavigationRequestProperties = (body, request, response) => {
+		return body.replace(
+			'{{{WP_NAVIGATION_REQUEST_PROPERTIES}}}',
+			JSON.stringify({
+				method: request.method,
+				status: response.status,
+			})
+		);
+	};
+
 	const offlinePostRequestHandler = ({ event }) => {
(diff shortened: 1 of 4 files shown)


wp-includes/js/service-worker-offline-post-request-handling.js
```diff
@@ -102,11 +127,17 @@
...
 								'{{{WP_SERVICE_WORKER_ERROR_MESSAGE}}}',
 								`${connectionMessage} <strong>${errorMessages.submissionFailure}</strong>`
 							);
 
+							body = injectNavigationRequestProperties(
+								body,
+								event.request,
+								response
+							);
+
```
thelovekesh, line 140: Nice touch 👍🏼 . This will identify the correct response now.

thelovekesh (approved):
Changes LGTM 👍🏼

pooja-muchandikar:
QA Passed ✅  

The changes are working as expected. Automatic page reload is not happening if user is making the POST requests in offline mode.

PR #770, opened 2022-04-21
Fix maskable icon setting behaviour based on site icon

### LINKED ISSUE #304: Facilitate indicating the site icon as being maskable

thelovekesh:
This is a follow-up to #304 and was originally discussed in https://github.com/GoogleChromeLabs/pwa-wp/issues/304#issuecomment-1102984440. 

This PR aims to:

- Add logic to hide the `Maskable Icon` checkbox if site icon is not set.
- Add logic to remove `checked` from  `Maskable Icon` checkbox if the user has removed the site icon.

### DIFF
--- a/wp-admin/js/customize-controls-site-icon-pwa.js
+++ b/wp-admin/js/customize-controls-site-icon-pwa.js
@@ -18,6 +18,9 @@ wp.customize(
 					);
 				};
 
+				// Force the value to always correspond to whether the site icon is present, disregarding server control.
+				siteIconMaskableControl.active.validate = hasSiteIcon;
+
 				/**
 				 * Toggle site icon maskable active state based on whether the site icon is set.
 				 */
--- a/wp-admin/js/customize-controls-site-icon-pwa.js
+++ b/wp-admin/js/customize-controls-site-icon-pwa.js
@@ -18,9 +18,6 @@ wp.customize(
 					);
 				};
 
-				// Force the value to always correspond to whether the site icon is present, disregarding server control.
-				siteIconMaskableControl.active.validate = hasSiteIcon;
-
 				/**
 				 * Toggle site icon maskable active state based on whether the site icon is set.
 				 */
--- a/wp-admin/js/customize-controls-site-icon-pwa.js
+++ b/wp-admin/js/customize-controls-site-icon-pwa.js
@@ -31,6 +31,13 @@ wp.customize(
 				// Update active state whenever the site_icon setting changes.
 				siteIconSetting.bind(updateActive);
 
+				// Change the site_icon_maskable if site_icon is not set.
+				siteIconSetting.bind((newSiteIconValue) => {
+					if (!newSiteIconValue) {
+						siteIconMaskableSetting(false);
+					}
+				});
+
 				/**
 				 * Validate site icons for its presence and size.
 				 */
--- a/wp-includes/class-wp-customize-manager.php
+++ b/wp-includes/class-wp-customize-manager.php
@@ -35,13 +35,10 @@ function pwa_customize_register_site_icon_maskable( WP_Customize_Manager $wp_cus
 		$wp_customize->add_control(
 			'site_icon_maskable',
 			array(
-				'type'            => 'checkbox',
-				'section'         => 'title_tagline',
-				'label'           => __( 'Maskable icon', 'pwa' ),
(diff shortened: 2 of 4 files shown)


wp-includes/class-wp-customize-manager.php
```diff
@@ -35,10 +35,13 @@ function pwa_customize_register_site_icon_maskable( WP_Customize_Manager $wp_cus
...
-				'section'  => 'title_tagline',
-				'label'    => __( 'Maskable icon', 'pwa' ),
-				'priority' => $site_icon_control->priority + 1,
+				'type'            => 'checkbox',
+				'section'         => 'title_tagline',
+				'label'           => __( 'Maskable icon', 'pwa' ),
+				'priority'        => $site_icon_control->priority + 1,
+				'active_callback' => function() use ( $wp_customize ) {
+					return (bool) $wp_customize->get_setting( 'site_icon' )->value();
+				},
```
westonruter, line 44: Good. This sets the right initial state.

wp-admin/js/customize-controls-site-icon-pwa.js
```diff
@@ -31,6 +31,13 @@ wp.customize(
 				// Update active state whenever the site_icon setting changes.
 				siteIconSetting.bind(updateActive);
 
+				// Change the site_icon_maskable if site_icon is not set.
+				siteIconSetting.bind((newSiteIconValue) => {
+					if (!newSiteIconValue) {
+						siteIconMaskableSetting(false);
+					}
+				});
+
```
westonruter, line 40: This does capture @pooja-muchandikar's comment in https://github.com/GoogleChromeLabs/pwa-wp/issues/304#issuecomment-1102267429:

> If Site icon was set and set as maskable and then removed the site icon, Maskable Icon setting remains checked.

However, I didn't think this was necessarily a bug. It's just whether one thinks of the maskable property being linked to the icon, or if maskable is something that you would persist across icon changes.

westonruter:
In d19d86e I've addressed @pooja-muchandikar's point in https://github.com/GoogleChromeLabs/pwa-wp/issues/304#issuecomment-1102267429:

> * Maskable checkbox is in place but when the checkbox is checked from customizer, an error is shown on manifest, ideally it should show any error. Also the same error is displayed twice any reason?
>   image
> 
> image

Now when I set the `maskable` checkbox, I no longer get the warnings:

> image

DevTools seems to group the icons with duplicate dimensions horizontally (although it strangely doesn't automatically apply the mask to them).

westonruter (approved):

wp-admin/js/customize-controls-site-icon-pwa.js
thelovekesh: > However, I didn't think this was necessarily a bug.

Agreed, rather it is just a check to ensure that by default we are assuming that an icon is maskable. 

If we had to save maskable property in an icon meta then we will be able to guess it correctly, for an icon maskable property should be true or not, but saving maskable property in meta could be overkill IMO.

pooja-muchandikar:
QA Passed ✅ 

No warnings are displayed if maskable checkbox is checked. 

Screenshot 2022-04-25 at 10 03 36 AM

---

Backend settings are also working fine as expected.


https://user-images.githubusercontent.com/41000648/165021584-ac536bec-a789-4fb5-a09d-6ad9701d98ed.mov

PR #778, opened 2022-04-30
Add wp-env and utilize for running PHPUnit locally

westonruter:
This incorporates work from @delawski in https://github.com/ampproject/amp-wp/pull/6802, as well as prior work from @pierlon in https://github.com/ampproject/amp-wp/pull/5201.

In order to run tests with Xdebug in IntelliJ/PhpStorm in the included wp-env environment:

0. Open Preferences and go to **Languages & Frameworks > PHP**.

1. Open **Frameworks** and under **WordPress**, toggle on to Enable WordPress integration. 

For the **WordPress installation path**, obtain the root via `npm run wp-env install-path` and then append `/wordpress-develop/src`. For example:

```
/Users/westonruter/.wp-env/e6c93f608f26ae7e0e7a5f1fa29a6a3e/wordpress-develop/src
```

(The hash will be different on your system.)

Then add this path without `/src` to the PHP Include Path:

```
/Users/westonruter/.wp-env/e6c93f608f26ae7e0e7a5f1fa29a6a3e/wordpress-develop
```

2. Then open the **Servers** section and add a `pwa` server with the host set to `host.docker.internal` and port to 8888, with the project root mapping to `/var/www/html/wp-content/plugins/pwa`:

image

Also under **Include path** add the path mappings from `/src` and `/tests` from the previous step to map to `/var/www/html/` and `/var/www/wordpress-develop/tests` respectively:

image

3. Open the **Debug** section, and configure the External connections and Xdebug as following (which it may be by default):

image


westonruter:
I always get tripped up on configuring Xdebug with PhpStorm/IntelliJ, so if anyone has improvements to what I've outlined above I'd appreciate it. I want to make sure the best most-streamlined practice is followed. The notes are based on configuring a plugin-specific project, where the project root is the PWA repo root.

PR #793, opened 2022-05-30
Add id attribute to check network and reload script

thelovekesh:
While working on https://github.com/ampproject/amp-wp/issues/7122 I realized that the check network and reload script should have an id attribute that can be consumed later for some use cases related to DOM.


wp-includes/template.php
```diff
@@ -184,7 +184,7 @@ function wp_service_worker_offline_page_reload() {
 
 	?>
 	<script id="wp-navigation-request-properties" type="application/json">{{{WP_NAVIGATION_REQUEST_PROPERTIES}}}</script><?php // phpcs:ignore WordPressVIPMinimum.Security.Mustache.OutputNotation ?>
-	<script type="module">
+	<script id="check-network-and-reload" type="module">
```
westonruter:
```suggestion
	<script id="wp-offline-page-reload" type="module">
```

westonruter (approved):

PR #919, opened 2023-01-31
Handle the `site_icon_maskable` setting within a full-site-editing context

### LINKED ISSUE #915: Replace the need for the customizer when using a FSE theme
Hello everyone here,
hello @westonruter,

when using a full-site-editing theme together with the pwa plugin, the only need for having the customizer visible is the setting to set the `site_icon_maskable` option.

I thought about creating a [higher-order-component](https://developer.wordpress.org/block-editor/how-to-guides/block-tutorial/extending-the-query-loop-block/#adding-additional-controls) for the [`core/site-logo`](https://developer.wordpress.org/block-editor/reference-guides/core-blocks/#site-logo) block to fill this gap.

*What do you think?*

carstingaxion:
Introduce additional UI for the `core/site-logo`-block to toggle the existing setting and give a preview of the Image choosen, with the setting applied.

This should help deprecating the customizer at all, without loosing functionaliy and flexibility for the user.

This fixes #915


site-icon-maskable/site-icon-maskable-block-editor.php
```diff
@@ -0,0 +1,66 @@
...
+ * @see  https://developer.wordpress.org/reference/hooks/enqueue_block_editor_assets/
+ */
+function enqueue_block_editor_assets__site_icon_maskable() {
+	$dir = __DIR__;
+
+	$script_asset_path = "$dir/../build/site-icon-maskable.asset.php";
+	if ( ! file_exists( $script_asset_path ) ) {
+		throw new Error(
+			'You need to run `npm run blocks:start` or `npm run blocks:build` first, for the "core/site-logo"-block filter to work.'
+		);
```
westonruter: Since `Error` this is PHP 7+, I think this should be reworked to be be compatible with PHP 5.6 since that's still the minimum version supported.

site-icon-maskable/site-icon-maskable-block-editor.php
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * [pwa__enqueue_block_editor_assets description]
+ *
+ * @package PWA
+ * @since   0.8.0-alpha
+ * @throws Error Fatals out when the block-filter files weren't built.
+ *
+ * @see  https://developer.wordpress.org/reference/hooks/enqueue_block_editor_assets/
+ */
+function enqueue_block_editor_assets__site_icon_maskable() {
```
westonruter: I think these functions could just as well be added to the `WP_Web_App_Manifest` class since that's where we're working with the site icon and `maskable`.

Nevertheless, this is somewhat similar to https://github.com/GoogleChromeLabs/pwa-wp/blob/develop/wp-includes/class-wp-customize-manager.php but the intention there was that the code would eventually be added to `WP_Customize_Manager`.

carstingaxion:
Thank you @westonruter for all your feedback. 
And sorry for beeing a little late.

site-icon-maskable/site-icon-maskable-block-editor.php
carstingaxion: We can remove this completely
```suggestion
```

And instead go with `pwa.php#L99`
```php
if ( ! file_exists( PWA_PLUGIN_DIR . '/wp-includes/js/workbox-v' . PWA_WORKBOX_VERSION ) || 
	 ! file_exists( PWA_PLUGIN_DIR . '/wp-includes/js/workbox-v' . PWA_WORKBOX_VERSION . '/workbox-sw.js' ) || 
	 ! file_exists( PWA_PLUGIN_DIR . '/wp-includes/js/dist/site-icon-maskable.asset.php' )
	) {
	add_action( 'admin_notices', '_pwa_print_build_needed_notice' );
	return;
}
```

site-icon-maskable/site-icon-maskable-block-editor.php
carstingaxion: >I think these functions could just as well be added to the WP_Web_App_Manifest class since that's where we're working with the site icon and maskable.

:+1:  I agree, as this is what I thought by myself, too. 

The `WP_Customize_Manager` would not be the best place, from my opinion, because the *Customizer* as a whole maybe not future-proof, in favor of the new *Site editor*.

site-icon-maskable/site-icon-maskable-block-editor.php
carstingaxion: - [X] Remove `throw new error` and instead make use of `_pwa_print_build_needed_notice` instead, when files were not built https://github.com/GoogleChromeLabs/pwa-wp/pull/919/commits/47370814503dab51d4ddbcd45862479ca2ce9db6

site-icon-maskable/site-icon-maskable-block-editor.php
carstingaxion: - [X] Move the registration of the *maskable icon* settings into `WP_Web_App_Manifest` class, similar to the handling of the `short_name` option https://github.com/GoogleChromeLabs/pwa-wp/pull/919/commits/bf1a0a5dc9cd7cbe497c719b26119234762b31b8
- [X] Found a better place for the javascript (source) files and their enqueing function https://github.com/GoogleChromeLabs/pwa-wp/pull/919/commits/96cf3da4c8920717435ade52d7d42fb54e2c1ace, https://github.com/GoogleChromeLabs/pwa-wp/pull/919/commits/d203583242c3fe89c66cfa84ede85abd81ef70a0

carstingaxion:
What do you think about the current state @westonruter ?

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -677,4 +707,37 @@ function updateShortNameField() {
...
+	 */
+	public function enqueue_site_icon_maskable_block_editor_assets() {
+		$dir  = '/wp-includes/js/dist';
+		$path = PWA_PLUGIN_DIR . $dir;
+
+		// Stop, if needed file was not built successfully.
+		$script_asset_path = "$path/site-icon-maskable.asset.php";
+		if ( ! file_exists( $script_asset_path ) ) {
+			return;
+		}
```
westonruter, line 731: This is a redundant check with `_pwa_print_build_needed_notice`, right? So is it needed?

```suggestion
```

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -546,6 +554,28 @@ public function register_short_name_setting() {
 				'show_in_rest'      => true,
 			)
 		);
+		/**
```
westonruter: Since this is not a DocBlock per https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-2-multi-line-comments

```suggestion

		/*
```

wp-includes/js/src/site-icon-maskable/maskable-icon-controls.js
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * This package includes a library of generic WordPress components
+ * to be used for creating common UI elements shared between screens
+ * and features of the WordPress dashboard.
+ *
+ * @see https://developer.wordpress.org/block-editor/reference-guides/packages/packages-components/
+ */
+import {
+	// __experimentalHStack as HStack,
```
westonruter:
```suggestion
```

wp-includes/js/src/site-icon-maskable/maskable-icon-controls.js
```diff
@@ -0,0 +1,116 @@
...
+import { useSelect } from '@wordpress/data';
+
+export default function MaskableControls() {
+	const [siteIconMaskable, setSiteIconMaskable] = useEntityProp(
+		'root',
+		'site',
+		'site_icon_maskable'
+	);
+
+	// mainly borrowed from ...
```
westonruter, line 55: From what?

wp-includes/class-wp-web-app-manifest.php
```diff
@@ -677,4 +707,37 @@ function updateShortNameField() {
...
+		$script_asset_path = "$path/site-icon-maskable.asset.php";
+		if ( ! file_exists( $script_asset_path ) ) {
+			return;
+		}
+
+		$script_asset = require $script_asset_path;
+		$index_js     = "$dir/site-icon-maskable.js";
+		wp_enqueue_script(
+			'pwa-site-icon-maskable-block-editor',
+			plugins_url( $index_js, __FILE__ ),
```
westonruter: I'm getting a 404 for the script, with the URL including `...wp-includes/wp-includes...`. This is due to the plugin file not being used here as expected:

```suggestion
			plugins_url( $index_js, PWA_PLUGIN_FILE ),
```

wp-includes/js/src/site-icon-maskable/maskable-icon-controls.js
```diff
@@ -0,0 +1,116 @@
...
+			/>
+		);
+	}
+
+	return (
+		<InspectorControls>
+			<PanelBody>
+				<Flex align="start">
+					<FlexBlock>
+						<ToggleControl
```
westonruter, line 100: When I toggle this on the first tine I get a React warning:

```
Warning: A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
```

Perhaps this is because it is missing a default `false` value?

wp-includes/js/src/site-icon-maskable/maskable-icon-controls.js
```diff
@@ -0,0 +1,116 @@
...
+			<img
+				alt={__('Site Icon')}
+				className="components-site-icon"
+				src={siteIconUrl}
+				width={64}
+				height={64}
+				style={siteIconStyle}
+			/>
+		);
+	}
```
westonruter, line 93: I tried selecting a non-square image as the Site Logo and this is resulting in an unexpected result for the maskable icon:

> image

Three things aren't quite right here:

1. The toggle for a maskable icon is showing even if I've not selected that I want to use the Site Logo as the Site Icon. I should only see the toggle if I'm using the Site Logo as the Site Icon.
2. A Site Logo need not be square, but a Site Icon must be. Nevertheless, WP allows selection of a rectangular icon but then it crops it to be square. The same logic used for cropping to be square should also be replicated here in the preview. (Aside: I'm surprised that core doesn't have such a Site Icon preview when the "Use as site icon" toggle is checked, as there should be a way to see whether the square crop will even work. This would be an improvement for core.)
3. If I _change_ the selected image for the Site Logo while the block controls are open, the preview for the maskable icon does not update to use the newly selected image. See below:

> image

wp-includes/js/src/site-icon-maskable/maskable-icon-controls.js
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @see https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/
+ */
+import { useSelect } from '@wordpress/data';
+
+export default function MaskableControls() {
+	const [siteIconMaskable, setSiteIconMaskable] = useEntityProp(
+		'root',
+		'site',
+		'site_icon_maskable'
```
westonruter, line 52: This is unexpectedly showing up as the label in the pre-save dialog when it should have a label:

> image

westonruter (changes requested):
Off to a good start!

westonruter:
Relatedly, I just filed https://github.com/WordPress/gutenberg/issues/48608

wp-includes/class-wp-web-app-manifest.php
carstingaxion: You are right, that is a redundant check, but leaving it out will result in fatal error with the following `require` statement on the next line. Or am I wrong?

Additionally, because this is called inside the block-editor, where the normal `admin_notice`s will not be visible, this should prevent errors while trying to enque a non-existent file. To help the user mentioning, that the plugin was not built properly, we could introduce one of the newer [`createWarningNotice`](https://developer.wordpress.org/block-editor/reference-guides/data/data-core-notices/#createwarningnotice) for the block-editor. But this should be new issue, I think.

wp-includes/class-wp-web-app-manifest.php
carstingaxion: This seems to be the difference of: knowing some standards vs. living the standards.

Thank you, I appreciate to learn!

wp-includes/js/src/site-icon-maskable/maskable-icon-controls.js
carstingaxion: :+1:

wp-includes/js/src/site-icon-maskable/maskable-icon-controls.js
carstingaxion:
```suggestion
	// mainly borrowed from the `SiteIcon` component 
```
[`SiteIcon`](https://github.com/WordPress/gutenberg/blob/51e25a59affdb416eab193de764d75396d21f72e/packages/edit-site/src/components/site-icon/index.js#L15)

Unfortunately there is no nice docs-page to reference to, propably because this component is part of the *Edit Site* components, which are described in its own [README](https://github.com/WordPress/gutenberg/blob/51e25a59affdb416eab193de764d75396d21f72e/packages/edit-site/README.md) as
> [...] but please keep in mind that it might never get fully documented.

wp-includes/class-wp-web-app-manifest.php
carstingaxion: Yes, sorry.

wp-includes/js/src/site-icon-maskable/maskable-icon-controls.js
carstingaxion: This is unfortunately a known bug in gutenberg, I'm going to ref. the issue here, when I'll find it again ;)

wp-includes/js/src/site-icon-maskable/maskable-icon-controls.js
carstingaxion: OMG. But you are totally right on all of your points. 

I'll try to find some solutions.

wp-includes/class-wp-web-app-manifest.php
westonruter: > You are right, that is a redundant check, but leaving it out will result in fatal error with the following `require` statement on the next line. Or am I wrong?

If you see the change you made in `pwa.php`:

https://github.com/GoogleChromeLabs/pwa-wp/blob/81d28600276b84d91581cd0969fb9124bf99a978/pwa.php#L100-L106

Note that in the bootstrap, it does a `return` to prevent the rest of the PWA functionality from being loaded. So if the file doesn't exist, then this conditional won't ever be reached.

PR #935, opened 2023-02-21
Fix PHPStan analysis errors; Update `Build, test & measure` workflow

thelovekesh:
- Remove `default` arg for `site_icon_maskable` customizer setting. This setting is checkbox hence default value as bool not required.
- Update lint and static analysis jobs to use PHP 8.0
- Update deprecated `::set-output` to use `$GiTHUB_OUTPUT`.

### DIFF
--- a/wp-includes/class-wp-customize-manager.php
+++ b/wp-includes/class-wp-customize-manager.php
@@ -25,7 +25,6 @@ function pwa_customize_register_site_icon_maskable( WP_Customize_Manager $wp_cus
 		array(
 			'capability' => 'manage_options',
 			'type'       => 'option',
-			'default'    => false,
 			'transport'  => 'postMessage',
 		)
 	);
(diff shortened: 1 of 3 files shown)


wp-includes/class-wp-customize-manager.php
```diff
@@ -25,7 +25,6 @@ function pwa_customize_register_site_icon_maskable( WP_Customize_Manager $wp_cus
 		array(
 			'capability' => 'manage_options',
 			'type'       => 'option',
-			'default'    => false,
```
westonruter, line 28: Was this causing a problem?

westonruter (approved):

wp-includes/class-wp-customize-manager.php
thelovekesh: It was causing static analysis error as default arg is string.

thelovekesh: See [job](https://github.com/GoogleChromeLabs/pwa-wp/actions/runs/4107427913/jobs/7345812124) for runner logs.

https://github.com/WordPress/WordPress/blob/6d0eaf9674e550788dce7942d87af0e4f0147784/wp-includes/class-wp-customize-setting.php#L62-L68

PR #983, opened 2023-05-14
Bump node; Updates packages; Fix wp-env tests container

thelovekesh:
- Bump Node from 14 to 18 as 14 is EOL.
- Generated new lockfile with npm lockfile version 3.
- Fix wp-env tests container from phpunit to tests-WordPress as phpunit container is not supported now.
- Bump all deps to the latest version.


thelovekesh:
@westonruter We need to wait until https://github.com/WordPress/gutenberg/pull/50590 gets merged upstream. Will notify you once done.

PR #984, opened 2023-05-14
Fix multi-byte characters sanitization in short name field

### LINKED ISSUE #971: Short Name field sanitization does not account for multi-byte characters
As [reported](https://wordpress.org/support/topic/remove-tool-bar-2/#post-16666916) by @RavanH:

> By the way, I just found that if I set a 12 character Short Name with special character in it, like “On the Rhône” it will get cut off at 11 characters after saving. The short name with a simple “o” instead of the “ô” does not suffer this issue.

Yes, this is due to having used `substr()` instead of `mb_substr()` in [`sanitize_short_name`](https://github.com/GoogleChromeLabs/pwa-wp/blob/456f039c7d051e6c642bd5820e318c649b7b39e7/wp-includes/class-wp-web-app-manifest.php#L573-L579). I can see that using the latter will fix the issue: https://3v4l.org/DeJdr

Note that we _are_ using the multibyte-safe function in `is_name_short`:

https://github.com/GoogleChromeLabs/pwa-wp/blob/456f039c7d051e6c642bd5820e318c649b7b39e7/wp-includes/class-wp-web-app-manifest.php#L191-L195

We just need to use it in `sanitize_short_name` as well.

thelovekesh:
Fixes: #971

### DIFF
--- a/tests/test-class-wp-web-app-manifest.php
+++ b/tests/test-class-wp-web-app-manifest.php
@@ -581,6 +581,8 @@ class Test_WP_Web_App_Manifest extends TestCase {
 			'whitespace_padding' => array( '     WP Dev ', 'WP Dev' ),
 			'script_contains'    => array( 'WP <script>evil</script> Dev ', 'WP Dev' ),
 			'too_long'           => array( 'WordPress Develop', 'WordPress De' ),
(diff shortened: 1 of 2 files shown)


westonruter (approved):
Thanks!

PR #985, opened 2023-05-14
Add web app manifest display setting in Reading settings page

### LINKED ISSUE #970: Expose the Web App Manifest `display` member as setting in admin
One relatively common user request is to change the [`display`](https://developer.mozilla.org/en-US/docs/Web/Manifest/display) member of the Web App Manifest from the default `minimal-ui` to something else like `standalone`. There is currently a [wiki section](https://github.com/GoogleChromeLabs/pwa-wp/wiki/Web-App-Manifest#display) with the plugin code to enable this. However, since it comes up frequently enough, it should be exposed as a setting similar as the Short Name was done in #210.

thelovekesh:
Fixes: #970

### DIFF
--- a/wp-admin/options-web-app-manifest-display.php
+++ b/wp-admin/options-web-app-manifest-display.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * Web App manifest display field for the reading settings administration panel.
+ *
+ * As part of a core merge, the code in this file would go inside wp-admin/options-reading.php
+ *
+ * @package PWA
+ */
+
+namespace PWA_WP;
+
+/**
+ * Register web app manifest display setting.
+ */
+function register_web_app_manifest_display_setting() {
+	register_setting(
+		'reading',
+		'web_app_manifest_display',
+		array(
+			'type'              => 'string',
+			'sanitize_callback' => 'sanitize_text_field',
+		)
+	);
+}
+add_action( 'init', __NAMESPACE__ . '\register_web_app_manifest_display_setting' );
+
+/**
+ * Register web app manifest display setting field.
+ */
+function add_web_app_manifest_display_setting_field() {
+	add_settings_field(
+		'web_app_manifest_display',
+		__( 'Display', 'pwa' ),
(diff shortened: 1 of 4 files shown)


wp-admin/options-web-app-manifest-display.php
```diff
@@ -0,0 +1,71 @@
...
+ */
+function register_web_app_manifest_display_setting() {
+	register_setting(
+		'reading',
+		'web_app_manifest_display',
+		array(
+			'type'              => 'string',
+			'sanitize_callback' => 'sanitize_text_field',
+		)
+	);
```
westonruter, line 23: We should make sure this doesn't end up adding `web_app_manifest_display` as an autoloaded option, since it is only used in the web app manifest request.

I think this may mean adding a call to `add_option` before this call.

```php
add_option( 'web_app_manifest_display', 'minimal-ui', '', false );
```

Although there may be a better place for it.

I realize this should also be done for the `short_name` field.

wp-admin/options-web-app-manifest-display.php
```diff
@@ -0,0 +1,71 @@
...
+}
+add_action( 'admin_init', __NAMESPACE__ . '\add_web_app_manifest_display_setting_field' );
+
+/**
+ * Render web app manifest display setting field.
+ */
+function render_web_app_manifest_display_setting_field() {
+	// See: <https://developer.mozilla.org/en-US/docs/Web/Manifest/display#values>.
+	$allowed_values = array( 'fullscreen', 'standalone', 'minimal-ui', 'browser' );
```
westonruter, line 45: While probably out of scope of this UI, there is also a `display_override` property: https://developer.chrome.com/articles/display-override/

wp-admin/options-web-app-manifest-display.php
westonruter: And for the `offline_browsing` option.

westonruter: For existing installs that have `offline_browsing` and `short_name` options that are autoloaded, they should switch to non-autoloaded.

PR #1095, opened 2024-01-18
Fix PHP CI including PHPCS and PHPUnit; bump PHP/WP versions

westonruter:
Closes #983
Closes #1056


composer.json
```diff
@@ -5,28 +5,27 @@
...
   "config": {
     "allow-plugins": {
-      "dealerdirect/phpcodesniffer-composer-installer": true,
-      "civicrm/composer-downloads-plugin": true
+      "civicrm/composer-downloads-plugin": true,
+      "dealerdirect/phpcodesniffer-composer-installer": true
     },
     "platform": {
-      "php": "5.6"
+      "php": "7.0"
```
thelovekesh, line 28: Need to emulate it to `7.0` else it can download deps that don't support PHP 7.0.

westonruter: Humm. I had updated the `composer.lock` file for PHP 8.0 so we'll need to update it for 7.0. However, when I do so then I get errors when running tests locally:

```
$ npm run test:php

> test:php
> wp-env run tests-cli --env-cwd=/var/www/html/wp-content/plugins/pwa/ env WORDPRESS_TABLE_PREFIX=wptests_ vendor/bin/phpunit

ℹ Starting 'env WORDPRESS_TABLE_PREFIX=wptests_ vendor/bin/phpunit' on the tests-cli container. 

[18-Jan-2024 22:34:19 UTC] PHP Warning:  Private methods cannot be final as they are never overridden by other classes in /var/www/html/wp-content/plugins/pwa/vendor/phpunit/phpunit/src/Util/Configuration.php on line 176
[18-Jan-2024 22:34:19 UTC] PHP Fatal error:  Cannot acquire reference to $GLOBALS in /var/www/html/wp-content/plugins/pwa/vendor/phpunit/phpunit/src/Util/Configuration.php on line 543
✖ Command failed with exit code 255
```

If I change it to 7.4 and run `composer update` then I can run PHPUnit successfully, even when I'm running PHP 8.1. 

Can we change this then to (and run `composer update`):

```suggestion
      "php": "7.4"
```

If that means removing tests for PHP 7.0-7.3, I'm fine with that.

thelovekesh: It's because the PHPUnit version is not compatible with the PHP version inside the container. So I think we need to use updated PHPUnit for local testing.

So we have two options now:
- We save PHPUnit 9.6 PHAR in bin like we do for PHPStan and keep 7.0 tests.
- Remove 7.0 tests and bump PHP emulation to 7.1.

thelovekesh: Now PHPUnit version is customizable in local development with - https://github.com/GoogleChromeLabs/pwa-wp/pull/1095/commits/9fa812d36fc96f8ff6b6acde216675f475db2f71

With this, we can download any PHPUnit version and keep the tests for PHP 7.0.

westonruter: Clever!

PR #1202, opened 2024-11-05
Filter the 'no-cache' value for the 'Cache-Control' header

adekbadek:
Adds a filter for the value of the `Cache-Control` header sent with the SW. This will make it possible for developers to alter the caching behavior for the SW file.

As noted in the [PR introducing the header](https://github.com/GoogleChromeLabs/pwa-wp/pull/39), "Generally, most developers will want to set the Cache-Control header to no-cache" – this filter will be for the other developers :)


westonruter:
Couldn't you instead just do:

```php
add_action( 'wp_front_service_worker', function () {
   header( 'Cache-Control: ... ' );
} );
```

This will replace the header sent by the plugin.

adekbadek:
Ah, that works! Thank you.

